PR #1, opened 2016-08-03
Fixing connectivity 

bquenin:
appcelerator/amp-project#329

Setting setting `--net=host on container` in Makefile
Refactoring the code a little

test like this:
make run
make test

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -2,16 +2,16 @@ package server
 
 import (
 	//	"fmt"
+	"fmt"
 	"log"
 	"net"
+	"os"
 
-	"fmt"
 	"github.com/appcelerator/amp/api/rpc/project"
 	"github.com/appcelerator/amp/api/rpc/service"
 	"github.com/appcelerator/amp/data/elasticsearch"
 	"google.golang.org/grpc"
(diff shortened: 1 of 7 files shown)


api/server/server.go
```diff
@@ -5,11 +5,33 @@ import (
 	"log"
 	"net"
 
+	"fmt"
```
subfuzion: **Totally** nitpicking here, but but "fmt" and "os" should get moved up to be by "log" and "net". Run `make fmt` and gofmt will automatically sort the imports.

api/server/server.go
```diff
@@ -5,11 +5,33 @@ import (
...
+	"os"
 )
 
+var (
+	es elasticsearch.ElasticSearch
+)
+
+func init() {
+	// Get elasticsearch url from environment
+	ES_URL := os.Getenv("ES_URL")
```
subfuzion: All caps implies a constant, which it is not. In any case, it isn't idiomatic in Go to use all caps even for constants.

subfuzion:
@bertrand-quenin I'm working with Chris and wanted to get your PR merged into master, so I made the minor formatting edit for you. It's still not connecting to Elasticsearch for us, but I'll investigate.

PR #5, opened 2016-08-05
Cleaning up project code and tests

bquenin:


chrisccoy:
Hey @bertrand-quenin the client package that you are using is not the grpc version. clientV3 is the latest and greatest for etcd version 3 and it uses grpc. I think @subfuzion  was doing some stuff as well, so I think we should wait until after the sync.

subfuzion:
@bertrand-quenin @chrisccoy Please review https://github.com/appcelerator/amp/pull/7 for the direction I want to go in. That PR is for clientv3 support and replaces this one. I will flesh out the interface a bit more shortly, then add support for storing proto3 binary objects instead of strings/json. I will be adding value revision support so that mutate operations (update, delete) only succeed for requests at the correct state (needed for concurrency in a distributed environment).

PR #6, opened 2016-08-05
Adds flag support to amplifier

subfuzion:
Added flag support so that port and etcd endpoints can be supplied on the command line. A config object is supplied to the server that contains the parsed flags.

To be able to be parameterized by endpoints supplied on the command line, api/server can't use package initialization because the endpoints flag hasn't been parsed yet. `api/server/server.go` explicitly initializes etcd now after it is supplied with a config object.

Note: This PR doesn't fix the etcd code that erroneously says it has successfully established a connection. It only focuses on parsing flags and supplying a config object.


chrisccoy:
Review the amplifier changes and LGTM

PR #7, opened 2016-08-08
Added preliminary etcd clientv3 store support

subfuzion:
This PR makes a few improvements to the `Makefile`, including the ability to compile `*.proto` definitions wherever they are found in the source tree (excluding `.git` and `.vendor`. This is in preparation to being able to support storing and retrieving proto3 binary data objects instead of strings to the kv store (etcd).

This PR also provides a preliminary interface and implementation for a kv datastore. The interface is defined at `data/store.go` and the implementation is at `data/etcd/store.go`. There is initial transaction support, and I will be adding value revision support so that mutate operations (update, delete) only succeed for requests at the correct state (needed for concurrency in a distributed environment) in a subsequent PR.

Note: all the project code is currently commented out pending updates coming in upcoming PRs.
## Verification

Run the etcd implementation test:

```
$ ./swarm start etcd   # only start the etcd service
$ glide install
$ cd data/etcd
$ go test
```

The test will start the server, create a `foo` key under the `amp` path and set it to `bar`, then get it, then delete it.


data/etcd/store.go
```diff
@@ -0,0 +1,171 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package etcd
+
+import (
+	"context"
```
bquenin: this import doesn't compile on my environment
I think we need to use `golang.org/x/net/context`

data/store.go
```diff
@@ -0,0 +1,46 @@
+package data
+
+import (
+	"context"
```
bquenin: idem

data/etcd/store_test.go
```diff
@@ -0,0 +1,90 @@
+package etcd_test
+
+import (
+	"context"
```
bquenin: idem

bquenin:
All good!

```
go test
2016/08/08 15:14:21 connecting to etcd at http://localhost:2379
2016/08/08 15:14:21 connected to etcd at http://localhost:2379
2016/08/08 15:14:21 amplifier is listening on port 50101
PASS
ok      github.com/appcelerator/amp/data/etcd   1.047s
```

PR #8, opened 2016-08-08
Etcd ci

generalhenry:
runs data/etcd/store_test.go


subfuzion:
Don't remove the `vendor` directory. Please recommit and push with `vendor`. We had the conversation about this on the slack `amp` channel last week. Although we need to run `glide install` in the `dockerfile` to make sure any new deps get installed, we still don't want to exclude it from the docker build context (much faster to copy files than pull them from remotes), so please remove `vendor` from `.dockerignore`. (Actually, I don't think you added `vendor` to `.dockerignore` (sorry), but can you please remove it with this PR? Thanks!)

.dockerignore
```diff
@@ -1 +1,2 @@
-vendor
\ No newline at end of file
+vendor
```
subfuzion: I know you didn't put it here, but can you please remove `vendor` with your PR?

Makefile
```diff
@@ -64,7 +64,7 @@ install-host: proto-host
 
 # used to build under Docker
 proto-host: $(PROTOFILES)
-	@for DIR in $(DIRS); do cd $${DIR}; ls *.proto > /dev/null 2>&1 && protoc *.proto --go_out=plugins=grpc:.; done
+	@for DIR in $(DIRS); do cd $(BASEDIR)/$${DIR}; ls *.proto > /dev/null 2>&1 && protoc *.proto --go_out=plugins=grpc:. || true; done
```
subfuzion: Thanks for catching. I had this in place before, not sure how I managed to commit without it.

Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: all clean install install-server install-cli fmt simplify check version build run
+.PHONY: all clean install install-server install-cli fmt simplify check version build run test
```
subfuzion: Can you add `proto` as a dependency to the `test` rule? I got the following error when I ran `make test`:

`amp (etcd-ci)$ make test
api/server/server.go:10:2: no buildable Go source files in /Users/tony/projects/go/src/github.com/appcelerator/amp/api/rpc/service`

subfuzion:
Seems to be breaking my test at `data/etcd` (`go test`). Can you verify it works for you before I spend time investigating?

Makefile
subfuzion, line 67: You must not have rebased with master before submitting this PR? My previous PR incorporated this and it's in the `master` branch. See: https://github.com/appcelerator/amp/blob/master/Makefile#L58

generalhenry, line 67: Looks like you fixed it for proto, this is proto-host

generalhenry:
I'm running go test from the data/etcd folder, they're running fine on the build server https://build.amp.appcelerator.io/p/appcelerator/amp/7

subfuzion:
I don't think you rebased with master before you submitted your PR. It's breaking my test (`data/etcd/store_test.go`), plus you don't seem to have the changes that my early PR that merged this morning by @bertrand-quenin included (for example: https://github.com/appcelerator/amp/blob/master/Makefile#L58). See: https://github.com/appcelerator/amp/pull/7.

subfuzion:
> I'm running go test from the data/etcd folder, they're running fine on the build server https://build.amp.appcelerator.io/p/appcelerator/amp/7

Make sure you can run locally as well:

```
$ cd data/etcd
$ go test
```

generalhenry:
As to breaking local testing, it's using http://etcd:2379 as the etcd endpoint instead of localhost.  
`docker-compose -f docker-compose.test.yml start etcd`
`docker-compose -f docker-compose.test.yml build`
`docker-compose -f docker-compose.test.yml run sut` 
runs the tests

subfuzion:
I'm not comfortable with merging. It doesn't look like you rebased with `master` before you pushed the PR...  https://github.com/appcelerator/amp/pull/8/commits.  Your commit history should include my PR (https://github.com/appcelerator/amp/commits/master).

generalhenry:
I'll rebase it

subfuzion:
> As to breaking local testing, it's using http://etcd:2379 as the etcd endpoint instead of localhost.

Correct, but I don't want to break local tests (which we do all the time while developing) just to pass CI checks. We need to parameterize the swarm script I guess.

generalhenry:
created rebased PR https://github.com/appcelerator/amp/pull/9

PR #9, opened 2016-08-08
fix tests for ci

generalhenry:
rebased version of https://github.com/appcelerator/amp/pull/8


docker-compose.test.yml
```diff
@@ -4,9 +4,16 @@ services:
...
-      ES_URL: http://elasticsearch:9200
-  elasticsearch:
-    image: appcelerator/elasticsearch-amp:latest
+    command: ["go", "test", "-v", "github.com/appcelerator/amp/data/etcd"]
+  etcd:
+    image: quay.io/coreos/etcd
+    command: [
+      "etcd",
+      "-name", "etcd0",
+      "-advertise-client-urls", "http://etcd:2379,http://etcd:4001",
```
subfuzion: Port `4001` not needed. 2379 and 2380 are the IANA assigned ports now (http://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=etcd).

data/etcd/store_test.go
```diff
@@ -15,9 +15,9 @@ const (
 )
 
 var (
-	config server.Config = server.Config{
+	config = server.Config{
 		Port:          ":50101",
-		EtcdEndpoints: []string{"http://localhost:2379"},
+		EtcdEndpoints: []string{"http://etcd:2379"},
```
subfuzion: As much as I want CI tests to pass, I hate to break local tests which we use during development, and I don't want it to be error prone changing this and then remember to change it back. I suppose a quick and dirty way to fix for local development right now might be to add an entry to the HOSTS file that maps `etcd` to 127.0.0.1. I don't see any other easy ways besides running the tests themselves from a container attached to the swarm, but that involves a somewhat tedious build step. Thoughts?

generalhenry: I'm adding a command line flag

generalhenry: I added an env for setting it, defaulting to localhost

generalhenry:
I added an env to set the etcd hosts so local and ci tests flows are preserved

docker-compose.test.yml
```diff
@@ -4,9 +4,16 @@ services:
...
+      "-listen-client-urls", "http://0.0.0.0:2379,http://0.0.0.0:4001",
```
subfuzion: Same: just use 2379.

subfuzion:
LGTM. It's really nice to see passing CI builds now. Thanks! 👍

PR #17, opened 2016-08-09
Update swarm script to start min services without amp-pilot

### LINKED ISSUE #18: Remove amp-pilot dependency for swarm min services
amp-pilot is preventing services from starting. Remove the amp-pilot dependencies for the minimum service set to unblock development & testing.

subfuzion:
Closes #18 

Looks like an issue with `amp-pilot` that was preventing influxdb from starting. I've removed the amp-pilot specifics so that the service will start. I also removed the amp-pilot dependency for all the other `--min` services.

Verify service starts (1/1 replicas):

```
$ ./swarm start influxdb
$ docker service ls
ID            NAME      REPLICAS  IMAGE                         COMMAND
exgy13wjnegc  influxdb  1/1       appcelerator/influxdb:latest
```

Can also verify UI is available at: http://localhost:8083/

Confirm all of the services used with `--min` flag all start up successfully.

```
$ ./swarm start --min
. . .
```

### DIFF
--- a/swarm
+++ b/swarm
@@ -185,16 +185,13 @@ amplogservice() { # Owner: generalhenry
     appcelerator/amp-log-service:latest
 }
 
+# DEPENDENCIES kafka
 amplogworker() { # Owner: bertrand-quenin
   docker service create --network amp-swarm --name amp-log-worker \
     --label amp.swarm="infrastructure" \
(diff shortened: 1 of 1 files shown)


chrisccoy:
Services all came up normally

Issue #12, opened 2016-08-09
Add stat service


subfuzion:
Add an rpc stat service used to query status and various performance metrics from various backend services. Add a cli `stat` command for querying the service.

subfuzion:
@freignat91 Nice work, thanks. I'm going to close this issue now that the basic implementation is working well. We can open new issues for ongoing enhancements, etc.

PR #19, opened 2016-08-09
Stat

chrisccoy:
Initial Influx Client and Test for #12


api/rpc/stat/influx/stats.go
```diff
@@ -0,0 +1,92 @@
...
+		fmt.Printf("ERR: %s\n", err)
+		return "", err
+	}
+	if err = response.Error(); err != nil {
+		fmt.Printf("ERR: %s\n", response.Error())
+		return "", err
+	}
+	data, err := json.Marshal(response)
+	return string(data), err
+}
```
subfuzion: Need newline

api/rpc/stat/influx/stats_test.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+	if !strings.Contains(res, "measurements") {
+		t.Errorf("Expected String to contain %s, actual=%s \n", "Measurement", res)
+	}
+
+}
+func statsInit() {
+	s = New("http://localhost:8086", "_internal", "admin", "changme")
```
subfuzion: See [this](https://github.com/appcelerator/amp/blob/master/data/etcd/store_test.go#L32) for an example of parameterizing the host (`influxdb`) in the swarm CI environment.

api/rpc/stat/stats.go
```diff
@@ -0,0 +1,17 @@
+package stat
```
subfuzion: I don't think we need an interface for stat. I don't see us using a factory pattern to create instances from different implementations unless we also have an effective abstraction for the query interface (string vs query object, etc). In any case, I think the extra effort right now is premature. I would just make the `stat` type public (`Stat`) in your implementation file and have `New()` return a new instance of it directly for now. It would be easy enough to return an interface instead in the future if that becomes necessary. Finally, our stat service is intended to support queries for status as well as metrics aggregated from several different backend services, so it's not really about just abstracting influxdb vs another time-series db.

api/rpc/stat/influx/stats.go
```diff
@@ -0,0 +1,92 @@
...
+	return err
+}
+
+// writeSJON takes the response and marshals it to JSON
+func (s *stats) writeJSON(response *client.Response, w io.Writer) {
+	var data []byte
+	var err error
+	data, err = json.Marshal(response)
+	if err != nil {
+		fmt.Fprintf(w, "Unable to parse json: %s\n", err)
```
subfuzion: Just a reminder to remove this method altogether or else have the method return an error.

subfuzion:
Let's rename stats to stat. Other comments inline.

subfuzion:
Ok, looking good. Will you rename `stats` to `stat` on this PR or the next one?

docker-compose.test.yml
```diff
@@ -7,6 +7,7 @@ services:
     command: ["go", "test", "-v", "github.com/appcelerator/amp/data/etcd"]
```
subfuzion: @generalhenry @chrisccoy: `sut` command shouldn't be `amp/data/etcd` for this test. There are one of two choices: 1) The reviewer updates the test compose file for a PR or 2) invoke `make test` which should execute all the project tests. I don't like the first choice. With the second, we could call a new rule, like `make test-docker` that exports environment variables and runs the tests. You can supply environment variables to make ([ref](https://www.gnu.org/savannah-checkouts/gnu/make/manual/html_node/Environment.html), but this isn't ideal, and requires updating the compose file environment for sut. @generalhenry You should make the guidance clear for the rest of the team. In this PR, Chris is relying on `swarm start --min` to be run first before the test, which works for local testing but probably doesn't work with the current CI environment...

subfuzion:
LGTM. Just waiting on @generalhenry to chime in.

docker-compose.test.yml
generalhenry: Yeah we should be running `make test` and keeping make test up to date.

I don't see a good reason to make a separate `make test-docker` the local docker flow is to use `docker-compose -f docker-compose.test.yml` to bring up the backing services and run the sut.

Which means we'll need to keep `docker-compose.test.yml` up to date, and it should contain all services needed to test

subfuzion: That sounds good to me. @chrisccoy @bertrand-quenin @freignat91, what this means is add the services that need to be running to the compose file (and keep up-to-date with your PRs). `sut` should just invoke `make test`. Make sure you add your Go packages to test to the `test` in the `Makefile`.

subfuzion:
@chrisccoy  PRs should have a reference to an issue and some kind of description that summarizes what to expect out of the PR. Here are examples: https://github.com/appcelerator/amp/pull/7, https://github.com/appcelerator/amp/pull/17, https://github.com/appcelerator/amp/pull/6. Thanks.

generalhenry:
`- package: github.com/influxdata/influxdb/client/v2` is missing from glide.yaml

generalhenry:
I tried the tests locally with docker-compose.test.yml:

``` yaml
version: "2"

services:
  sut:
    image: appcelerator/sut
    build: .
    command: ["make", "test"]
    environment:
      endpoints: http://etcd:2379
      influxhost: influxdb
  etcd:
    image: quay.io/coreos/etcd
    command: [
      "etcd",
      "-name", "etcd0",
      "-advertise-client-urls", "http://etcd:2379",
      "-listen-client-urls", "http://0.0.0.0:2379",
      "-initial-advertise-peer-urls", "http://etcd:2380",
      "-initial-cluster-token", "etcd-cluster-1",
      "-initial-cluster", "etcd0=http://etcd:2380",
      "-initial-cluster-state", "new"
    ]
  influxdb:
    image: appcelerator/influxdb
    environment:
      FORCE_HOSTNAME: auto
      PRE_CREATE_DB: telegraf
      CONFIG_ARCHIVE_URL: "https://github.com/appcelerator/amp-config/archive/0.1.0.tar.gz"
```

and the tests passed

subfuzion:
@chrisccoy Can you incorporate Henry's `docker-compose.test.yml` contents from above and resubmit? Assuming all goes well, the CI build should pass and that should be an accurate refection of the PR's actual CI status.

chrisccoy:
@subfuzion fix has been incorporated

generalhenry:
Yay failing test :+1:  (It's the glide.yaml issue I pointed out)

subfuzion:
@chrisccoy As I mentioned on slack, you didn't push an update to your `glide.yaml` (and `glide.lock`) file to add `github.com/influxdata/influxdb/client/v2`. After you ran `go get github.com/influxdata/influxdb/client/v2` in your branch, it should have modified those files, so you would have needed to add them to your commit. You mentioned glide issues earlier, so I have the feeling you ran `go get` when you were on a different branch or something. I don't know, but bottom line is your glide files weren't part of your push.

subfuzion:
@chrisccoy Not quite sure yet about the build failure, but go ahead and rebase your PR on top of commits from `master` and reconcile the updates to `docker-compose.test.yml` and then force push. We'll go from there.

api/rpc/stat/influx/stat_test.go
```diff
@@ -29,11 +30,14 @@ func TestQuery(t *testing.T) {
 }
 
 func statsInit() {
+	//Need to sleep for CI swarm to launch stacks
+	time.Sleep(1000 * time.Millisecond)
```
subfuzion: Don't do this. If you can't connect, fail fast with a panic. To get CI build passing is a responsibility outside of our server. I'm adding a script to make sure upstream services are healthy before running `make test`.

subfuzion:
@chrisccoy Don't add a sleep to the initialization code to get the CI build to pass. If the service can't connect to an upstream service during initialization, fail fast with a panic. To get CI build passing based on service orchestration is a responsibility outside of this scope. We'll only do this as a last resort as a total kludge if we have to, but we shouldn't. I will be adding a script to make sure upstream services are healthy before running make test. Go ahead and focus on reconciling the branch conflicts first.

subfuzion:
LGTM. Tests passed, nice work. 👍

Issue #15, opened 2016-08-09
Consolidated agent


subfuzion:
Create a single agent in Go that can provide the same functionality as our current `amp-docker-agent` and `amp-log-service`/`amp-log-agent`. Evaluate what functionality of our original docker agent is still necessary at the node level and make recommendations on what, if anything, can be eliminated if the information we previously captured is available with alternative strategies under Docker 1.12. If we do not need an agent on every node, then we can consider consolidating the functionality into `amplifier`.

Our concept of log collation corresponds to Docker Cloud's log multiplexing and is very similar to the work that @generalhenry worked on with `amp-log-service`, so please coordinate with him. @bertrand-quenin will be working on the log consumer side and adding the cli interface for it (see https://github.com/appcelerator/amp/issues/11).

freignat91:
added project repository: amp-agent: https://github.com/appcelerator/amp-agent
A new service v1.0.0-0, on this version this agent:
- Read container Docker events to have an up to date internal list of the started containers
  -Get service_id, service_name, node_id, state, health of this containers
  -For each create a logs stream and send it to kafka
  -has a API for its own healthcheck and containers info (to check if all are logged)

its service command is

  docker service create --network amp-swarm --name amp-agent \
    --mode global \
    -p 5001:3000 \
    --label amp.swarm="infrastructure" \
    --mount type=bind,source=/var/run/docker.sock,target=/var/run/docker.sock \
    appcelerator/amp-agent

Next step make it HA using Elasticsearch last timestamp

subfuzion:
@freignat91 Looking good so far!

subfuzion:
@freignat91 Let's have a conversation about the pros/cons of consolidating this into amp and what that would mean if we did...

freignat91:
That only the first version, needs to be tested more in details, but the main principles are there except HA for now.

freignat91:
amp-agent v1.0.0-1: Added: send dockers events to kafka
commit: https://github.com/appcelerator/amp-agent/commit/0904c9a387aea344da54779d337c1dadabf34f52

freignat91:
Testing and stabilizing the version before going further.

freignat91:
create PR on appcelerator/app: https://github.com/appcelerator/amp/pull/20
to add amp-agent, remove amp-docker-agent and amp-log-service, add kibana, kafkamanager
then logs can be viewed on kibana after amp startup
amp-log worker work can be followed in kafkamanager on its consumer ui part.

[...]

PR #20, opened 2016-08-10
update app swarm for amp-agent appcelerator/amp#15

freignat91:
appcelerator/amp#15

add amp-agent
remove amp-docker-agent and amp-service-log
add kafka-manager and kibana in amp 'min' to make kafka and log tests easier

### DIFF
--- a/swarm
+++ b/swarm
@@ -20,9 +20,8 @@ USAGE
 
 # please keep sorted
 IMAGES=(
-   appcelerator/amp-docker-agent:latest
+   appcelerator/amp-agent:latest
    appcelerator/amp-log:latest
-   appcelerator/amp-log-service:latest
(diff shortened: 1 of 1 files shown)


bquenin:
tested locally before merge

swarm
```diff
@@ -66,12 +64,14 @@ SERVICES=(
 )
 
 MINSERVICES=(
-# amp-agent
+  ampagent
   amplogworker
   etcd
   elasticsearch
   kafka
+  kafkamanager
```
subfuzion, line 72: Do we really need kafkamanager in the our min services set?

subfuzion:
Not sure we really need kafkamanager in our set of min services, but otherwise LGTM.

Issue #14, opened 2016-08-09
Add GitHub authorization to amp


subfuzion:
In the short term, we will need to implement the GitHub non-web app oauth flow for authenticating cli users (`amp`) to the backend `amplify`.

See: https://developer.github.com/v3/oauth/#non-web-application-flow

PR #23, opened 2016-08-10
github login

generalhenry:
https://github.com/appcelerator/amp/issues/14

login command using github

not complete

to run: `amp login` and then enter your username, password and OTP as needed.


subfuzion:
Looking good. I just want to mention I wouldn't spend effort right now on beautifying it too much with colors, etc. Let's get functionality in place first. The docker cli (and `docker login`) are very utilitarian and functional and I'm fine with that. Not saying not worth the effort to beautify later, but lean and mean right now is just fine.

generalhenry:
The colors are mostly just me getting used to golang

PR #24, opened 2016-08-10
CI adjustments

generalhenry:
added vendor to `.dockerignore` to prevent any working around the glide files

changed go test output to verbose (better output on build server)

changed command back to `make test`

moved glide installation to bottom of the `Dockerfile` to prevent unneeded reinstalls

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 *.pb.go
+vendor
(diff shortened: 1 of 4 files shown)


subfuzion:
LGTM. Will merge after pending build passes.

PR #28, opened 2016-08-12
Implemented gRPC Service and Test for Stat

chrisccoy:
Ongoing development for #12 
- Implemented grpc code
- implemented stat_test that launches grpc server and executes as client

Note: I decided not to integrate with the api/server module at this point. Once this gets fleshed out a bit, I will integrate, but this test will execute through "make test"

### DIFF
--- a/api/rpc/stat/stat.proto
+++ b/api/rpc/stat/stat.proto
@@ -3,13 +3,14 @@ syntax = "proto3";
 package stat;
 
 service Stat {
-  rpc GetMeasurments (MeasurementRequest) returns (MeasurementReply) {}
+  rpc ExecuteQuery (QueryRequest) returns (QueryReply) {}
 }
 
-message MeasurementRequest {
+message QueryRequest {
   string database = 1;
+  string query=2;
 }
 
-message MeasurementReply {
-  repeated string measurement = 1;
+message QueryReply {
+  string response = 1;
 }
--- a/api/rpc/stat/stat_test.go
+++ b/api/rpc/stat/stat_test.go
@@ -1,1 +1,60 @@
 package stat
+
+import (
+	"log"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"golang.org/x/net/context"
+)
+
+const (
+	svcAddr  = "localhost:51001"
+	queryVal = "measurements"
+)
+
+var (
+	srv  *Stat
+	host string
+)
+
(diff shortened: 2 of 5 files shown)


api/rpc/stat/stat.proto
```diff
@@ -3,13 +3,14 @@ syntax = "proto3";
 package stat;
 
 service Stat {
-  rpc GetMeasurments (MeasurementRequest) returns (MeasurementReply) {}
+  rpc ExecuteQuery (ExecuteQueryRequest) returns (ExecuteQueryReply) {}
```
subfuzion: Just nitpicking here, but `ExecuteQueryRequest` strikes me as a tad verbose. Could this be shortened to `QueryRequest` and `QueryReply`?

api/rpc/stat/stat.proto
```diff
@@ -3,13 +3,14 @@ syntax = "proto3";
...
 }
 
-message MeasurementRequest {
+message ExecuteQueryRequest {
```
subfuzion: Just nitpicking here, but ExecuteQueryRequest strikes me as a tad verbose. Could the messages be shortened to QueryRequest and QueryReply?

api/rpc/stat/stat_test.go
```diff
@@ -1 +1,72 @@
...
+	lis, err := net.Listen("tcp", svcAddr)
+	if err != nil {
+		panic(err)
+	}
+	//Create inline server
+	s := grpc.NewServer()
+	//Register Server
+	RegisterStatServer(s, createStatServer())
+	//Launch Server with defered stop
+	go s.Serve(lis)
```
subfuzion: Not sure it's worth all the extra boilerplate just to explicitly start one service instead of all of them when the server starts as fast as it does. You sure you don't just want to start the server like [this test](https://github.com/appcelerator/amp/blob/master/data/storage/etcd/store_test.go#L53)? Maybe we ought to just create a utility function for testing that just starts a single rpc service on a port that at least doesn't require copying all this boilerplate over for each test. And frankly, until we're doing actual integration tests that goes through our client code to talk to a running server and actually test our network interface end-to-end, we really don't even need to be starting a server at all right now. I think this is what you are alluding to in the description, but why don't you just simplify things and remove altogether right now until it's actually ready to be used?

subfuzion:
LGTM

generalhenry:
:+1:

Issue #10, opened 2016-08-09
Add protocol buffer support to etcd store


subfuzion:
We will store typed data in our key/value store using using protocol buffers. Add the ability to marshal/unmarshal protocol buffers to the implementation of our etcd store interface.

subfuzion:
#32 is a small PR that implements initial support for marshaling/unmarshaling protocol buffers to etcd key/value storage with create, get, and delete operations. The PR does not close this issue. Subsequent PRs will wrap the object so that metadata (such as version) is also encapsulated by the storage object, and will add more storage operations, like guaranteed updates with put.

subfuzion:
I've provided a storage.Object that is used for wrapping a payload (protocol buffer message) and metadata for using with the backend. Working on providing tests for it now.

subfuzion:
Basic implementation is in place. Closing this issue now; we can open new ones for specific enhancements, etc.

PR #32, opened 2016-08-14
Etcd protocol buffer storage support

subfuzion:
#10

This is a small PR that implements initial support for marshaling/unmarshaling protocol buffers to etcd key/value storage with create, get, and delete operations. This PR does not close the issue. Subsequent PRs will wrap the object so that metadata (such as version) is also encapsulated by the storage object, and will add more storage operations (such as put).
## Verification

```
$ cd data/storage/etcd
$ go test -v
```


generalhenry:
Just a clarification note, it will wrap it with the database(etcd) metadata

subfuzion:
Yes, we're talking about things like `createdIndex`, `modifiedIndex` ... the type of metadata used in a distributed k/v store for reasoning about ordering in the presence of concurrent updates. ([reference](https://coreos.com/etcd/docs/latest/api.html#get-the-value-of-a-key))

PR #33, opened 2016-08-15
Add a monitor command to the swarm script

subfuzion:
This PR adds a `monitor` command to the `swarm` script to provide a continuously updating display of the swarm's replica status. Use from another terminal to the see the state of your services. By default, it updates the display every 5 seconds; you can change the refresh interval by providing an argument in seconds.

```
$ ./swarm monitor
$ ./swarm monitor 2  # refresh every 2 seconds, 5 is the default
```

### DIFF
--- a/swarm
+++ b/swarm
@@ -82,7 +82,7 @@ ROLE=infrastructure
 usage() {
   cat << USAGE >&2
 Usage:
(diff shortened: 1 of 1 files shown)


chrisccoy:
Tested this and LGTM

generalhenry:
monitored my swarm starting :+1:

PR #34, opened 2016-08-15
Initial restructuring of client code

subfuzion:
This PR moves the old `cli` code to `api/client` as part of the initial refactoring to make the distinction clearer that this code shouldn't have any CLI responsibilities. It's really meant as a single client shim that wraps individual rpc services from a single CLI perspective (such as the CLI) and is parameterized with whatever configuration we want to supply that applies environment wide. This may end up being totally unnecessary, and we might just have the CLI communicate with RPC services directly, but we'll evaluate this as we proceed forward.

Note that the cobra.Command that @bertrand-quenin introduced will need to be refactored out. This PR just lays the groundwork for that.
## Verification

```
$ make install
$ amp
```

CLI should build and start without errors. There isn't anything else to test for now.


generalhenry:
:+1:

chrisccoy:
:+1:

PR #35, opened 2016-08-16
GitHub oauth

### LINKED ISSUE #14: Add GitHub authorization to amp

generalhenry:
closes https://github.com/appcelerator/amp/issues/14
## Verification (testing only flow)
- [Create](https://github.com/settings/applications/new) an OAuth application (to simulate the amp backend)
- Note the `Client ID` and `Client Secret`
- Start `amplifier` with the following flags:
  $ amplifier --clientid XXX --clientsecret XXX &
- Test the CLI login command
  $ amp login
  
  When prompted, enter username, password, and two-factor auth code
- CLI should indicate success


generalhenry:
In order to function properly amplifier needs to be run with both the clientid and clientsecret flags

To generate your own github application see: https://developer.github.com/v3/oauth/

and use https://github.com/settings/applications/new to generate your own application to test with

subfuzion:
@To clarify you only need to set up an app for testing. I used http://ignore.me for URLs.

subfuzion:
Something like the following should be in the description...

## Verification (testing only flow)
- [Create](https://github.com/settings/applications/new) an OAuth application (to simulate the amp backend)
- Note the `Client ID` and `Client Secret`
- Start `amplifier` with the following flags:
  $ amplifier --clientid XXX --clientsecret XXX &
- Test the CLI login command
  $ amp login
  
  When prompted, enter username, password, and two-factor auth code
- CLI should indicate success

api/client/login.go
```diff
@@ -0,0 +1,118 @@
...
+			fmt.Println(err)
+			os.Exit(1)
+		}
+	}
+
+	a.Configuration.Github = lastEight
+	a.Configuration.Save()
+
+	color.Set(color.FgCyan)
+	fmt.Println("token saved")
```
subfuzion: Nitpicking on a great PR ... but can you change this to something like "Logged in" or "<user> is now logged in" and on failure, something to correspond, like "Login failed for <user>"...

chrisccoy:
:+1: 

```
amp (cli version: 1.0.0, build: 7b164aa3)
Login using github
github username or email: chrisccoy
github password: 

Login using github
github username or email: chrisccoy
github password: 

Two Factor Authentication required
authentication code: 

token saved
```

api/client/login.go
```diff
@@ -0,0 +1,118 @@
...
+	"regexp"
+
+	"github.com/appcelerator/amp/api/rpc/oauth"
+	"github.com/fatih/color"
+	"github.com/howeyc/gopass"
+	"golang.org/x/net/context"
+)
+
+func getUsername() (username string) {
+	color.Set(color.FgMagenta, color.Bold)
```
subfuzion: @generalhenry We were chatting about this yesterday when I mentioned we need to fix logs. It wasn't my intention that `api/client` have any CLI/terminal concerns. I figured those would be helpers that would be part of `cmd/amp` (we discussed having `cmd/amp/cli` for all the helpers or even just `cli` as I had originally. The `api/client` package would be where I would expect to do things like ensure the security token is included in the grpc context to the backend, etc.

api/client/amp.go
```diff
@@ -18,14 +23,48 @@ const (
 // Configuration is for all configurable client settings
 type Configuration struct {
 	Verbose bool
+	Github  string
 	Target  string
 	Images  []string
 }
 
+// Save saves the configuration to ~/.ampswarm.yaml
+func (c *Configuration) Save() (err error) {
```
subfuzion: This isn't a concern of the `api/client` package which is more about our client-server communication protocol. It should be moved to `cmd/amp/cli`...

subfuzion:
Looking good. Please run `make fmt` and push again.

```
amp (github-oauth)$ make check
./cmd/amp/cli/cli.go
[WARN] Fix formatting issues with 'make fmt'
```

cmd/amp/cli/cli.go
```diff
@@ -0,0 +1,36 @@
...
+func SaveConfiguration(c interface{}) (err error) {
+	homedir, err := homedir.Dir()
+	if err != nil {
+		return
+	}
+	contents, err := yaml.Marshal(c)
+	if err != nil {
+		return
+	}
+	err = ioutil.WriteFile(path.Join(homedir, ".ampswarm.yaml"), contents, os.ModePerm)
```
subfuzion: Please rename `.ampswarm.yaml` -> `.amp.yaml` (corresponds to the executable name)

subfuzion:
@generalhenry Looking good. Please clean up history (don't need to see all the fmt commits, etc), rename `.ampswarm.yaml` to `.amp.yaml` to match the binary, and run `make fmt`. Thanks!

subfuzion:
(Thanks for renaming `.ampswarm` -> `.amp` ... I actually forgot the original reason I named it that; it was so it wouldn't clash with the `amp` bash cli at the time, which is no longer relevant.) 👍

api/client/login.go
```diff
@@ -0,0 +1,76 @@
+package client
+
+import (
+	"regexp"
+
+	"github.com/appcelerator/amp/api/rpc/oauth"
+	"github.com/appcelerator/amp/cmd/amp/cli"
```
subfuzion: This shouldn't have to refer (or even know about) the `cli` package.

api/client/login.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return
+	}
+	lastEight = authReply.SessionKey
+	name = authReply.Name
+	return
+}
+
+func basicLogin(a *AMP) (username, password, lastEight, name string, err error) {
+	username = cli.GetUsername()
```
subfuzion: These flows should be controlled by the `cli` package's log helper.

subfuzion:
Much nicer! Thanks 👍

PR #38, opened 2016-08-16
Vendor repair

generalhenry:
glide install --strip-vcs --strip-vendor


subfuzion:
Also needs `--updated-vendored` (when there is an existing `vendor` directory).

```
$ glide install --strip-vcs --strip-vendor --update-vendored
```

and 

```
$ glide update --strip-vcs --strip-vendor --update-vendored
```

I'm pushing an update that adds this to the Makefile.

```
$ make install-deps
$ make update-deps
```

subfuzion:
Thanks for helping with this @generalhenry - I think this will help with the issues that have been nagging everyone, while also addressing the main reason for committing the `vendor` directory in the first place. I also added the `--update-vendored` flag and then added these all to the makefile to make it easy for everyone to remember.

subfuzion:
I will add glide to run in a container in the makefile (like protoc), but everyone should just make sure they are running the latest glide (0.11.1).

This will put it in `$GOPATH/bin`

```
$ curl https://glide.sh/get | sh
```

PR #43, opened 2016-08-17
Refactor Influx Code to Return Response struct

chrisccoy:
Related to #12 Changed the Query function to return the Influx Struct and added a separate Marshal func. This will make it easier to marshal the  Influx Response struct into the proto buf messages

### DIFF
--- a/api/rpc/stat/stat.go
+++ b/api/rpc/stat/stat.go
@@ -29,9 +29,16 @@ func New(cfg Config) (*Stat, error) {
 	return &Stat{conn: c}, err
 }
 
-// ExecuteQuery implements business logic for StateServer interface
+// ExecuteQuery implements business logic for StatServer interface
 func (s *Stat) ExecuteQuery(ctx context.Context, req *QueryRequest) (*QueryReply, error) {
 	resp, err := s.conn.Query(req.Query)
-	rep := &QueryReply{Response: resp}
+	if err != nil {
+		return nil, err
+	}
+	data, err := s.conn.Marshal(resp)
+	if err != nil {
+		return nil, err
+	}
+	rep := &QueryReply{Response: data}
 	return rep, err
 }
--- a/data/influx/influx.go
+++ b/data/influx/influx.go
@@ -38,16 +38,24 @@ func (s *Influx) query(query string, database string) client.Query {
 }
 
 // Query executes the provided query string and returns the results as a JSON object
-func (s *Influx) Query(q string) (string, error) {
+func (s *Influx) Query(q string) (*client.Response, error) {
 	// ExecuteQuery runs any query statement
 	response, err := s.client.Query(s.query(q, s.dbname))
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 	if err = response.Error(); err != nil {
+		return nil, err
+	}
+	return response, err
+}
+
+//Marshal Convert client.Respnse to JSON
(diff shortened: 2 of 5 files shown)


subfuzion:
@chrisccoy Take a look at the CI failure and work with @generalhenry to resolve.

Also, the test still doesn't work for me locally:

```
--- FAIL: TestQuery (0.02s)
        influx_test.go:23: unable to decode json: received status code 405 err: invalid character 'M' looking for beginning of value
```

Until I can get to the bottom of it, you'll need it to pass for Henry and the CI build.

api/rpc/stat/stat.go
```diff
@@ -32,6 +32,13 @@ func New(cfg Config) (*Stat, error) {
 // ExecuteQuery implements business logic for StateServer interface
```
subfuzion: I don't think the comment is current.

chrisccoy: Definitely a typo, should say StatServer, but I think it is current per the proto.pb.go files declaration:
```

type StatServer interface {
	ExecuteQuery(context.Context, *QueryRequest) (*QueryReply, error)
}
```
Am I missing something?

data/influx/influx.go
```diff
@@ -38,16 +38,24 @@ func (s *Influx) query(query string, database string) client.Query {
 }
 
 // Query executes the provided query string and returns the results as a JSON object
```
subfuzion: Since we're not returning JSON for a REST API, do we need to marshal to JSON? It seems that you might want to use helpers that transform client.Response to typed objects. Or perhaps an expedient method for now is to leverage json and then unmarshal into protocol buffer messages used by the rpc service using [jsonpb](https://godoc.org/github.com/golang/protobuf/jsonpb). This would make it a bit more like the data/storage/etcd package that accepts an out parameter that is an instance of the actual empty proto.Message instance to unmarshal the result into.

api/rpc/stat/stat.go
```diff
@@ -29,9 +29,16 @@ func New(cfg Config) (*Stat, error) {
 	return &Stat{conn: c}, err
 }
 
-// ExecuteQuery implements business logic for StateServer interface
+// ExecuteQuery implements business logic for StatServer interface
```
subfuzion: I didn't mean this is a typo. I meant literally is this statement accurate? I don't think we have a StatServer interface. Am I missing it?

data/influx/influx.go
chrisccoy: Actually, I intend to use helpers as you suggest, for the transformation, as the JSON is a bit nested and would not be very friendly for use in a proto message. I left the JSON because I thought it was useful for testing and debugging.

api/rpc/stat/stat.go
chrisccoy: This is the interface generated from the "make proto" step, so it exists with the stat.pb.go module

subfuzion: I would absolutely not expect implementation of an rpc interface to be in `data/influxdb`. This isn't the right abstraction. That belongs in `stat.go`, which should be a consumer of this library package.

subfuzion: Sorry @chrisccoy -- totally my fault here. I was thinking about the influx package, not the stat package when I wrote that.

subfuzion:
👍

PR #44, opened 2016-08-18
Adds preliminary list support

subfuzion:
#10

This PR adds initial list support for fetching values from etcd. The interface has a placeholder for Filter functions, but this isn't implemented yet.
## Verification

```
$ ./swarm start etcd
$ make test-storage
```

### DIFF
--- a/data/storage/etcd/store_test.go
+++ b/data/storage/etcd/store_test.go
@@ -1,8 +1,11 @@
 package etcd_test
 
 import (
+	"fmt"
 	"log"
 	"os"
+	"path"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -141,6 +144,7 @@ func TestUpdate(t *testing.T) {
 		t.Error(err)
 	}
 
+	// confirm
 	out := &storage.Project{}
 	ignoreNotFound := false
 	ctx2, cancel2 := newContext()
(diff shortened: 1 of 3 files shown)


data/storage/etcd/store_test.go
```diff
@@ -163,6 +168,77 @@ func TestUpdate(t *testing.T) {
...
+	ctx := context.Background()
+
+	// store everything under amp/foo/
+	key := "foo"
+
+	// this is a "template" object that provides a concrete type for list to unmarshal into
+	obj := &storage.Project{}
+
+	// unlimited ttl
+	ttl := int64(0)
```
generalhenry: why not the more ideomatic:
``` go
var ttl int64
```

generalhenry:
:+1:

data/storage/etcd/store_test.go
subfuzion: I do get the idiom of letting Go initialize things to the [zero value](https://golang.org/ref/spec#The_zero_value), but for my tests I wanted to be explicit. I had been playing with different TTLs in the tests and decided for now to leave at 0 although I'll replace with a reasonable timeout once I'm ready to close the issue. It also matches the way I'm initializing everything else, as well as shows others how the TTL can be initialized to a different value.

chrisccoy:
:+1:

PR #46, opened 2016-08-20
Support running individual tests

subfuzion:
This PR supports running individual tests from make (ex: `make test-storage`). To run all the tests, just run `make test` like before.


chrisccoy:
Good idea :+1:

generalhenry:
:+1:

PR #47, opened 2016-08-20
Update glide deps

subfuzion:
This PR explicitly tags the influx client to match the server version in our image (0.13), and updates the remaining dependencies to the latest versions.


ndegory:
@subfuzion influxdb 1.0 is not far (planned for summer), we can already start to work on the RC. I'll create a branch but I'll leave master (and latest) on 0.13.

subfuzion:
@ndegory Yes, I was actually testing with `v1.0.0-beta3`. The main thing we have to ensure is to update the entire TICK stack at the same time to maintain version alignment, and then we have to update our glide files (for the client package dependencies).

chrisccoy:
LGTM 
for verification:
make install-deps
make test

generalhenry:
:+1:

PR #50, opened 2016-08-23
add CLI stat cpu

freignat91:
PR for test a way to implement stat CPU command which can be easily duplicate to implement mem, io, net


api/rpc/stat/stat.proto
```diff
@@ -3,14 +3,29 @@ syntax = "proto3";
...
+  string filterServiceID=4;
+  string filterContainerName=5;
+  string filterContainerID=6;
+  string filterContainerImage=7;
+  string filterNodeID=8;
 }
 
-message QueryReply {
-  string response = 1;
+message CPUEntry {
```
chrisccoy, line 20: I beleive the values should all be float. 
https://github.com/influxdata/telegraf/blob/master/plugins/inputs/system/cpu_test.go#L18

api/rpc/stat/stat.go
```diff
@@ -21,24 +21,164 @@ type Config struct {
...
+		cpuReply.Entries[i] = entry
+	}
+	return &cpuReply, nil
+}
+
+//TODO: It's possible to have container/service/node 6 functions in only two and simply the code.
+
+//Extract CPU information for one container according to CPURequest
+func (s *Stat) containerCPUQuery(ctx context.Context, req *CPURequest, containerID string) (*CPUEntry, error) {
+	ql := fmt.Sprintf("SELECT container_name, usage_in_kernelmode, usage_in_usermode, usage_system, usage_total FROM docker_container_cpu WHERE container_id='%s' ORDER BY time DESC LIMIT 1", containerID)
```
chrisccoy, line 67: The code below could be consolidated, since it is repeated throughout

api/rpc/stat/stat.proto
freignat91: in fact it's the number of CPU ticks per type (user, system, kernel, total) so integer, then the % can be computed dividing user / total or system / total in float and it can be computed the same way even with group by statement for instance because int can be added considering the / is done at the end.

api/rpc/stat/stat.go
freignat91: yes, for sure, considering it's not the final code, it's more an evaluation, i won't do it on this code, but on the real one which depends on the decision on the amp stat ... format

api/rpc/stat/stat.proto
chrisccoy: I think there are two different measurements. docker and then system. The docker_container_cpu  stats are stored as int, but cpu is stored as float

freignat91:
Close PR. it's replaced by 'amp stat' #61 one

PR #57, opened 2016-08-24
Export connect/disconnect

subfuzion:
Need to export connect and disconnect from `api/client` so that they are accessible to CLI commands.

Verification

```
$ make
# start amplifier with --clientid and --clientsecret
$ amp login
```


chrisccoy:
Not sure about build fail. The code seems straight fwd though. :+1: once it passes

chrisccoy:
:+1:

PR #60, opened 2016-08-24
keep only one input (kafka) on telemetry worker

### LINKED ISSUE #59: netstat metrics from telemetry workers appeared in influxdb
Main issue is that when the Kafka input fails, the telemetry worker doesn't exit, because there's at least one input that is working correctly. the netstat plugin (enabled by default in the image) should be explicitely removed from the telemetry worker.

ndegory:
fix #59 

Removes the netstat input plugin on the telegraf telemetry worker


freignat91:
LGTM

subfuzion:
LGTM

PR #58, opened 2016-08-24
Updated readme with login info, other minor edits

subfuzion:
Provides instructions to start amplifier so that you can use GitHub auth with `amp login`.

PR #61, opened 2016-08-24
amp stats

freignat91:
#12

Initial amp stat command implementation (based on proposal [here](https://drive.google.com/open?id=1ii8iTXIZFpneYSiqD7muVWLD6c_7NegsypnJ1oOG7Rw)).

Not implemented
- Historic parameters 
- `--status` and `--health` (requires amp-agent/etcd)
## Verification

```
$ amp stat --cpu --container --service-name=kafka
```

Should display cpu metrics of all containers for the `kafka` service

```
$ amp stat --mem --service
```

Should display memory metrics for all services

```
$ amp stat --io --node
```

Should display io metrics for all nodes


subfuzion:
@freignat91 You merged [my PR](https://github.com/appcelerator/amp/pull/58) yesterday with the updated README instructions for CLI login. Is it not working for you?

subfuzion:
Closing for now, we can reopen when ready.

api/rpc/stat/stat.go
```diff
@@ -1,44 +1,239 @@
...
+	} else {
+		req.Discriminator = "node"
+	}
+	return idFieldName, nameFieldName
+}
+
+
+//Compute the influx 'sql' query string concidering StatRequest
+func buildQueryString(req *StatRequest, ressourceFieldName string, metric string) string {
+	query := fmt.Sprintf("SELECT distinct %s FROM docker_container_%s WHERE container_id!=''", ressourceFieldName, metric)
```
chrisccoy, line 205: Not sure this is the best approach. This query will return all of the id values from the beginning of time. As the measurement grows this will yield undesirable results. Also, a single query can be used to accomplish the same thing. Check the Observabiilty doc related to "group by and limit 1"

api/rpc/stat/stat.go
```diff
@@ -1,44 +1,239 @@
...
+	res, err := s.Influx.Query(query)
+	if err != nil {
+		return nil, err
+	}
+
+	cpuReply := CPUReply{}
+	list := res.Results[0].Series[0].Values
+	cpuReply.Entries = make([]*CPUEntry, len(list))
+	for i, row := range list {
+		entry, err := s.ressourceCPUQuery(ctx, req, idFieldName, nameFieldName, row[1].(string))
```
chrisccoy, line 29: I think we can improve on this. I do not like the idea of making a net call for each resource id obtained in the first query; this is inefficient. There is a way to accomplish this by executing 1 query using the group by and limit 1.  for example,

```
select 
"com.docker.swarm.service.name" as service, 
container_name, 
container_id, 
usage_in_kernelmode as usage_kernel, 
usage_in_usermode as usage_user, 
usage_total, 
usage_system 
from docker_container_cpu 
where time > now() - 1m 
GROUP BY "com.docker.swarm.service.name", container_name 
LIMIT 1
```

In the even that we should need to execute multiple queries in succession, then this approach is preferred:
https://docs.influxdata.com/influxdb/v0.13/query_language/data_exploration/#multiple-statements-in-queries

freignat91:
No, the CLI even following intruction didn't work for me

freignat91:
Can rebase the stat branch on master, too much conflicts
I close this PR and recreate a new branch from scratch

api/rpc/stat/stat.go
freignat91: After test, LIMIT 1 doesn't work as excepted. we got just on row where we need to got all the row corresponding to all the container matching the filter

chrisccoy: what is the query that you are testing? I have been using the Web UI and influx with positive results.

PR #64, opened 2016-08-25
clean should also uninstall amplifier

subfuzion:
`make clean` removes all autogenerated files and uninstalls the `amp` CLI. It should also uninstall `amplifier` (the amp server).


generalhenry:
:+1:

chrisccoy:
:+1:

Issue #70, opened 2016-08-25
Investigate telegraf worker connection with kafka


subfuzion:
Telegraf occasionally doesn't connect to kafka, which means logs aren't produced. telemetry-worker can say that it's up and healthy yet not have successfully connected to kafka. We probably need a healthcheck for our telegraf image that actually checks to see if it can connect to kafka, and if not, exit with an error so that telegraf is correctly reported as unhealthy. This will cause telegraf to be restarted according to the healthcheck retry policy.

ndegory:
@subfuzion I wasn't able to reproduce this issue with swarm start --min, do you have a specific test to make it happen?

subfuzion:
@ndegory It doesn't happen all the time, but it is frequent enough and everyone of us has experienced this. The only recourse for us is to restart the swarm. You can talk to @freignat91 about it; maybe he can try to reproduce for you in the Puteaux office.

ndegory:
reproduced with latest version of zookeeper and kafka, issue is still there.

> 2016/08/26 16:53:40 Connected to 172.19.0.3:2181
> 2016/08/26 16:53:40 Authenticated: id=96484092623519747, timeout=4000
> 2016/08/26 16:53:40 Started the kafka consumer service, peers: [zookeeper:2181], topics: [telegraf]
> 2016/08/26 16:53:40 Kafka Consumer Error: kafka: error while consuming telegraf/-1: zk: node does not exist
> 2016/08/26 16:54:01 Output [file] buffer fullness: 0 / 1000 metrics. Total gathered metrics: 0. Total dropped metrics: 0.

ndegory:
I'm now able to reproduce the issue in a repeatable way, and I'm trying to patch telegraf to make it fail when the kafka input is not properly initialized.

ndegory:
Workaround now available in Docker hub (tag telegraf-0.13)

subfuzion:
@ndegory Awesome, thanks for this. This is good for now, at least until we switch away from Kafka! ;)

Issue #48, opened 2016-08-20
Influx test fails on OS X


subfuzion:
Running the influx test on OS X (Docker for Mac) fails with the following output:

```
$ make test-influx
=== RUN   TestQuery
--- FAIL: TestQuery (0.01s)
        influx_test.go:23: unable to decode json: received status code 405 err: invalid character 'M' looking for beginning of value
FAIL
```

The test failed on my Mac and on freshly installed Mac. However, it passes on Linux and it passes on the CI build.

[This issue](https://github.com/influxdata/influxdb/issues/6540) indicates that the problem is a mismatch between the client and the server versions. However, I made sure to pull the latest appcelerator/influxdb image on my Mac as well as the test Mac. I also ensured that the client package dependency was [explicitly tagged](https://github.com/appcelerator/amp/pull/47/commits/40d1de1c6b56c8f9c4557e2926d1416202faa280#diff-395f41b2a27b70ce44399c91c82158c2R28) for `0.13` in glide. I furthermore experimented by creating an updated image on my machine for `v1.0.0-beta3` and updated the client package for the same tag with glide. I still get the same error and `make test-stat` also fails for the same reason. It's critical that we resolve this issue quickly and ensure that we maintain parity between Linux, Docker for Mac and Docker for Windows.

[This PR](https://github.com/appcelerator/amp/pull/47) explicitly sets the version tag for the client dependency to match the server version use by [influxdb image](https://github.com/appcelerator/docker-influxdb/blob/master/Dockerfile#L4).

subfuzion:
I reported the issue at the InfluxDB repo: https://github.com/influxdata/influxdb/issues/7184

ndegory:
Great, now I can ask IT to provide me  with a macbook :)

subfuzion:
Yes! You should definitely have one. So I also posted to the [influxdb slack channel](https://gophers.slack.com/archives/influxdb/p1471682768001183) ... hopefully someone there can offer some insight also.

subfuzion:
Resolved. It was coincidental that the issue manifested  on two machines that happened to be running Docker for Mac. There was in fact a version mismatch between the client and the server depending on whether or not the client package was in sync with the repo's server image. We should not pull `latest` from the swarm script, but pull explicitly tagged versions instead so that any mismatch is a bit more conspicuous.

ndegory:
[...]

PR #74, opened 2016-08-26
TICK images version explicitely set

ndegory:
related to #70 and #48 

version is still "0.13", once the influxdb client is updated, the INFLUXDATA_VERSION should be changed to "1.0"


subfuzion:
Awesome, thanks @ndegory 👍

PR #75, opened 2016-08-26
Update vendor deps to use semver

subfuzion:
This PR updates `glide.yaml` to use semver for specifying dependencies. (It also updates `glide.lock` and `vendor`).
## Verification

```
$ make
# should succeed

# confirm:
$ rm -rf vendor
$ make install-deps
$ should succeed
```


freignat91:
tested ok

PR #82, opened 2016-08-26
Add kafka hostname to /etc/hosts

ndegory:
#69
- at start, adds the hostname of the Kafka container in /etc/hosts for local resolution (for now works when only one container is launched)
- at stop, removes the hostname of the Kafka container from /etc/hosts

### DIFF
--- a/advertise_service
+++ b/advertise_service
@@ -14,7 +14,7 @@ _advertise_service(){
   local _service=$1
   if [[ -z "$_service" ]]; then echo "ERROR: $(basename $0) should be called with the name of a service"; return 1 ; fi
   while [[ -z $_container_id ]]; do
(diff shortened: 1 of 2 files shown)


subfuzion:
@ndegory Thanks for getting this done so quickly. I don't think you need to exit with an error if there is more than one replica. I think it's fine to just return the first hostname since the client should be able to talk to any kafka replica in the cluster.

ndegory:
@subfuzion you're right, commit 2c578a8 should do it

subfuzion:
Awesome, works great for me. Thanks. 👍

subfuzion:
I'm leaving the issue open while we consider our long term strategy.

Issue #88, opened 2016-08-27
advertise_service error modifying /etc/hosts


subfuzion:
Relates to #69 #82 

I closed #82 prematurely. After stopping the swarm, I get the following error:

```
sed: 1: "/etc/hosts": extra characters at the end of h command
```

If I modify the `advertise_service` to use `127.0.0.1`, I get the error when starting the swarm.

The script should work with `127.0.0.1` (it should default to this and perhaps be settable with another script parameter, ie $2):

```
$(dirname $0)/advertise_service kafka 127.0.0.1
```

ndegory:
@subfusion: can you please cat your /etc/hosts?

chrisccoy:
I am seeing errors when I run:
`./swarm monitor`
`sed: couldn't open temporary file /etc/sedxq4SUL: Permission denied`
Also this one:
`has been added to /etc/hostshttp:///var/run/docker.sock/v1.24/containers/json?filters={"name":{"kafka\\.":true}}: EOF`

subfuzion:
@chrisccoy Because `advertise_host` needs to access `/etc/hosts`, use sudo when starting and stopping the swarm:  `sudo ./swarm start --min` and `sudo ./swarm stop`). You're seeing the permission denied error coincidentally after running `./swarm monitor` because `advertise_host` is still running in the background, but that's where the error came from.

Not sure about the second error. That looks pretty grisly, but is probably not accurate (unless you ran with sudo, it couldn't have written to /etc/hosts). See if it goes away once you run the commands with sudo.

@ndegory Looks like the script still needs work.

subfuzion:
@ndegory Per your request, my `/etc/hosts` file is fine and as long as I leave your default of `127.0.1.1`, this is what it looks like after starting the swarm:

```
$ cat /etc/hosts
##
# Host Database
#
# localhost is used to configure the loopback interface
# when the system is booting.  Do not change this entry.
##
127.0.0.1       localhost
255.255.255.255 broadcasthost
::1             localhost
fe80::1%lo0     localhost

127.0.1.1 92dc3bf0006b
```

The last entry will be removed when I stop the swarm.

The error I described occurs when I update the script to use `127.0.0.1`. Keep in mind that OS X (and maybe Windows, although I don't know) don't use `127.0.1.1` as a loopback alias. I know that Debian (and Ubuntu) bind 127/8 (ie, 127.0.0.0 => 127.255.255.255) to the loopback interface, but I don't know how universal that is.

[...]

PR #92, opened 2016-08-28
appcelerator/amp#88 - fix advertise_service on Mac OS

ndegory:
#88
- in place sed has more option limits on Mac OS than linux
- use 127.0.0.1 for local loop
- minor fixes

### DIFF
--- a/swarm
+++ b/swarm
@@ -125,7 +125,7 @@ main() {
         startservices "${@:2}"
       ;;
       stop)
-        $(dirname $0)/advertise_service remove kafka &
+        $(dirname $0)/advertise_service -r kafka &
         removeservices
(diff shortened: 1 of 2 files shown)


swarm
```diff
@@ -125,7 +125,7 @@ main() {
         startservices "${@:2}"
       ;;
       stop)
-        $(dirname $0)/advertise_service remove kafka &
+        $(dirname $0)/advertise_service -r kafka &
```
subfuzion, line 128: Why not put this in `removeservices` (and be symmetric with `startservices`)?

subfuzion:
Tested, works great. Thanks @ndegory 👍

PR #93, opened 2016-08-28
Updated deps for amp logs testing

subfuzion:
This PR adds the [difflib](https://github.com/pmezard/go-difflib/tree/master/difflib) needed for @bertrand-quenin's `amp logs` tests. Separating the dependency update in its own PR makes it easier to view just the relevant changes in the following PR after rebasing.

The way `glide get` works is that other related dependencies in the dependency graph also get updated when installing a new packaged, hence the total number of files.
## Verification

Run tests after pulling to verify package updates don't break anything.


chrisccoy:
:+1:

PR #101, opened 2016-08-29
better cleanup of /etc/hosts

ndegory:
if a container name has been left in /etc/hosts, advertise_service will remove it instead of complaining about it


chrisccoy:
I have confirmed these results upon swarm start:

```
WARN - e69d64212a3b is advertised and look like a Docker hostname, removed from /etc/hosts
88602e2ab212 has been added to /etc/hosts
```

subfuzion:
LGTM

Issue #86, opened 2016-08-27
Update Kafka image to latest stable version (10.0.0.1)


ndegory:
current image is 10.0.0.0, many fixes in the minor point update:
https://archive.apache.org/dist/kafka/0.10.0.1/RELEASE_NOTES.html

subfuzion:
@ndegory thanks. So is the [commit](https://github.com/appcelerator/docker-kafka/commit/1bb575d5dc14429cd57154d2c85b4553c6cb0aa7) what [latest](https://hub.docker.com/r/appcelerator/kafka/tags/) is now? We should lock down the [tag](https://github.com/appcelerator/amp/blob/master/swarm#L24) (like we did for the tick stack) and continue that process for all our dependencies now.

ndegory:
Not latest, I didn't want to push it until the PR is accepted.
But the PR pushed a tag amp-issue-86. There's no unit test on this repo so I didn't want to tag it with a version on Docker hub.
Update: tag amp-issue-86 has also been copied to 10.0.0.1, and I made a copy of latest to 10.0.0.0.

subfuzion:
👍

Issue #79, opened 2016-08-26
Update Zookeeper to latest stable version (3.4.8)


ndegory:
Current zookeeper image is version 3.4.6, there's been many issues fixed since this version: http://zookeeper.apache.org/doc/r3.4.8/releasenotes.html

subfuzion:
@ndegory Same comment/recommendation as [here](https://github.com/appcelerator/amp/issues/86#issuecomment-243004513). Thanks.

ndegory:
PR still open, so not yet a `latest` image.
But unit tests are OK, so I pushed a tag `3.4.8`
I'll push a tag `3.4.6` from before the change, but I'll leave the tag `latest` to the merge of the PR.

subfuzion:
👍

Issue #78, opened 2016-08-26
Zookeeper and Kafka images include an Oracle JDK


ndegory:
appcelerator/zookeeper and appcelerator/kafka images are based on https://hub.docker.com/r/anapsix/alpine-java/~/dockerfile/, which breaks the Oracle license terms.
see http://blog.takipi.com/running-java-on-docker-youre-breaking-the-law/ for details

subfuzion:
@ndegory: thanks for pointing out the issue.

I was mulling over switching to [NATS](http://nats.io/) and @bertrand-quenin also mentioned it a few days ago on slack. This just adds another argument for eliminating anything jvm-based from our platform, and will probably make this a priority in the next sprint. I'm pretty sure as long as we agree to the license, build our own images, and host them ourselves, we haven't violated any agreement by containerizing; however, I don't feel like worrying about and dancing around Oracle's license, or switching to OpenJDK, or jumping through other hoops to avoid the mess with Java in any case. The heck with Oracle and Java. NATS is written in Go, and is available as [open source on GitHub](https://github.com/nats-io/gnatsd) with an [MIT license](https://raw.githubusercontent.com/nats-io/gnatsd/master/LICENSE).

subfuzion:
@ndegory We will evaluate [NATS Streaming](http://nats.io/documentation/streaming/nats-streaming-intro/) as the alternative. I'm in favor and pretty comfortable with the idea of switching based on early looks, but we need to ensure we evaluate message durability using [file storage](https://github.com/nats-io/nats-streaming-server#user-content-file-store) and feel as comfortable with it as with kafka, which has been battle-tested and is well-documented. This evaluation is something I'd like to coordinate with you and @bertrand-quenin.

ndegory:
@subfuzion The 2 (open) PR linked to this issue replace the Oracle JDK by an openJDK JRE.
So it's not a blocker anymore (once accepted), but if we can get rid of Java all together, I won't complain ;)

subfuzion:
@bertrand-quenin Is already working on prototyping a switch to NATS. 👍

subfuzion:
Since we're switching to NATS, this is no longer an issue for us.

ndegory:
anyway, it's fixed (not in latest tags of the images, as the PR were not accepted), amp swarm already uses the new images.

PR #107, opened 2016-08-31
Kafka zk versions

ndegory:
#78 #79 #86
- versions of kafka and zookeeper are now explicit instead of _latest_
- updated to kafka 10.0.0.1 and zookeeper 3.4.8


subfuzion:
LGTM

PR #109, opened 2016-08-31
add ETCDCTL_API env var to enable etcdctl on etcd container

freignat91:
little update on swarm file for etcd service 
added ETCDCTL_API=3 env. var
then it allows to use etcdctl using:
docker exec -it [containerId] /etcdctl [command...]
to manually add, get, browse etcd data

### DIFF
--- a/swarm
+++ b/swarm
@@ -269,6 +269,7 @@ etcd() { # Owner: subfuzion
     --label amp.swarm="infrastructure" \
     -p 2379:2379 \
     -p 2380:2380 \
+    -e ETCDCTL_API=3 \
     quay.io/coreos/etcd:v3.0.4 etcd \
       --name etcd \
       --listen-client-urls http://0.0.0.0:2379 \


subfuzion:
@freignat91 I don't think this is necessary. This is what I do:

```
docker run -t --rm --net=host -e ETCDCTL_API=3 quay.io/coreos/etcd:v3.0.4 etcdctl
```

I actually have that as an alias in my `.bashrc` file.

```
# etcd
alias etc='docker run -t --rm --net=host-e ETCDCTL_API=3 quay.io/coreos/etcd:v3.0.4 etcdctl'
```

With that, I can test `etcd` running in the swarm:

```
$ etc put hello world
OK
$ etc get hello
hello
world
```

With this approach, we don't have to modify the swarm `etcd` service, and it also isn't necessary to get its container id. This is fine for local testing, and only this approach can be updated to connect to a remote etcd host (you can't use `docker exec` to do that).

subfuzion:
@freignat91 I'll close the PR shortly. If you feel strongly that we should merge this, please provide your reasons or discuss with me on slack. Thanks.

freignat91:
Alias is ok, i close it. Thanks

PR #114, opened 2016-09-01
gitignore should not ignore amp and amp directories

subfuzion:
This PR fixes `.gitignore` so only the platform-specific binaries generated by `make build` for `amp` and `amplifier` are ignored, but not the `amp` and `amplifier` subdirectories under `cmd`.


bquenin:
LGTM

PR #115, opened 2016-09-01
Added engine-api for swarm service

subfuzion:
This PR adds the `engine-api` vendor package, which will be needed for `amp service`.
## Verification

```
$ amp install-deps
```


bquenin:
LGTM

Issue #81, opened 2016-08-26
Kafka / Kapacitor port conflict


ndegory:
Both Kafka and Kapacitor declare the same port (9092), resulting in conflicts when both services run on same node

subfuzion:
@ndegory @chrisccoy I don't know that we will be able to demo alerts on Monday anyway. Our choice right now is to not run Kapacitor if the config can't be updated before then. In any case, we will eventually need to update Kapacitor's config to use a different port, even if we disable it for now.

chrisccoy:
@ndegory were you able to confirm that this will fix the intermittent missing measurement issue?

ndegory:
@chrisccoy I'm not sure it's related.
@subfuzion we still have to work on how the alerts work, it's not yet demo ready, so I guess for Monday it's better to remove Kapacitor from the list of services.
I don't think mapping the port to a different one for Kapacitor will work, because Kapacitor subscribes to InfluxDB with its hostname and port, which I guess are the ones it knows, from the container, not the host, so the mapping would fail.

subfuzion:
@chrisccoy @ndegory We have a lot of issues in flight at the moment. I'm going to deemphasize the priority for alerts for the time being until after the Limited Preview.

ndegory:
@subfuzion  we have another option, we can use dynamically allocated UDP ports on Kapacitor for the InfluxDB subscriptions, it works with docker-compose (ports are not exposed, but that's not an issue), I hope it works too in Swarm, I'll do a test...

bquenin:
Obsoleted by #113

PR #116, opened 2016-09-01
Kapacitor udp

ndegory:
#81
- use udp dynamic ports instead of tcp:9092
- used TICK configuration v0.2.0 (update needed for influx-1.0.X)


subfuzion:
LGTM

Issue #113, opened 2016-09-01
Switch from Kafka to NATS Streaming


subfuzion:
Replace message queuing infrastructure with NATS streaming service. This will require:
- [ ] Updating agents that push messages to queue.
  - [x] ampagent (appcelerator/amp-agent#2)
  - [ ] telegrafagent (this should be done as a plugin)
- [ ] Updating workers that pull from the queue to update
  - [x] amplogworker (appcelerator/amp-log-worker#10)
  - [ ] telemetryworker
- [ ] rpc streaming service clients
  - [x] logs (appcelerator/amp#123)
  - [ ] stats

NATS streaming uses protocol buffers.

subfuzion:
@bertrand-quenin is currently working on amp-log-worker.

subfuzion:
@chrisccoy researched NATS and is following this issue https://github.com/influxdata/telegraf/pull/1697 for the telegraf -> NATS output plugin.

subfuzion:
@bertrand-quenin @chrisccoy @ndegory @freignat91 As I mentioned on slack yesterday, I'm going to relax the requirement for file store for guaranteed delivery for the alpha. We can focus on that as part of our productionizing/hardening efforts subsequently. @ndegory If you have the bandwidth to research and evaluate, that's great, but it's not a priority for the moment.

chrisccoy:
As a quick update, I confirmed that when using the image `appcelerator/nats-streaming:0.2` from `nats` branch, the basic client works with the streaming server without changes. I think @ndegory mentioned this was possible today in the sync, but I was able to confirm. The streaming server embeds the core server, so it works like a charm:
https://github.com/nats-io/nats-streaming-server#running

I am going to start working on integrating the strategy with `telegrafagent` and `telemetryworker`

chrisccoy:
Next steps are to integrate with influxdb, but, so far, NATS producer and consumer are working with telegraf, which is encouraging.

bquenin:
Nats prototype for logging is done. Instruction on how to test is available in #123 PR.

subfuzion:
I mentioned that I was willing to relax the requirement for guaranteed delivery for now [above](https://github.com/appcelerator/amp/issues/113#issuecomment-244138687) (although I wasn't expecting pathological behavior) as noted [here](https://github.com/appcelerator/amp/issues/141).

[...]

PR #118, opened 2016-09-01
Nats prototype

bquenin:
#113

Work in progress


swarm
```diff
@@ -347,6 +351,14 @@ haproxy() { # Owner: freignat91
     appcelerator/haproxy:latest
 }
 
+nats-streaming() { # Owner: bquenin
```
subfuzion, line 354: It's a legal name in bash, but not recommended for shell variables (should just be letters, digits, and underscores) and definitely not legal in `sh`.

PR #128, opened 2016-09-02
make install-deps fix for linux

subfuzion:
When glide is run in a container with `make install-deps`, the current directory is mounted. On Docker for Windows, the packages that are installed have the correct user id and group, but on Linux, they are created with root as the owner, which causes issues when attempting to checkout different branches.

This PR fixes the issue by supplying the current user and group to the `docker run` command. Files that are created should correctly indicate the current user as the owner.
## Verification

```
$ sudo rm -rf ~/.glide .glide vendor   # `glide cc` should be enough, but not always
$ make install-deps
```

Confirm that `vendor` and files below it are owned by the correct owner, not root.


subfuzion:
This is currently not a good fix because glide gets confused that the user doesn't actually exist in the image (no entry in `etc/passwd`). I am going to try a different strategy, so closing this PR for now.

PR #130, opened 2016-09-03
Fix make install-deps for glide 0.12x

### LINKED ISSUE #132: Update to latest glide, update vendor dependencies and add debugging support
Glide `0.12.x` offers performance improvements when installing/updating the vendor dependency graph based on caching.
- As part of containerizing the developer toolchain, `make install-deps` runs Glide in a container. Update [appcelerator/gotools](https://github.com/appcelerator/docker-gotools) to the latest version of Glide
- Update the [amp](https://github.com/appcelerator/amp) project dependencies.
- Add debugging support to the Makefile to help identify installation issues by passing `--debug` as an option to glide.

subfuzion:
Closes #132

This PR takes advantage of the performance enhancements of glide `0.12.x` with respect to vendor caching. This PR does not fix the file ownership issue for files created on a mounted volume by tools running in a container.

The [appcelerator/gotools](https://github.com/appcelerator/docker-gotools) image was [updated](https://github.com/appcelerator/docker-gotools/commit/468bc6fb89ff74697ed11aad67a87497d4808606) to Glide `0.12.1`.

When running `make install-deps` (or `make update-deps`), glide runs in a container based on an image that is using glide `0.12.1`. To take advantage of the performance benefits of vendor caching for development, the make rules currently mount the glide cache from the current host machine locations:
- `$HOME/.glide`
- `./.glide` (ie, the working directory)
## Warning

If these caches were populated with a different version of glide, then there is a possibility of glide corruption when running the version in the container. If `make install-deps` fails, it may not be enough to run `glide cc` to clear these caches. It may be necessary to physically delete the directories:
    $ rm -rf ~/.glide .glide
## Debugging

This PR also adds the ability to debug glide issues when running make:

```
$ GLIDE_OPTS=--debug make install-deps
```

This is slow, but prints out very useful information to help identify glide problems.
## Verification

```
$ rm -rf ~/.glide .glide ./amp ./amplifier
$ make install-deps
$ make build
```

Confirm no errors and `./amp` and `./amplifier` were built.


chrisccoy:
After further review, the verification steps worked as instructed.  The PR looks good, except the permissions appear to be an issue still when attempting to switch branches. 

```
drwxr-xr-x 3 root      root          4096 Sep  3 09:02 .glide
drwxr-xr-x 6 root      root          4096 Sep  3 09:20 vendor
-rwxr-xr-x 1 root      root      13653027 Sep  3 09:21 amp
drwxrwxr-x 9 chrisccoy chrisccoy     4096 Sep  3 09:21 .
-rwxr-xr-x 1 root      root      19638603 Sep  3 09:21 amplifier
```

subfuzion:
@chrisccoy I'm going to submit it as a separate PR since I prefer one fix per PR, but also because it's a more comprehensive fix that affects not only vendored files, but everything else that gets generated on a mounted volume by a tool running in a container. I have a fix in place and will submit separately. I'm going to close this PR since the glide part worked.

PR #139, opened 2016-09-05
Fix telegraf nats url

ndegory:
nats url (input and output) for telegraf were not the good ones


freignat91:
Tested ok for this issue, I merged the PR
but I got issues on 'amp stats' not all the stats are there, to be fixed in another PR

PR #142, opened 2016-09-05
Adding task_id and task_name fields to log entry

bquenin:
Adding extra fields to log entry
## Verification

```
make
./swarm start --min
make test-logs
```


freignat91:
tested ok

PR #146, opened 2016-09-06
Restore kafka

bquenin:
## Verification

```
$ ./swarm pull
$ ./swarm start
$ make test-logs
```

### DIFF
--- a/api/rpc/build/build_test.go
+++ b/api/rpc/build/build_test.go
@@ -22,7 +22,6 @@ const (
 	elasticsearchDefaultURL = "http://localhost:9200"
 	kafkaDefaultURL         = "localhost:9092"
 	influxDefaultURL        = "http://localhost:8086"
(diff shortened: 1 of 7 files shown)


chrisccoy:
Overall, looks good , but I am having a dns lookup problem on this step:

```
=== RUN   TestShouldStreamLogs
--- FAIL: TestShouldStreamLogs (0.10s)
```

This appears to be isolated to my docker setup and `/etc/resolv.conf` and unrelated to PR

subfuzion:
@chrisccoy That's because you need to run `sudo ./swarm start`, not just `./swarm start` (as the description says) so that @ndegory's `advertise_service` script can update your `/etc/hosts` file. You will also need to run `sudo ./swarm stop` when finished.

subfuzion:
Tests pass, LGTM.

PR #148, opened 2016-09-08
gRPC Service for Project data storage

chrisccoy:
This is an initial submission with limited scope to address the conversion of the CI storage from mongo to etcd via gRPC service. To start, this addresses only the "project" data model and its storage within etcd.  
## Verification

`sudo ./swarm start`
`# from amp root dir`
`go test -v ./api/rpc/project`
## TODO
- [ ] Needs to be added to amplifier. Current tests are not exercising remote calls, but are using gRPC/proto
- [ ] Tests should be added to Makefile, once solution matures.

### DIFF
--- a/api/rpc/project/project.go
+++ b/api/rpc/project/project.go
@@ -43,7 +43,7 @@ func (p *Proj) Update(ctx context.Context, req *ProjectRequest) (*ProjectReply, 
 func (p *Proj) Get(ctx context.Context, req *ProjectRequest) (*ProjectReply, error) {
 	// Build The Key for the Project
 	key := fmt.Sprintf("%s/%v", prefix, req.Project.RepoId)
-	reply := &ProjectReply{Project: &Project{}}
+	reply := &ProjectReply{Project: &ProjectEntry{}}
 	// Retrieve the Value for the Given Key
 	err := p.Store.Get(ctx, key, reply.Project, true)
 	return reply, err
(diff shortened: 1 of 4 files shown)


api/rpc/project/project.go
```diff
@@ -1,40 +1,105 @@
...
+)
+
+// Proj structure to implement ProjServer interface
+type Proj struct {
+	Store storage.Interface
+}
+
+// Create adds a new entry to the k,v data store
+func (p *Proj) Create(ctx context.Context, req *Project) *ProjectReply {
+	ctx, cancel := context.WithTimeout(context.Background(), defTimeout)
```
bquenin: I like the timeout concept, however, since we delegate the processing to a third party (in this case etcd), i would let this underlying library time out by itself and return the error as is. That would be a different story if we performed a long running operation ourselves.

api/rpc/project/project.proto
```diff
@@ -2,15 +2,30 @@ syntax = "proto3";
 
 package project;
 
-service Project {
-  rpc Create (CreateRequest) returns (CreateReply) {}
+service ProjectSvc {
+  rpc Create (Project) returns (ProjectReply) {}
```
bquenin: In order to follow the RPC style, I would use a "CreateRequest" that includes a "Project" instead of directly passing the project as input parameter. In this case, you can update the creation request without breaking compatibility (and just add new fields).
The same applies for other RPC calls.

api/rpc/project/project.proto
```diff
@@ -2,15 +2,30 @@ syntax = "proto3";
 
 package project;
 
-service Project {
-  rpc Create (CreateRequest) returns (CreateReply) {}
+service ProjectSvc {
```
bquenin: In order to be consistent with other proto files, I would name the service "Project".

api/rpc/project/project.proto
```diff
@@ -2,15 +2,30 @@ syntax = "proto3";
...
+message Key {
+  int32 repo_id=1;
+}
+message Empty{}
 
-message CreateReply {
-  string message = 1;
+message ProjectReply {
+    Project project=1;
+    string error_msg=2;
```
bquenin: I think errors should not be included in messages since gRPC provides an error mechanism for each RPC calls already.

bquenin:
Tested locally and working.

api/rpc/project/project.go
chrisccoy: removed, deferring to the backend

generalhenry (approved):

subfuzion:
Will reopen when we're ready to focus on build again.

PR #149, opened 2016-09-08
update group by to get sum not avg

freignat91:
amp stats was returning stats for nodes with avg values of the containers
this update make amp stats returns stats with sum of the containers

test:
- ./amp stats --mem - > get stats memory for local node
- ./amp stats --containers --mem -> get stats memory by containers
- the sum of all container used memory should be about equal to the memory used for the node.


subfuzion:
LGTM

bquenin:
tested ok

PR #151, opened 2016-09-08
fix stats computing issue

freignat91:
Fix computing stat issues

Test: 
- execute ./amp stats --container
- execute docker stats
- we should have about the same numbers


freignat91:
test ok

PR #160, opened 2016-09-14
Stack CLI / RPC support for up command

subfuzion:
Relates to #162 

Adds the `amp stack up` command and RPC support for up. The CLI command currently only supports the `-f, --file` option (and it's currently not optional!).
## Verification

Create a sample config file. It doesn't matter what's in if for now:

```
$ touch sample-config.yml

$ make install
# start swarm and amplifier
$ amp stack up -f sample-config.yml
```

Confirm an ID is printed on the terminal.


chrisccoy:
:+1:

PR #165, opened 2016-09-15
Stack protobuf storage

### LINKED ISSUE #164: Stack Protobuf storage
relates to https://github.com/appcelerator/amp/issues/162
Store stacks as protocol buffers

follow on to https://github.com/appcelerator/amp/issues/161

generalhenry:
closes https://github.com/appcelerator/amp/issues/164

Defines services and stacks as in the stack.proto, casts parsed format into them and stores them on stack create (and up)


api/rpc/stack/parse.go
```diff
@@ -9,23 +9,13 @@ import (
 type serviceMap struct {
 	Image       string      `yaml:"image"`
 	Ports       []string    `yaml:"ports"`
-	Replicas    int         `yaml:"replicas"`
+	Replicas    int32       `yaml:"replicas"`
```
subfuzion: I know it seems excessively large (I thought so as well), but the actual type used by docker for `replicas` is a `uint64`, and it's what I'm using for service right now.

api/rpc/stack/parse.go
```diff
@@ -9,23 +9,13 @@ import (
...
-	Ports       map[nat.Port][]nat.PortBinding
-	Replicas    int
-	Environment map[string]string
-	Labels      map[string]string
-}
-
-func parseStackYaml(in string) (out []Service, err error) {
-	out = []Service{}
+func parseStackYaml(in string) (out Stack, err error) {
+	out = Stack{}
```
subfuzion: Minor thing, but I'd save the value copying for the return and do this instead;  `out = &Stack{}`

api/rpc/stack/stack.go
```diff
@@ -7,34 +7,37 @@ import (
 	"golang.org/x/net/context"
 )
 
-// Stack is used to implement stack.StackServer
-type Stack struct {
+// Server is used to implement stack.Server
+type Server struct {
```
subfuzion, line 10: I definitely don't prefer calling it `Server`. Conflict with some other type, like a message?

subfuzion (changes requested):
Looks good overall, but a few questions or things to address...

api/rpc/stack/parse.go
generalhenry: done

api/rpc/stack/parse.go
generalhenry: done

api/rpc/stack/stack.go
subfuzion: I got the rationale, thanks.

subfuzion (approved):

subfuzion:
LGTM, thanks 👍

Issue #168, opened 2016-09-15
Add and resolve vendor dependencies for docker and protobuf


subfuzion:
There is a vendor dependency ordering issue with glide that is causing it to get the wrong versions of the following dependencies:
- github.com/docker/docker
- github.com/golang/protobuf

PR #169, opened 2016-09-15
fixing/adding/updating deps

### LINKED ISSUE #166: missing vendor dependency
github.com/docker/go-connections/nat
missing from the stack yaml parser

subfuzion:
Fixes #166, #168 

Fixes the broken build, plus fixes issues related to the following dependencies:
- github.com/docker/docker

This fix removes the version tag `^1.12.1` so that the latest version is fetched from `master` in the source repo. This is because `engine-api` has been deprecated and the packages are now part of `docker/docker`. This fix also updates the sub-package mapping because `types` moved to `api/types`.
- github.com/golang/protobuf

This fix handles an error cause by the `protobuf` repo's lack of semantic versioning. Based on dependency ordering, glide was populating its cache with an older version of protobuf that didn't support version 2. The fix was to manually update `glide.lock` and force it to use the latest commit (currently: https://github.com/golang/protobuf/tree/1f49d83d9aa00e6ce4fc8258c71cc7786aec968a).
## Verification

```
$ glide cc && rm -rf vendor
$ glide install
$ make install
```


subfuzion:
Because the build is broken for `master`, I'm going to merge my own PR.

PR #171, opened 2016-09-15
Removing "from" option because it doesn't make sense

bquenin:
## Verification

```
$ sudo ./swam start
$ make test-logs
```


subfuzion:
Elaborate?

bquenin:
The "from" parameter is only use for pagination purposes which aren't used from the CLI.
This parameter was added for a potential amp web UI but isn't used anyway and doesn't work with the CLI.
Needs to be rethought when we work on a web UI, which will probably be using streaming instead of pagination anyway.

subfuzion:
ah, makes sense. thanks.

Issue #162, opened 2016-09-14
Stack service


subfuzion:
Our initial stack implementation will support the following command:

```
$ amp stack up -f stackfile.yml
```

The CLI will call the stack rpc up operation and supply the stackfile. The service will be responsible for orchestrating the following activities:
- parse the stackfile and store the configuration information in `etcd`
- create and start the services specified in the stackfile
- update the reverse proxy service with any public mapping rules

The process of creating and starting the stack will result in various updates to state that need to be stored. Ultimately, all stack operations need to go through a state machine that ensures valid transitions and updates the current state in etcd.

The stack service should return the ID assigned to the stack. The stack ID can be used to subsequently query state (`amp stack inspect <identifier>`, where `identifier` is either the name or UUID (either short or long).

After the initial implementation, there will be support for additional options, such as name (`-n, --name`), and the ability to block the command until the async operation has finished (`--sync`).

Another feature will be a command to follow the state changes of the stack. This will be analogous to the `amp status` command, but specifically filtered for the stack and will include the option to follow the update stream: `amp stack status [-f, --follow] <identifier>`).

subfuzion:
(this is a wip)

Stack states
- Starting
- Running
- Stopped (initial)
- Redeploying

Transitions

| State / Next State | Starting | Running | Stopped | Redeploying |
| --- | --- | --- | --- | --- |
| Starting |  | started |  |  |
| Running |  |  | stop | redeploy |
| Stopped (initial) | start |  |  | redeploy |
| Redeploying | on redeploy success |  | on redeploy error |  |

PR #175, opened 2016-09-15
Initial stack states implementation

bquenin:
Related to #162 
## Verification

```
$ sudo ./swarm init
$ make test-stack
```

### DIFF
--- a/api/rpc/stack/stack-state.go
+++ b/api/rpc/stack/stack-state.go
@@ -2,105 +2,84 @@ package stack
 
 import (
 	"fmt"
-	"sync"
 )
 
-// State holds the current stack state, and has methods to get and
-// set the state.
-type State struct {
-	sync.Mutex
-	starting    bool
-	running     bool
-	stopped     bool
-	redeploying bool
+type State int
+
+const (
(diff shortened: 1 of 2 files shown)


api/rpc/stack/stack-state.go
```diff
@@ -0,0 +1,106 @@
+package stack
+
+import (
+	"fmt"
+	"sync"
+)
+
+// State holds the current stack state, and has methods to get and
```
generalhenry: Shouldn't the state be stored in etcd?

bquenin: It definitely should! :)

generalhenry (approved):

api/rpc/stack/stack-state.go
subfuzion: hehe, we already had that conversation out of band.

subfuzion (commented):
Per our earlier conversation, goals include
- [x] convert to a state transition table
- [x] remove lock: we'll synchronize on etcd
- [ ] always query etcd state for current state before validating/setting new state transition
- [ ] provide ability to supply the transition table

PR #179, opened 2016-09-16
Make backend services accessible

subfuzion:
This PR moves backend objects out of server to avoid cycles for dependents.
## Verification

make install, start amplifier, make test


generalhenry (approved):

bquenin (approved):
Tested and working

PR #180, opened 2016-09-16
propagate registry auth on all Swarm nodes

ndegory:
When the Swarm cluster is on more than one node, private Docker images won't be pulled.
This PR adds the flag do the docker service create call so that nodes have the Docker registry auth.


generalhenry (approved):
:+1:

subfuzion (approved):
LGTM

PR #181, opened 2016-09-16
Initial stack state implementation

bquenin:
Relates to #162 
## Verification

```
$ sudo ./swarm start
$ make test-stack
```


api/rpc/stack/stack-state.go
```diff
@@ -0,0 +1,98 @@
...
+	stack := &Stack{}
+	err := runtime.Store.Get(context.Background(), "stacks/"+s.Id, stack, false)
+	if err != nil {
+		return -1, err
+	}
+	return stack.State, nil
+}
+
+func (s *Stack) setState(state int32) error {
+	stack := &Stack{}
```
subfuzion, line 66: Technically we will need to do get/set atomically (part of a transaction). You don't want the state changing during the processing of this request. We can add it later, but please leave a TODO comment. (This is a general comment that applies to `transitionTo`, etc).

api/rpc/stack/stack-state.go
```diff
@@ -0,0 +1,98 @@
...
+	Starting
+	// Running  state
+	Running
+	// Redeploying  state
+	Redeploying
+)
+
+var stackRuleSet = RuleSet{
+	//   | Stopped   | Starting  | Running   | Redeploying
+	[]bool{false /**/, true /* */, false /**/, true /* */}, // Stopped (initial state)
```
generalhenry, line 23: looks like they should be arrays eg: `[4]bool{ false, true, false, true },`

bquenin: I'm not sure why slice is not enough ?

generalhenry: Slice will get the job done, array just makes it clearer.

subfuzion: If you want an array literal, you can let the compiler do the counting: `[...]bool`, but i think a slice literal is fine unless we want to explicitly fix the size.

api/rpc/stack/stack-state.go
subfuzion: The rest LGTM. Will merge after a comment is added.

subfuzion:
Nice work 👍

freignat91 (commented):
Need to use last version of service.ServiceSpec when available instead of stack.Service

subfuzion (approved):
Nice work 👍

PR #187, opened 2016-09-16
Adding CompareAndSet to store interface and make stack state transiti…

bquenin:
## Verification

```
$ sudo ./swarm start
$ make test-storage
$ make test-stack
```


api/rpc/stack/parse.go
```diff
@@ -3,9 +3,12 @@ package stack
 import (
 	"strings"
 
-	"gopkg.in/yaml.v2"
-
+	"context"
```
generalhenry: shouldn't that be "golang.org/x/net/context"

subfuzion:
Great work on the initial implementation. Will merge so it's in place for now for stack, and you can work on refactoring the implementation out of stack to create a generic state machine service that can be initialized with a state transition table and used by stack, service, etc.

👍

subfuzion (approved):
Great work on the initial implementation. Will merge so it's in place for now for stack, and you can work on refactoring the implementation out of stack to create a generic state machine service that can be initialized with a state transition table and used by stack, service, etc.

👍

api/rpc/stack/parse.go
subfuzion: context is in core now: https://golang.org/pkg/context/

PR #188, opened 2016-09-17
Stack services

freignat91:
Test: 
amp stack up test -f stack.yaml

where stack.yaml is for instance

pinger:
  image: appcelerator/pinger
  replicas: 2

We have service pinger available with 2 instances and service structs save in etcd


data/storage/etcd/store_test.go
```diff
@@ -70,7 +70,7 @@ func TestCreate(t *testing.T) {
 func TestGet(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), defTimeout)
 	key := "foo"
-	out := &storage.Project{}
+	out := &storagerecursive.Project{}
```
subfuzion: This isn't right... (won't build).

subfuzion (approved):
👍

PR #191, opened 2016-09-19
Stack parse tests

subfuzion:
This PR adds parse tests for stackfiles.

To use these new parse tests:
1. `Add <stackfile>.yml` files to the `api/rpc/stack/test_samples/`
2. Add an expected `serviceMap` map to the `parse_test.go` var section
3. Create a mapping between the filename and the expected serviceMap (`compareStructs`)
4. Update the comparison logic, if necessary (for example, `labels` aren't supported in this PR.
## Verification

```
$ go test -v -run TestSamples github.com/appcelerator/amp/api/rpc/stack
```


freignat91 (approved):
LGTM

freignat91 (approved):
LGTM

PR #193, opened 2016-09-19
add docker socket mount on amplifier

ndegory:
amplifier needs access to the docker socket for amp CLI commands (service and stack)


subfuzion (approved):
👍

PR #195, opened 2016-09-19
disable client side auth check

### LINKED ISSUE #194: Disable client side auth check
Currently running `amp stats . . .` without logging in first results in a `Requires login` error, however none of the current API endpoints require login.

As such the client side check should be skipped and an empty string should be sent as the auth token.

generalhenry:
closes https://github.com/appcelerator/amp/issues/194

to test:
- remove ~/.amp.yaml
- run `amp stats`


subfuzion (approved):
👍

PR #196, opened 2016-09-19
add stack stop and remove

freignat91:
add amp command, 
- amp stack stop [stackId]
- amp stack remove [stackId]

test:  make test-stack


subfuzion (approved):
There are a few haproxy issues running on localhost, but we can iterate on those. The rest LGTM.

PR #206, opened 2016-09-20
Stack name

freignat91:
add "amp stack ls"
update amp stack stop, rm to handle both id and name
fixe some issues

test:  make test-stack


api/rpc/stack/parse_test.go
```diff
@@ -144,20 +146,25 @@ func parse(t *testing.T, test *TestSpec) {
 
 func (a serviceMap) compare(b serviceMap) bool {
 	if a.Image != b.Image {
+		fmt.Printf("Images don't match: %v != %v\n", a.Image, b.Image)
```
subfuzion: It's better to use `t.Log`/`t.Logf` since these print to stderr and since the output is in order and where you expect to see it when it's printed as the test runs. You can add a `t *testing.T` arg to compare and pass it from the calling function.

subfuzion (commented):
Very trivial comment on using `t.Logf` instead of `fmt.Println`. Rest looks good. 👍

api/rpc/stack/parse_test.go
freignat91: ok done

bquenin (approved):
Tested locally and working.

subfuzion:
👍

PR #216, opened 2016-09-21
improve stats command line

freignat91:
amp stats default is now "service"
amp stats [serviceName] is equivalent to amp stats --service-name=[serviceName]

we can use part of the service name
'amp stats amp' will return all services starting by 'amp'
same behavior with arguments:
- service-id
- task-id
- task-name
- container-id
- container-name
- node-id

added stats tests

test:  make test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,4 +130,4 @@ test:
 	@go test $(REPO)/data/storage/etcd
 	@go test $(REPO)/api/rpc/stack
 	@go test $(REPO)/data/influx
-#	@go test $(REPO)/api/rpc/stat
+	@go test $(REPO)/api/rpc/stats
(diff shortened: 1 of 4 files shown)


bquenin (commented):
Working for service name but not for service id

generalhenry (changes requested):
When I `make test` locally I encounter a number of "Unexpected empty answer from server" errors

freignat91:
@Henry, verify that Grafana show stats in its dashboard without stats the 'make test' will fail'

freignat91:
@bertrand, this pull request doesn't address service-id in its description. It's only for service-name

bquenin (approved):
Tested locally and working

generalhenry (approved):

PR #217, opened 2016-09-21
use amp stats serviceName or Id

freignat91:
we can now use:
amp stats [service name or service Id]
to perform stats on services with part of a service name or service id

test: make test


bquenin (approved):
Tested locally and working

PR #223, opened 2016-09-21
add stack state to stack ls

freignat91:
add a column 'state' to amp cmd: amp stack ls
have a better output with column format

test: make test


bquenin (approved):
Tested locally and working

PR #226, opened 2016-09-21
stack rollback on error

freignat91:
Rollback stack on error when bringing it up

prerequisite:
- have a [stack yaml file] with a service publishing a port

test:
- amp stack up test1 -f [stack yaml file]

The stack test1 is created
- amp stack up test2 -f [stack yaml file]

The stack test2 should failed because use the same port is published, then the test2 stack creation is rollbacked
- amp stack ls

Shows only the stack test1


chrisccoy (approved):
Performed as expected :+1:

PR #230, opened 2016-09-22
Stack rm q

freignat91:
add command: amp ls -q
to list the available stack id
can be use with amp rm to remove all stacks using:
amp stack rm -f $(amp stack ls -q)

added command amp info to display amp version
amp version is not anymore displayed at each command (not compatible with grep usage)

test: make test


bquenin (approved):
Tested locally and working

Issue #232, opened 2016-09-22
Logs by stack


craigivy:
We would like the ability to see the logs based the stack we are running from the amp logs command line.

freignat91:
amp-agent has been updated to add stack-id and stack-name in log using labels added in #237 and #236 allows to extract them.
To be tested.

PR #236, opened 2016-09-22
Adding stack_id and stack_name to LogEntry

bquenin:
Contributes to #232 
## Verification

```
$ ./swarm start
$ make test
```


freignat91 (approved):
Tested ok

PR #237, opened 2016-09-23
add stack-id and stack-name as labels in containers

freignat91:
related to #232
add stack id and stack name as labels in container to allow amp-agent to add them in logs

test:
- make, start swam amp
- create a stack: amp stack up test -f [stack file]
- verify using docker inspect than containers created for stack services have labels: io.amp.stack.id and io.amp.stack.name

regression tests: make test


bquenin (approved):
Tested locally and working

Issue #235, opened 2016-09-22
Service Env


generalhenry:
Env is implemented by the docker swarm API as a `[]string`

1) Change the ServiceSpec for env to `repeated string`
2) Update the parser to use `[]string`
3) Pass Env to containerSpec

freignat91:
@Henry. Our serviceSpect was already not the same than the docker one. I let Tony decide what to do with this PR, modify the parsing to have []string (the docker type) to be adapted to []*string (the repeated string type) whatever, or use the one the parsing already give now: map[string]string, this PR

subfuzion:
@freignat91 See my comment here: https://github.com/appcelerator/amp/pull/238#issuecomment-249398791. I've gone ahead and submitted a separate PR that also includes some other cleanup/refactoring (#244).

PR #238, opened 2016-09-23
add env in service

freignat91:
#235

add env variable in service creation
translated the map[string]string of the service.ptoto to []string of swarm.ServiceSpec

test: make test

### DIFF
--- a/api/rpc/service/service.go
+++ b/api/rpc/service/service.go
@@ -92,6 +92,15 @@ func CreateService(docker *client.Client, ctx context.Context, req *ServiceCreat
 		},
 		EndpointSpec: nil, // &EndpointSpec
 	}
+	//add env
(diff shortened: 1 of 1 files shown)


generalhenry:
This handles part three of https://github.com/appcelerator/amp/issues/235 but skips part one and two.  I don't think we should store stacks in any format other than the format that the docker service API expects.

subfuzion:
@freignat91 I think it's more natural to express it as a map as well, which is why I implemented it that way with the service spec. However, it does seem to be a waste that we parse the CLI environment string entries to create a map, only to subsequently convert the map back to an array of joined strings to pass to Docker. Because there was other refactoring I wanted to do, I create an entire PR for everything (#244), so I'm going to go ahead and close this one. Thanks.

PR #243, opened 2016-09-25
Cleaned up etcd test

subfuzion:
Cleaned up the test for `data/storage/etcd` by moving the test protobuf definition into the same directory and renaming it from "Project" to "TestMessage" to make its purpose clearer.
## Verification

```
$ sudo ./swarm start
$ make test
```


freignat91 (approved):
Tested ok

PR #246, opened 2016-09-25
Removed grafana and kibana from min swarm

subfuzion:
## Verification

```
$ sudo ./swarm start --min
```


freignat91 (approved):
Tested ok

PR #251, opened 2016-09-25
Container labels

subfuzion:
Closes #250 
## Verification

Merge #249 first.

```
$ make test

$ make install
# start amplifier
$ amp service create --name pinger --container-label foo=bar --container-label hello=world appcelerator/pinger
$ docker inspect -f '{{ .Spec.ContainerSpec.Labels }}' pinger
$ docker service rm pinger
```

PR #253, opened 2016-09-26
Service rm

### LINKED ISSUE #252: Add service rm
Add a CLI command to be able to remove a service.

subfuzion:
Closes #252 
## Verification

Merge #251 first.

```
$ make test

$ make install
$ amp service create --name pinger appcelerator/pinger
$ amp service rm pinger

$ amp service create --name pinger1 appcelerator/pinger
$ amp service create --name pinger2 appcelerator/pinger
$ amp service rm pinger1 pinger2

$ amp service create --name pinger1 appcelerator/pinger
$ amp service rm pinger1 pinger2
# verify pinger1 removed, good error for pinger2
```


freignat91 (approved):
Tested ok

api/rpc/service/service.proto
```diff
@@ -35,3 +18,25 @@ message PublishSpec {
...
+message ServiceCreateRequest {
+  ServiceSpec service_spec = 1;
+}
+
+message ServiceCreateResponse {
+  string id = 1;
+}
+
+message RemoveRequest {
+  string ident = 1;
```
generalhenry, line 31: why ident instead of id?

chrisccoy:
Looks good, needs rebase

api/rpc/service/service.proto
subfuzion: making it clear it can be an ID, short ID, or name. Actually, that was something that @freignat91 recommended. I have no problem calling it `id`, which matches docker docs.

generalhenry: Sounds fine, it could use a comment to make that clear.

follow up: should RemoveResponse return an ident instead of an id?

PR #256, opened 2016-09-27
Added support for replicated/global mode for service & stack

### LINKED ISSUE #255: Need support for replicated/global mode
Requirements
- `amp service create` should have support for `--mode replicated|global`
- `amp stack up` should have stackfile support for:
  `mode: replicated|global`

Rules
- `mode` default = `replicated`
- if `replicas` > 0 && `mode` == "global" => error ("replicas can only be used with replicated mode")
- if `replicas` < 1 && `mode` == ""  => set `replicas` equal to 1

subfuzion:
Closes #255 
## Verification

```
$ make test

$ amp service create --name pinger appcelerator/pinger
# verify service created with 1/1 replicas
$ amp service rm pinger

$ amp service create --replicas 2 --mode replicated --name pinger appcelerator/pinger
# verify service created with 2/2 replicas
$ amp service rm pinger

$ amp service create --mode global --name pinger appcelerator/pinger
# verify service created as global
$ amp service rm pinger

$ amp service create --replicas 4 --mode global --name pinger appcelerator/pinger
# verify error: replicas can only be used with replicated mode
```


generalhenry (approved):

generalhenry:
the verify error resulted in a `2016/09/26 19:31:39 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp: context canceled"; Reconnecting to {"localhost:50101" <nil>}`

Which is no good, but I think it's pretty much unrelated to this PR

subfuzion:
@generalhenry ruh-roh. Let's see if @chrisccoy can duplicate. Thanks.

chrisccoy:
I also am seeing the issue reported by @generalhenry, however,  it is not every time. I believe it is the problem with the `/etc/hosts` entry. If the amplifier service does not come up clean, then I see the `2016/09/26 20:17:20 grpc: addrConn.resetTransport failed to create client transport:` issue, but I have had several clean starts and witnessed the test suite works as expected.

I believe that in certain instances, the swarm is restarting amplifier resulting in a bad entry in the hosts file. Not sure of an elegant approach to fix the issue, but this does appear to be pre-existing and the PR was successful.

subfuzion:
@chrisccoy Thanks for the analysis. We definitely need to investigate further. 👍

Issue #239, opened 2016-09-23
Stack network stage 1


freignat91:
Summarize of discussion we had on this subjects:

Stack networks:
- we have a default network created for a stack having the same name than the stack
- by default all services created in the stack joins this default network.
- each service in the stack can create/join new networks, using yaml syntax:
  networks:
  - networkName1
  - networkName2
- the real name of the created networks are prefixed by the stack name 
- for each stack a service haprroxy is created to handle the stack services reverse proxy specifically
- a default [stackName]-public network is created
- the service Haproxy joins the [stackName]-public network only (so HAproxy does't see a service if it's not in the [stackName]-public network)
- each service whiich  have to be reserve proxy (so having at least a -name value in the -public item) will join thie [stackName]-public network automatically (with nothing more to add in the service definition)

Service alias name:
- on each network a service joins, we create an alias name for the service using its short name. The real name of the service is composed with [stackName]-[shortName]

...

freignat91:
As a stage 1, i propose to implement:
- have an automated haproxy service created for each stack
- have default network created for each stack
- have public network created for each stack and make service automatically joins it if they have to be reverse proxy (if we have a -name in -public service definition)

make first that stable, before added multiple custom networks

PR #259, opened 2016-09-28
move docker connection to server

freignat91:
To prepare #239
docker connection was done on api/rpc/service
move it on api/server to be shared by service and stack

test: make test

### DIFF
--- a/api/rpc/stack/stack.go
+++ b/api/rpc/stack/stack.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/appcelerator/amp/api/rpc/service"
 	"github.com/appcelerator/amp/data/storage"
+	"github.com/docker/docker/client"
 	"github.com/golang/protobuf/proto"
 	"golang.org/x/net/context"
 )
@@ -19,7 +20,8 @@ const stackNameLabelName = "io.amp.stack.name"
 
 // Server is used to implement stack.StackService
 type Server struct {
-	Store storage.Interface
+	Store  storage.Interface
+	Docker *client.Client
 }
 
 // Up implements stack.ServerService Up
@@ -86,9 +88,14 @@ func (s *Server) getStack(ctx context.Context, in *StackRequest) (*Stack, error)
 // clean up if error happended during stack creation, delete all created services and all etcd data
 func (s *Server) rollbackServiceStack(ctx context.Context, stackID string, serviceIDList []string) {
 	fmt.Printf("removing created services %s\n", stackID)
+	server := service.Service{
+		Docker: s.Docker,
+	}
 	for _, ID := range serviceIDList {
 		if ID != "" {
-			service.Remove(ctx, ID)
+			server.Remove(ctx, &service.RemoveRequest{
+				Ident: ID,
+			})
 			s.Store.Delete(ctx, path.Join(servicesRootKey, ID), true, nil)
 		}
 	}
@@ -119,7 +126,9 @@ func (s *Server) processService(ctx context.Context, stack *Stack, serv *service
 	request := &service.ServiceCreateRequest{
 		ServiceSpec: serv,
 	}
-	server := service.Service{}
+	server := service.Service{
+		Docker: s.Docker,
+	}
 	reply, err := server.Create(ctx, request)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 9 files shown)


api/rpc/stack/stack.go
```diff
@@ -119,7 +126,9 @@ func (s *Server) processService(ctx context.Context, stack *Stack, serv *service
 	request := &service.ServiceCreateRequest{
 		ServiceSpec: serv,
 	}
-	server := service.Service{}
+	server := service.Service{
```
subfuzion, line 129: You probably don't need to do this (go through the protobuf interface). You already call [service.Remove](https://github.com/appcelerator/amp/blob/master/api/rpc/service/service.go#L160) directly, so you could probably call [service.Create](https://github.com/appcelerator/amp/blob/master/api/rpc/service/service.go#L61) directly as well. (We can refactor service.Create to be a bit more straightforward -- like service.Remove now or sometime later...)

cmd/amplifier/main.go
```diff
@@ -52,6 +56,8 @@ func parseFlags() {
 	flag.StringVarP(&clientSecret, "clientsecret", "c", defaultClientSecret, "github app clientsecret (default '"+defaultClientSecret+"')")
 	flag.StringVarP(&kafkaURL, "kafkaURL", "k", kafkaDefaultURL, "kafka URL (default '"+kafkaDefaultURL+"')")
 	flag.StringVarP(&influxURL, "influxURL", "", influxDefaultURL, "InfluxDB URL (default '"+influxDefaultURL+"')")
+	flag.StringVarP(&dockerURL, "dockerURL", "", dockerDefaultURL, "Docker URL (default '"+dockerDefaultURL+"')")
```
subfuzion: Don't use `*P` (`flag.StringVarP`) if you're not going to provide a short flag option and remove the corresponding empty `""` argument. Same with the next option (and also the same for the previous `influxURL` option (even though you didn't create that one).

Also, don't create options for linux with mixed case; that's not standard. Only use lowercase (and hyphens, if necessary). Ex: `--docker-version`.  However, I don't understand why we need to provide these options anyway. We only build and test for a specific version, so does it make sense to make the version an option? The version needs to be coordinated with the client. Also, why do we need docker url (will it ever be at a location other than `unix://var/run/docker.sock`)?

cmd/amplifier/main.go
```diff
@@ -18,6 +18,8 @@ var (
 	elasticsearchDefaultURL = "http://localhost:9200"
 	kafkaDefaultURL         = "localhost:9092"
 	influxDefaultURL        = "http://localhost:8086"
+	dockerDefaultURL        = "unix:///var/run/docker.sock"
```
subfuzion: Don't need to have `///` if you use backticks for [raw string literals](https://golang.org/ref/spec#String_literals) instead of normal quotes:

```
dockerDefaultURL = `unix://var/run/docker.sock`
```

cmd/amplifier/main.go
freignat91: ok, done

cmd/amplifier/main.go
freignat91: ok, done

api/rpc/stack/stack.go
freignat91: I removed the direct function Create and Remove to let only the (s _Service) Create and (s *Service) Remove
Because the direct functions (Create and Remove) need to use a Docker client as a local variable  (_client.Client) and this variable does't exist anymore. It is accessible only in the Service struct. 
So Stack use a Service struct to invoke (s *Service) Create and (s *Service) Remove, but it's as efficient as the direct function call.

cmd/amplifier/main.go
freignat91: For options lowercase, it's also true for kafakURL and influxURL, should have to lowercase them alo?

api/rpc/stack/stack.go
subfuzion: Ok, but I had a comment that I was going to be refactoring those helpers into a separate utility and you would be expected to supply the docker socket. That's why I intentionally had it at the file level and not in the service struct. I guess it's fine for now. I don't won't to hold this PR up, but I will be refactoring it again soon.

cmd/amplifier/main.go
subfuzion: Those should also be updated, but we don't have to hold up the PR for you. It can be done separately later.

subfuzion:
Thanks @freignat91

subfuzion (approved):
I plan on doing more refactoring here as well, but thanks for your updates.

PR #266, opened 2016-09-29
add service network attachment

freignat91:
#239 stage one preparation

add service networks attachment 

test:
- launch amp
- create a service attached to the network "amp-swarm" with two aliases: "essai1 and "essai2":
- amp service create appcelerator/pinger:latest --name essai --network amp-swarm:essai1:essai2
- inspect the service essai container to verify that the amp-swarm network is well in the Networks container setting
- Verify using amp-agent (which is on amp-swarm also) than essai service is visible:
- docker exec -it [amp-agent container id] sh
- ping essai: returns a valid IP address
- ping essai1: returns the same IP
- ping essai2: returns the same IP
- ping essai3: returns dns error, invalid address

regression test:
- make test


subfuzion (approved):
Great work, @freignat91. Looks good and tests pass. 👍

PR #267, opened 2016-09-29
the swarm script can be used standalone (not the full repo)

ndegory:
related to #264
don't try to execute the advertise_service script if it's not present


subfuzion (approved):
Good! 👍

Issue #260, opened 2016-09-28
Solution for localhost DNS with no prerequisite, all systems (linux, windows, mac)


freignat91:
[fist post re-written]:
Create a DNS entry in Route53  format *.local.[domainName] and having 127.0.0.1 as value
So any host names matching the form *.local.[domaineName] will be resolved to 127.0.0.1 on all machines.
This way we can use Haproxy reserve proxy using host with a local instance of amp running on our machine for test or development. All URL having host finishing by `local.[domainName]` will be routed to the local haproxy and processed.

freignat91:
I just created DNS entry: "local.axwaytest.net"
so xxx.local.axwaytest.net routes to localhost on all machines, no matter the what is 'xxx'
ui.local.axwaytest.net on your explorer will launch your local instance of amp-ui for instance

generalhenry:
I also set *.local.appcelerator.io to localhost.

subfuzion:
Modifying `etc/hosts` always works to provide a way to map host names. Although `/etc/hosts` wasn't intended for DNS, it works just fine for not only hostnames but multilevel domain names. Mapping dedicated names (`foo.appcelerator.io` to `127.0.01`, for example) works as long as there is no conflict with any 3rd level domain name. As per Henry's example, using `*.local.appcelerator.io` as a 4th level wildcard mapping for local.appcelerator.io is safer.

Ideally, we just want a simple tool to do to it in a safe reliable way that doesn't require the user to manually edit their hosts file (something a bit more robust than our current shell script).

freignat91:
Not sure i got your last point @Tony, it's what we propose with "local.axwaytest.net" or "local.appcelerator.io". 
Anyone executing `ping xxx.local.appcelerator.io` on his machine anywhere will see answer: 127.0.0.1
with nothing to do on the machine. It's safe and robust.
I updated the first post which wasn't very clear.

PR #270, opened 2016-09-30
Stack network stage 1

freignat91:
related to #239 

when creating a stack a default network is created having the same name than the stack
- each service of the stack is attached to this network and have an alias name equal to their short Name (the name declared in the stack file)
- each service of the stack has a public name equal to [stackNalme]-[shortName], the name visible on `docker sevice ls` command

if it exists at least one stack service having at least one publish name (PublishSpec.Name != "") then 
- a network [stackName]-public is also created.
- a haproxy service is added in the stack attached to the network [stackName]-public
- each stack service having at least one PublishSpec.Name != "" is  also attached on [stackName]-public network
- each PublishSpec.Name is an entry in the stack Haproxy service to reverse-proxy the service for each http request having a hostname matching the form: [PublishSpec.Name].*
- an entry is added to the main haproxy (the global one, not attached to a stack) in order to reverseproxy any http request matching the form:   *.[stackName].* to the right stack haproxy service.

see #239 (stage 1) description

test:
- stop your amp: sudo ./swarm stop
- make install (after rm -rf vendor and glide install)
- pull all images: ./swarm pull  (you should pull appcelerator/haproxy:1.0.1
- start your amp: sudo ./swarm start
- launch a local amplifier
- create stack : amp stack up test -f api/rpc/stack/test_samples/sample-03.yml

verify:
- the stack is created (amp command doesn't return error, tthe amplifier log shows no error)
- the services test-pinger and test.pinger2 are vislbe using command: docker service ls
- two new networks have been created: test and test-local, using command docker network ls
- the command: `curl www.test.local.axwaytest.net/ping` returns pong

then, delete the stack: amp stack rm -f test
verify:
- the two stack services are not visible anymore using: docker service ls
- the two stack networks have been deleted, using docker network ls

regression test:
- make test

PS: the amp-ui has been updated to be displayed using host name only (ui.local.axwaytest.net or ui.local.appcelerator.io)


subfuzion:
@freignat91 You should update the swarm script images to pull the specific version of `haproxy` that you want [here](https://github.com/appcelerator/amp/blob/master/swarm#L24) (ie, `appcelerator/haproxy:1.0.1`)

subfuzion:
Please update error message:

```
no serveur found for stack=local service=sample
```

to

```
no server found for stack=local service=sample
```

subfuzion:
@freignat91 It works for me, but I'm not seeing why we want to expose this through a URL, or why we're exposing it on the `stack` command line instead of a stackfile (however, exposing the option on `service` makes sense).

freignat91:
I don't get your last comment. 
It works like before there is no change expect that the host name required to reverse proxy a service is composed by [name].[stack].[domain], in our sample [www].[test].[local.axwaytest.net]
Here the domain `local.axwaytest.net` routes to localhost (as `local.appcelerator.io` created by Henry does, see #260).
So `curl www.test.local.appcelerator.io/ping` works also.
The amp stack commands didn't change, Only the way we process stack file change to create the right networks, attached the right services and to launch HAProxy with the right setting if needed.
All the needed information are in the stack file

freignat91:
I just see that i make a mistake in the PR description. The command to test the servce pinger reverse proxy is  not:
 www.test.local.axwaytest.net/ping
but
curl www.test.local.axwaytest.net/ping

or you can use also an explorer un enter "www.test.local.axwaytest.net/ping" in it.

freignat91:
I updated the sample03.yml stack file in order to make the stack haproxy service publish the port 83.

```
haproxy:
  public:
    - publish_port: 83
      internal_port: 80
```

Then it's possible to use directly the stack haproxy service like that:
`curl www.local.axwaytest.net:83/ping`
when using directly the stack haproxy service, it's not needed to specify the stack name in the host.

So to use the rest api url [restApiUrl] of a service in a stack [stackName], through its published public name [name], two ways;
- `curl [name].[stackName].[domainName]/[restApiUrl]` (using the global haproxy service on port 80)
- `curl [name].[domainName]:[port]/[restApiUrl]` (using the stack haproxy service publishing the port [port])

api/rpc/stack/test_samples/sample-03.yml
```diff
@@ -8,3 +8,8 @@ pinger2:
     - name: www
       protocol: tcp
       internal_port: 3000
+haproxy:
```
subfuzion, line 11: It's better to create new sample files for new features.

subfuzion (approved):
@freignat91 Worked perfectly! (I added `*.local.appcelerator.io` to my `/etc/hosts` file). Nice work! 👍

PR #281, opened 2016-10-04
Volume mounts for registry and etcd (#278)

### LINKED ISSUE #278: Add Persistence to etcd container
# Background

When the swarm restarts, stacks and services that have been previously created are left in an orphaned state, rendering amp cli unable to stop and remove.
# Solution

Add a mount to the etcd swarm command that will cause the etcd `<k,v>` to persist between restarts. 

For example,

```
 --mount type=bind,source=/etcd/data,target=/etcd-data \
    -e ETCD_DATA_DIR=/etcd-data \
```

Thoughts?

ndegory:
resolves #278
- replace bind mount of registry service by a volume mount
- add a volume mount on etcd service

### DIFF
--- a/swarm
+++ b/swarm
@@ -265,6 +265,7 @@ etcd() { # Owner: ndegory
     --label io.amp.role="infrastructure" \
     -p 2379:2379 \
     -p 2380:2380 \
+    --mount type=volume,source=amp-etcd,target=/data \
     appcelerator/etcd:${ETCD_VERSION} \
       --name etcd \
       --listen-client-urls http://0.0.0.0:2379 \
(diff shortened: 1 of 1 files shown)


chrisccoy (approved):
- Started the Swarm
- Created a Stack
- Restarted the Swarm
  
  Stack was still visible in  etcd
- Stopped Stack
- Removed Stack
  
  Previously, this was not possible. 

As long as everyone is on board with using the volume approach, LGTM.

ndegory:
I also checked with the registry service:
- swarm start --min
- docker push localhost:5000/whatever-image
- docker ps | grep registry
- docker kill _registry-container-id_
- wait for the service to bring back a new container
- check that the image is still in the registry: curl localhost:5000/v2/_catalog

chrisccoy:
:+1:

subfuzion:
LGTM 👍

PR #285, opened 2016-10-05
Creating ~/registry/data folder in order to avoid log flooding

bquenin:
# Verification

```
$ rm -rf ~/registry/data
$ ./swarm start
```

Verify that the `~/registry/data` folder is created.

### DIFF
--- a/swarm
+++ b/swarm
@@ -117,6 +117,7 @@ main() {
   else
     case $1 in
       start)
+        mkdir -p ~/registry/data
         initswarm
         createnetwork
         startservices "${@:2}"


freignat91 (approved):
LGTM

ndegory:
LNGTM

ndegory:
the registry service is now using a named volume (#281), so no need of a mount point on the host

bquenin:
Ok, reverting with #286

subfuzion:
@bertrand-quenin @ndegory @freignat91 The title of this PR alone is not a good enough explanation of the problem nor why creating a directory helps. This is a good example of the kind of PR "rubber stamping" we want to avoid. We want to ensure that PRs will make sense to anyone else reviewing them.

PR #287, opened 2016-10-05
use the proper version for the Kafka image

ndegory:
Now that we build the image with the correct version, we can update the swarm script


bquenin (approved):
At last :)

PR #290, opened 2016-10-05
Fix formatting and add missing comments

bquenin:
## Verification

```
$ make
```

Make sure there are no comments on formatting or comments.


freignat91 (changes requested):
I would prefer these changes, on sevice, merged after the stack network PR#270 to avoid too complicated rebase.

freignat91 (approved):
LGTM

PR #297, opened 2016-10-10
remove the explicit volumes after removing the services

ndegory:
etcd and registry use named volumes.
if the volume exist at service start, it is reused.
to avoid that, this PR drops the volumes after removing the services.


freignat91 (approved):
tested ok

Issue #318, opened 2016-10-15
Use nats streaming as messaging system


bquenin:
Replace our kafka messaging system by nats streaming.

PR #319, opened 2016-10-15
Updating vendor for nats streaming

bquenin:
Related to #318 
## Verification

```
$ glide cc
$ rm -rf vendor
$ glide install
$ make
```


generalhenry (approved):

subfuzion:
@bertrand-quenin Please resubmit as a PR with *only* the vendor changes. Then, submit another PR with the code changes. Also, I realize we have different styles, but please don't make stylistic changes to my makefile without at least asking me why I did things a certain why or discussing why you prefer to change it, otherwise it just becomes a "style" war. We can coordinate on a makefile style guide if you feel strong about it (I do tend to have strong opinions regarding makefiles myself).

Because the diff is so hard to look at in GitHub (loading and scrolling via GitHub is excruciatingly slow), and because we've encountered so many issues with vendor updates, let's all make a point of making very clear what the change to `glide.yaml` is in the PR description to help the review.

bertrand-quenin:
@subfuzion I did not change anything on our Makefile, you're probably mistaking with the docker Makefile that has been updated

subfuzion:
Thanks @bertrand-quenin 👍

PR #327, opened 2016-10-17
Removed stack noise

### LINKED ISSUE #326: stack listing quite mode is noisy
With no stacks `amp stack ls -q` should have no output other than a zero return code. The same as `docker service ls -q`, `docker ps -q` etc . . .

Current output is `No stack is available`

generalhenry:
closes https://github.com/appcelerator/amp/issues/326

to test:
`make install`
`amp service ls -q`
expected no output


subfuzion (approved):
LGTM

PR #333, opened 2016-10-18
fix stack test

freignat91:
fix stack test regression

test: make test


bquenin (approved):
Tested locally and working

PR #335, opened 2016-10-18
Refactored state machine to use strings instead of integers

bquenin:
## Verification

```
$ ./swarm start
$ make test
```


freignat91 (approved):
Tested ok

Issue #346, opened 2016-10-19
Kapacitor Slack webhook is hardcoded in Swarm script


ndegory:
No private data should be committed in the public repo. The Slack webhook is one of them.

PR #347, opened 2016-10-19
Slack output for Kapacitor is now based on environment variables (#346)

ndegory:
- use environment variables for Slack output specifications


bquenin (approved):
LGTM

Issue #350, opened 2016-10-19
Need to address make check issues


subfuzion:
```
$
amp ((HEAD detached at v0.2.0))$ make check
./api/rpc/stack/stack.go
./api/rpc/stack/stack_test.go
/bin/bash: line 0: test: ./api/rpc/stack/stack.go: binary operator expected
[WARN] Fix formatting issues with 'make fmt'
/go/src/github.com/appcelerator/amp/api/rpc/stack/parse.go:67:1: exported function ParseStackfile should have comment or be unexported
/go/src/github.com/appcelerator/amp/api/rpc/stack/stack.go:224:27: should drop = 0 from declaration of var publishPort; it is the zero value
```

freignat91:
done in PR#352

PR #352, opened 2016-10-20
issue#350

freignat91:
address #350

test: make test


ndegory (approved):
LGTM, make check and make test are passing

PR #372, opened 2016-10-24
Use latest release of NATS streaming on master

bquenin:
Related to #318 
## Verification

```
$ ./swarm pull
$ ./swarm start
$ make test
```


subfuzion:
@bquenin Based on our experience, I think we're better off using explicit versions even on the `master` branch for deterministic behavior. I know that we want to track their development as closely as possible for the clustering features, but I would prefer that we update explicitly to `0.3.0` and that all version bumps go through the PR/test process.

bquenin:
@subfuzion sure. I changed it to the 0.3.0 release instead of latest.

subfuzion (approved):
👍

PR #373, opened 2016-10-24
Only publish docker images when building master branch on travis

bquenin:
## Verification
- Log in to travis and check that only the master branch deploys docker image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,23 +8,20 @@ language: go
 go:
   - 1.7.1
 
-before_install:
+install:
+  - go get -u github.com/golang/lint/golint
   - sudo apt-get update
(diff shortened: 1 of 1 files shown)


ndegory:
@bquenin what was the issue with pushing from non master branches?

bquenin:
@ndegory Cluttering our docker hub repo with one image per branch which is not what we expect (I assume)

subfuzion:
I agree in essence with what @bquenin wants to accomplish here specifically for `appcelerator/amp`. We can manually push any other tagged images we need until we have a fully automated release process. I don't see a use case right now that justifies having every feature PR push `appcelerator/amp:{feature-branch}`, but if one materializes, this restriction will be easy enough to remove.

PR #375, opened 2016-10-24
Ordered cli tests

ndegory:
fixes #374 
- add a `step` field in the yaml file
- sort the map entries after unmarshaling the yaml file

how to check:
- swarm start
- go test ./cmd/amp/cli
- if tests are passing, that's great. If not, check in the logs that 08-restart-stack.yml is executed in same order as specified

PR #383, opened 2016-10-25
Regex for Logs

neha-viswanathan:
Added regex for _logs_ CLI commands and updated the restart-stack.yml to use the correct command. The test however fails at times because of the order in which the commands are called. See #375 for further details.

To test
`go test ./cmd/amp/cli`


subfuzion (approved):
LGTM 👍

PR #386, opened 2016-10-26
Refactor CLI tests

bquenin:
Related to #370 

Note that the tests are not working as they were not working already. I just made some changes to be able to fix them in a subsequent PR.
## Verification

```
$ ./swarm start
$ go test -v github.com/appcelerator/amp/cmd/amp/cli
```

Issue #370, opened 2016-10-24
Integration tests


subfuzion:
We need to increase test coverage and extend integration tests to include networking, etc (see current integration tests [here](https://docs.google.com/spreadsheets/d/1i_fZ9ArzQwnePRPpudxk69f4_CIYk7WBMX73m1IOifY/edit?usp=sharing)).

Work on removing any redundancy with current backend integration tests (ie, those that require starting amplifier), that can be defined by yaml test specs and added [here](https://github.com/appcelerator/amp/tree/master/cmd/amp/cli/test_samples).

bquenin:
Do you mind if I submit a PR for renaming the test files from:

```
00-invalid-sample-01.yml
00-invalid-sample-02.yml
01-create-service.yml
02-list-service.yml
03-curl-service.yml
04-remove-service.yml
05-create-stack.yml
06-list-stack.yml
07-stop-stack.yml
08-restart-stack.yml
09-remove-stack.yml
10-cpu-stats.yml
11-mem-stats.yml
12-io-stats.yml
13-task-stats.yml
14-container-stats.yml
15-net-stats.yml
16-node-stats.yml
17-service-stats.yml
18-stack-logs.yml
19-numbered-logs.yml
20-metadata-logs.yml
21-all-logs.yml
```

to

```
00-invalid-sample-01.yml
00-invalid-sample-02.yml
01-service-create.yml
02-service-list.yml
03-service-curl.yml
04-service-remove.yml
05-stack-create.yml
06-stack-list.yml
07-stack-stop.yml
08-stack-restart.yml
09-stack-remove.yml
10-stats-cpu.yml
11-stats-mem.yml
12-stats-io.yml
13-stats-task.yml
14-stats-container.yml
15-stats-net.yml
16-stats-node.yml
17-stats-service.yml
18-logs-stack.yml
19-logs-numbered.yml
20-logs-metadata.yml
21-logs-all.yml
```

?

ndegory:
@bquenin if you can wait a bit before renaming the yaml files, I have a PR (#375) in progress impacting those files, thanks.

subfuzion:
I'd like them renamed, but there is no reason for them to be numbered sequentially, giving a false sense of them running in order (see discussion here: https://github.com/appcelerator/amp/issues/374#issuecomment-255879628). Just drop the numbers altogether.

[...]

PR #387, opened 2016-10-26
Better regular expressions for CLI tests

ndegory:
Refs #370 

Rewrite of regexp to be more accurate

How to check:

```
swarm start
go test ./cmd/amp/cli
```

has to be rebased after #386 has been merged.

### DIFF
--- a/cmd/amp/cli/test_samples/service-list.yml
+++ b/cmd/amp/cli/test_samples/service-list.yml
@@ -2,4 +2,4 @@ service-list:
   cmd: docker service ls
   args:
   options:
-  expectation: ((ID)(\s)+(NAME)(\s)+(REPLICAS)(\s)+(IMAGE)(\s)+(COMMAND)((.)|(\s))+(pinger)(.)+(appcelerator/pinger)(\s)+(.)*)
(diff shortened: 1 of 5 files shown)


subfuzion (approved):
I think overall these look like good improvements. However, I'm not sure that changing the code to use posix character classes -- requiring an extra pair of enclosing brackets -- is worth the effort just for space, and without consensus and guidance, it's an arbitrary change on your part. I'm not opposed, although to me `\s` is just as readable and a lot less verbose than `[[:space:]]`, and the nuances between perl and posix and non-ASCII support don't really matter here. Of course, once you move beyond `\s`, '`\d`, `\w` classes, `[[:alnum:]]`, `[[:alpha:]]`, `[[:blank:]]`, `[[:lower:]]`, `[[:punct:]]`, `[[:upper:]]`, and `[[:xdigit:]]` become more readable. On that note, I wonder if we're accepting `[[:space:]]` in cases where `[[:blank:]]` is more appropriate? Anyway, we'll move all these into a standard regex dictionary and we can fine tune as needed. Thanks.

subfuzion:
@ndegory Please resolve conflicts so this can be merged. @bquenin's PR renamed the associated files.

ndegory:
@subfuzion I'll revert the [[:space:]] to \s if you prefer, anyway it's anecdotal, the main change is that the regexp now are matching different outputs.

subfuzion:
👍

PR #391, opened 2016-10-27
Add stacks-userguide.md

freignat91:
Add "stacks-userguide.md" to docs

Summary:
- 1 Stack commands
- 2 Stack file description
- 3 ETCD storage
- 4 Networks
- 5 HAProxy

Test: read-it :)

### DIFF
--- a/docs/stacks-userguide.md
+++ b/docs/stacks-userguide.md
@@ -1,0 +1,807 @@
+# AMP - Stacks documentation
+
+
+
+**Summary:**
+- 1 Stack commands
+- 2 Stack file description
+- 3 ETCD storage
+- 4 Networks
+- 5 HAProxy
+
+
+
+# 1 Stack Commands
+
+
+
+## 1.1 Create a new stack
+
+
+Command: **amp stack create `stack-name` -f `stack-file`**
+
+
+- Creates a new stack having the name  `stack-name` using the `stack-file`  definition (see chapter stack file description). 
+- Stores stack information in ETCD (see chapter ETCD storage). 
+
+
+## 1.2 Start a stack
+
+
+Command: **amp stack start `stack-name`**
+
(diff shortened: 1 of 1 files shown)


subfuzion (approved):
Excellent!

docs/stacks-userguide.md
```diff
@@ -0,0 +1,807 @@
...
+
+Sample:
+```
+myservice1:
+    image: appcelerator/pinger:latest
+    public:
+      -name: myexternalname
+       publish_port: 3001
+       internal_port: 3000
+```
```
jhaynie, line 330: curious why we're using `publish_port` and `internal_port`? why not use a more docker standard syntax `internal_port:external_port` such as:

``` yaml
myservice1:
   image: appcelerator/pinger:latest
   public:
       - name: myexternalname
         ports: 3000:3001
```

jhaynie: Protocol could be handled the same way as `3000:3001/udp`

subfuzion: @jhaynie We did a lot of debating over it initially because in this case, `publish_port` is not for exposing the container port on a node host port. Instead, it takes the service port (used to access the service internally within the stack network) and publishes it via the reverse proxy, mapping it as desired, for external use. The debate was over using a format that was familiar vs. confusing people with the different intention since `HOST:CONTAINER` mapping already has a well-established in Docker. However, that argument did go out the window when we decided to use that format for `amp service create` anyway, so we definitely should revisit this. The format we use for `amp service create`, is more powerful; it even supports specifying a name, so for example: `admin:3000` will map the service port to `admin` via haproxy.

jhaynie: ah ok that makes a lot of sense.

PR #393, opened 2016-10-28
Cli tests fixes

ndegory:
cherry picks from #375 
- cosmetic fix on log message
- don't run test if test file is not a yaml file (for instance you editor of choice .swp files)


JosephGJ (approved):
LGTM

PR #394, opened 2016-10-28
Use AMP images for TICK stack

### LINKED ISSUE #392: TICK images should not depend of external resources for configuration
Currently TICK images are downloading an archive from github for the configuration (dashboards, CQ, alerts, etc). If the Docker configuration fails to resolve github (for whatever reason), the container are not configured correctly.
We should use images that already contain the configuration to make it more robust.

ndegory:
- Using AMP images instead of the generic ones for influxdb, kapacitor and grafana

How to test:
- swarm start
- check that the services are up and stable: docker service ps kapacitor, etc
- connect to grafana (http://localhost:6001), check that the dashboard are defined, and check that the dashboards contain data.

### DIFF
--- a/swarm
+++ b/swarm
@@ -39,7 +39,7 @@ IMAGES=(
 MINIMAGES=(
   appcelerator/amp:${AMP_VERSION}
   appcelerator/elasticsearch-amp:${ELASTICSEARCH_VERSION}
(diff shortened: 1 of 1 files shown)


subfuzion (approved):
Nice! 👍

Issue #385, opened 2016-10-26
Examples


subfuzion:
### tl;dr

Create examples and place in subdirectories under `examples`. Examples shouldn't be overly complicated. They should be easy to demo for highlighting various things you can do with AMP related to services, stacks, and stats and log queries. 
### More detail

I asked Henry to create a stack example and open an issue for more examples that make it easier to show people how to work with AMP. Henry added a nice stack example using Docker's own canonical voting app, which you can see [here](https://github.com/appcelerator/amp/tree/master/examples/voting-app). I've closed the original issue since it was specific to stack files.

The Docker voting app example is very effective for a demo, but more simple examples are needed. They need to highlight a feature and demonstrate it in action, which can be seen by looking at the swarm, or logs, or curl'ing, etc. While the stack examples are probably our main priority for now, we also need examples for other CLI commands, such as `amp service`, etc.

You don't need to open an issue to create an example, just submit a PR for the example itself. However, it also might not be a bad idea either to mention your example proposal in a comment here for feedback.

I consider examples an important part of our documentation effort -- they showcase how something works. However, they are also not a substitute for documentation in terms of explaining all the options that are available to be used and the scenarios in which they matter. Each example should have a good README and provide a link to the relevant documentation. If the documentation doesn't exist, then open an issue for it.

Let's keep the examples folder at the top level for now and then place specific examples with their own README.md files in subdirectories. We can keep the subdirectories flat for now. A top level examples/README.md should provide a general overview.

generalhenry:
Idea:
  fully scaled wordpress stack
- varnish
- nginx/wordpress
- redis
- elasticsearch 
- mariadb slaves
- mariadb master

generalhenry:
TODO: add baseline assumptions to examples/README.md

ndegory:
@generalhenry I suggest that we postpone this one to milestone 0.4.0

subfuzion:
I'm happy with the examples we added, but this is an ongoing effort, so I'm going to update the milestone.

PR #395, opened 2016-10-31
Add pinger and counter examples

generalhenry:
progress towards https://github.com/appcelerator/amp/issues/385

Adds pinger and counter examples

Verify with instructions contained in their READMEs

### DIFF
--- a/examples/counter/README.md
+++ b/examples/counter/README.md
@@ -1,0 +1,16 @@
+Counter
+======
+
+Two replicated webapps connected to a redis counter
+
+Run in this directory:
+
+    $ amp stack up -f stack.yml counter
+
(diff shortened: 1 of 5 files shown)


subfuzion (commented):
These are good samples. The readme needs to be updated to mention that if they are connecting to the amplifier service in the local swarm, then the command to start the stack is:

```
amp --server localhost:8080 
```

(per conversation, we could also put this in top level readme)

Also would be good to mention that user can run `amp logs` and `amp stats` for these basic examples.

freignat91 (approved):
LGTM

subfuzion:
I pushed a few minor updates. Overall, very nice, thanks @generalhenry. 👍

PR #397, opened 2016-10-31
Updated regex for metadata logs

neha-viswanathan:
Minor update to regex in metadata logs since it failed. The cli test is however expected to fail since the blocking functionality for service-curl has not been implemented.

To test:
`go test github.com/appcelerator/amp/cmd/amp/cli`


JosephGJ (approved):
LGTM

PR #400, opened 2016-11-01
Add micro.js example with build

generalhenry:
adds to https://github.com/appcelerator/amp/issues/385

Adds an example of building and adding an image to the local registry
Adds a single function style application

To test:
go to `examples/micro`
run `./deploy`
check with `curl http://www.micro.localhost.tv/test` or browser


freignat91 (approved):
Tested ok.
Perhaps needs some more explanations in readme as:
have the amplifier launch locally to make the deploy.sh script run well (or add --server localhost:8080 in the script). My first try failed because of that. 
and explain that the micro service just display the request method/url it receives even if it obvious looking at the code, i had to look at the code instead just read the readme.

subfuzion:
I made a few edits for clarity, but otherwise, nice example. 👍

PR #424, opened 2016-11-07
Fix formatting error

bquenin:
Fixing formatting issues

## Verification
    $ make

There should not be any errors.


freignat91 (approved):
LGTM

PR #438, opened 2016-11-08
amp-cli: Giving more details in case of an error

### LINKED ISSUE #431: amp logs without arguments shouldn't be an error
This is the current output from running `amp logs` without any arguments:

```
$ amp logs
Error during:  Fetch log entries matching provided criteria. If provided, SERVICE can be a partial or full service id or service name.
```

The command should return default logs (eg, *everything*) or it should default to displaying a help message. Either is better than showing an error.

bquenin:
Fixes #431 

## Verification
    ./swarm pull
    ./swarm start
    make install
    amp --server fakehost:12345 logs

It should fail and give you details about the error

### DIFF
--- a/api/client/amp.go
+++ b/api/client/amp.go
@@ -69,16 +69,16 @@ type AMP struct {
 }
 
 // Connect to amplifier
-func (a *AMP) Connect() (conn *grpc.ClientConn, err error) {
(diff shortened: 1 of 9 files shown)


ndegory (approved):
LGTM

generalhenry (approved):

Issue #436, opened 2016-11-08
Create release binaries


subfuzion:
Need a `make dist` or `make release` rule to build release binary goals with the Makefile. These should be stored in an S3 bucket (see #435) and linked from the [Releases](https://github.com/appcelerator/amp/releases) page for Linux, OS X, and Windows.

PR #439, opened 2016-11-08
build binaries for Linux, MacOS and Windows (#436)

ndegory:
Refs #436 

- version is now based on the content of the VERSION file
- targets for amp and amplifier binaries on Linux/Darwin/Windows
- targets for distribution archives

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,7 @@
 /amplifier
 /amp-agent
 /amp-log-worker
+/dist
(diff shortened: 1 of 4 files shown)


ndegory:
https://github.com/appcelerator/docker-gotools/pull/1 would be nice to have, the builds are based on it.

subfuzion (approved):
LGTM. 👍

PR #441, opened 2016-11-08
Templating for generating random Ports

JosephGJ:
creates random port numbers for templating. Similar to random name generation. A range of ports can be supplied to the template.

Added curl method (uncommented) in `service.yml` as retry, timeout and delay fields implemented. Regex moved into `lookup.yml`

to test
`go test github.com/appcelerator/amp/cmd/amp/cli`

### DIFF
--- a/cmd/amp/cli/cli_test.go
+++ b/cmd/amp/cli/cli_test.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"strconv"
 	"testing"
 	"text/template"
 	"time"
@@ -119,13 +120,15 @@ func runTestSpec(t *testing.T, test *TestSpec) (err error) {
 		startTime := time.Now().UnixNano() / 1000000
 
 		for i = -1; i < cmdSpec.Retry; i++ {
+			err = nil
+
 			cmdString := generateCmdString(&cmdSpec)
 			tmplOutput, tmplErr := performTemplating(strings.Join(cmdString, " "), cache)
 			if tmplErr != nil {
 				err = fmt.Errorf("Executing templating failed: %s", tmplErr)
 				t.Log(err)
 			}
-			tmplString := strings.Fields(tmplOutput)
+			tmplString = strings.Fields(tmplOutput)
 
 			t.Logf("Running: %s", strings.Join(tmplString, " "))
 			cmdOutput, cmdErr := exec.Command(tmplString[0], tmplString[1:]...).CombinedOutput()
@@ -148,7 +151,7 @@ func runTestSpec(t *testing.T, test *TestSpec) (err error) {
 			t.Log("This command :", tmplString, "has re-run", i, "times.")
 		}
 	}
-	return
+	return err
 }
 
 func generateCmdString(cmdSpec *CommandSpec) (cmdString []string) {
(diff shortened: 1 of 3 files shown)


cmd/amp/cli/cli_test.go
```diff
@@ -148,7 +151,7 @@ func runTestSpec(t *testing.T, test *TestSpec) (err error) {
 			t.Log("This command :", tmplString, "has re-run", i, "times.")
 		}
 	}
-	return
+	return err
```
subfuzion, line 154: Adding `err` to return statement is not necessary (see [here](https://golang.org/doc/effective_go.html#named-results)).

cmd/amp/cli/cli_test.go
```diff
@@ -119,13 +120,15 @@ func runTestSpec(t *testing.T, test *TestSpec) (err error) {
 		startTime := time.Now().UnixNano() / 1000000
 
 		for i = -1; i < cmdSpec.Retry; i++ {
+			err = nil
+
 			cmdString := generateCmdString(&cmdSpec)
 			tmplOutput, tmplErr := performTemplating(strings.Join(cmdString, " "), cache)
 			if tmplErr != nil {
 				err = fmt.Errorf("Executing templating failed: %s", tmplErr)
```
subfuzion, line 128: If there is an error with templating, we don't want to keep retrying (which makes setting err to nil unnecessary).

cmd/amp/cli/cli_test.go
```diff
@@ -119,13 +120,15 @@ func runTestSpec(t *testing.T, test *TestSpec) (err error) {
 		startTime := time.Now().UnixNano() / 1000000
 
 		for i = -1; i < cmdSpec.Retry; i++ {
+			err = nil
```
generalhenry, line 123: This would be a good spot for a comment

generalhenry (approved):

cmd/amp/cli/cli_test.go
JosephGJ: Neha is putting up comments in her next PR.

cmd/amp/cli/cli_test.go
```diff
@@ -119,13 +120,15 @@ func runTestSpec(t *testing.T, test *TestSpec) (err error) {
 		startTime := time.Now().UnixNano() / 1000000
```
subfuzion, line 120: I think you can simplify your time code. Just use `time.now()` and later check elapsed (eg, `elapsed := time.Since(startTime)`).  I believe the units you're using for the timeout are seconds (needs a comment!), so multiply the timeout by time.Second so you can compare against elapsed.

subfuzion:
@JosephGJ @generalhenry Henry closed this before I finished my review. Please address my comments now as a separate PR.

cmd/amp/cli/cli_test.go
JosephGJ: I can make the templating error return the same as the timeout err statement, however, the reason the `err = nil` is there is for the regex err. If there is a miss matched output and the command is allowed to retry, it should retry that number of times.

cmd/amp/cli/cli_test.go
subfuzion: See [Go Playground](https://play.golang.org/p/InLedxmkUH) for a simple example.

neha-viswanathan: timeout and delay are defined in milliseconds.

PR #448, opened 2016-11-10
Comments to cli_test.go 

neha-viswanathan:
Added comments and performed minor cleanup to cli_test.go


generalhenry (changes requested):
The middle of runTestSpec needs a bit more explanation such as why https://github.com/appcelerator/amp/blob/985e9d49f3236e41e2e0836d02fd9bd5a2a895f5/cmd/amp/cli/cli_test.go#L130 is there

generalhenry (approved):

Issue #384, opened 2016-10-26
Testing


subfuzion:
This issue is parent for the following issues and others we create to increase test categories and coverage for the next release.

https://github.com/appcelerator/amp/issues/370
https://github.com/appcelerator/amp/issues/371
https://github.com/appcelerator/amp/issues/374

Test categories:
- smoke tests - verify the environment is valid, expected services are running, versions are correct, etc.
- unit tests
- integration tests
- API tests - integration tests help, but we also needed dedicated tests (for example, for fuzz testing)
- performance tests

subfuzion:
We made good progress on testing. This is an ongoing effort, so I'm going to update the milestone.

Issue #374, opened 2016-10-24
Non reproducibility of CLI tests


ndegory:
Ref #370 

CLI tests are passing ~ half of the times.
One reason may be because the test specs files contains ordered commands that are not executed in order.

Specs:

```
    grep "^[^ ]" ./cmd/amp/cli/test_samples/09-remove-stack.yml  
    stop-stack:
    remove-stack:
    list-stack:
```

Results:

```
    cli_test.go:41: test 09-remove-stack.yml
    cli_test.go:82: [amp stack ls] Command passed.
    cli_test.go:84: [amp stack ls] Running...
    cli_test.go:85: [amp stack ls] Iteration: 1
    cli_test.go:109: [amp stack ls] Error: miss matched expectation
    cli_test.go:82: [amp stack stop stack1] Command passed.
    cli_test.go:84: [amp stack stop stack1] Running...
    cli_test.go:85: [amp stack stop stack1] Iteration: 1
    cli_test.go:113: [amp stack stop stack1] Command result:
             c36915de96d73f28347551221745e525e26643d225a9c6147348c0a300565d4c

    cli_test.go:82: [amp stack rm stack1] Command passed.
    cli_test.go:84: [amp stack rm stack1] Running...
    cli_test.go:85: [amp stack rm stack1] Iteration: 1
    cli_test.go:113: [amp stack rm stack1] Command result:
             c36915de96d73f28347551221745e525e26643d225a9c6147348c0a300565d4c
```

This is because [iterating on a map with range does not guarantee the order](https://blog.golang.org/go-maps-in-action#TOC_7.).

generalhenry:
Why not a to level array? eg:

``` yaml
- remove-service:
  cmd: amp service rm
  args:
    - pinger
  options:
  expectation: (pinger)
- list-service:
  cmd: docker service ls
  args:
  options:
  expectation: ((ID)(\s)+(NAME)(\s)+(REPLICAS)(\s)+(IMAGE)(\s)+(COMMAND)((.)|(\s))+)
```

subfuzion:
Many integration test runners support running tests both sequentially (ordered) and in parallel. Within our framework, it would seem natural to expect that any yaml file defines a test suite boundary for being able to run in parallel with any other test suite, and within the yaml file, without needing to complicate things each test case should just default to running sequentially as this probably aligns the most with what a test author intuited.

[...]

PR #454, opened 2016-11-11
Asynchronous execution of test specs

JosephGJ:
TestSpecs run in parallel. Uses a waitgroup to do so.

Minor change to service port range.

to test -
`go test ./tests/cli`

### DIFF
--- a/tests/cli/cli_test.go
+++ b/tests/cli/cli_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"sync"
 
 	"bytes"
 	"context"
@@ -47,6 +48,7 @@ var (
 	lookupDir    = "./lookup"
 	regexMap     map[string]string
 	suiteTimeout string
+	wg sync.WaitGroup
 )
 
 //start amplifier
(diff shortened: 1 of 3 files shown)


bquenin:
I'm not sure it's a great idea to run tests in parallel, but maybe I missed something.

generalhenry:
@bquenin reasons to parallel test:
1. It prevents our test time from creeping up too fast
2. It's a more realistic scenario since in production things moving in parallel is the norm

If it becomes an issue we can fall back to serial testing, or work on improving isolation (such as having a fresh amp stack per suite)

subfuzion:
@bquenin It was a discussion that occurred at meetings in the Mountain View office and also in comments for several issues (such as [this](https://github.com/appcelerator/amp/issues/374#issuecomment-255879628)). However, that's not your fault for missing anything, it's my fault; these really should have been captured in the issue I opened around [testing](https://github.com/appcelerator/amp/issues/384), so I'll work on ensuring the thoughts and expectations previously discussed get captured in the parent issue.

The bottom line is that there generally is no expectation that tests run in a certain order, but we also wanted to provide support for explicit sequences. An ordered sequence can only happen within a test file when specified using yaml array syntax. Commands specified as array elements run in sequence, and the next command doesn't start until the previous command returns. Because the previous command may have started an asynchronous operation that won't actually complete until some time after the command itself returns, we're adding support for things like retrying API calls until a response indicates a success or failure status or else times out or exceeds the specified number of retries. Except for this, all other tests should be able to run with some configurable maximum degree of parallelism.

subfuzion:
@JosephGJ Please reference the associated issue in your PR descriptions. Thanks.

bquenin:
@generalhenry @subfuzion thanks for your feedback, it's more clear now. I was just concerned of side effects and particular care that we would need to write test but @generalhenry nailed it when he said **2. It's a more realistic scenario since in production things moving in parallel is the norm**

Acked.

JosephGJ:
Update to asynchronous commands to work in combination with contexts.

Each context uses a goroutine of the `checkTimeout` method to continuously call the `.Done()` function in the contexts select statement. The goroutine fails the test if the deadline is exceeded. If cancelled, the test has passed.

to test
`go test ./tests/cli`

If you want to test some failing cases, insert some random values into the commands timeout field in the yaml file as a duration in string form.

generalhenry (approved):

Issue #456, opened 2016-11-14
Unable to access amp UI


craigivy:
How do we access the amp UI now.  In the past is was http://HOST_NAME:8080

**Steps to reproduce the issue:**
1. intall amp
2. start amp
3. open browser to http://HOST_NAME:8080

**Describe the results you received:**
nothing

**Describe the results you expected:**
amp UI to be displayed

**Output of `amp version`:**
Installed 0.3.0
Command returned:
```
unknown command "version" for "amp"
```

generalhenry:
Instead of `amp version` use `amp info`
For the ui [ui.local.appcelerator.io](http://ui.local.appcelerator.io/) should work

craigivy:
What if its on a remote headless server?
You may want to update the bug template with `amp info`

generalhenry:
It's based on lowest level subdomain.

So if you have a domain set to your remote server
`http://ui.<your domain>.<tld>` should work

If you don't have a domain attached you can set your hosts file
`<server ip>` ui.amp.host

and use http://ui.amp.host/

craigivy:
Verified above works, assume you want to leave this open so it gets documented.  Adding @vivekschauhan so he also understands how it works.

Just curious why you don't want to expose a port for it?

PR #457, opened 2016-11-14
updates amp version information instruction

generalhenry:
see https://github.com/appcelerator/amp/issues/456

`amp version` is a built in from cobra that as of this writing does not return the version
`amp info` is our own command that works, and also provides more information

This PR fixes the issue template to ask for `amp info`


generalhenry:
@neha-viswanathan is working on fixing `amp version` https://github.com/appcelerator/amp/issues/458

Issue #470, opened 2016-11-16
Update docs for amp platform


generalhenry:
ref: https://github.com/appcelerator/amp/issues/437
ref: https://github.com/appcelerator/amp/issues/459
ref: https://github.com/appcelerator/amp/issues/472

The docs need to be updated to use `amp platform . . .`

bquenin:
Closing since all referenced issues are closed

PR #473, opened 2016-11-17
update doc with new amp commands

freignat91:
related to #470

Update documentation with new amp commands usages and options

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,22 +2,26 @@
 
 ## Starting AMP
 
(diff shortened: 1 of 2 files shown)


generalhenry:
There's also a reference in https://github.com/appcelerator/amp/tree/master/examples

freignat91:
Example doc updated

bquenin (commented):
We should probably also mention that the swarm needs to be created beforehand (with docker swarm init for instance, or `amp-swarm-deploy`)

bquenin (approved):
LGTM

Issue #410, opened 2016-11-03
Have a amp test service to test multi-nodes environment on powerful way


freignat91:
related to #301
for tests, we have on one hand:
- (1): Tests on server side as the ones executed launching `make test`. They are powerful because we can analyse returned struct data and any service status if needed, but these tests doesn't work on amp cluster having several nodes and need to start for each a separate amplifier.

on the other hand we have:

- (2): Tests on client side as the ones executed launching `go test -v ./cmd/amp/cli`. They are interesting because they work on any amp cluster, but because these tests work only using console output results, they doesn't allow us to perform powerful verification without a big parsing effort when it's possible (when needed data are in the results)

We need something in addition to (2), replacing (1) and able to work on amp cluster having several nodes.

We could develop a dedicated amp infrastructure test service, build as any other service, so able to connect naturally to any other services including the running amplifier service (or optionally a local amplifier instead) and having as tests all the server side tests we already have (or will have), then we have a way to test any amp clusters without having to launch several time amplifier, in the exact final environment if needed.

For instance, this test service could just start with a volume on amp project and run `make test`, if we refactor the tests in order to do not start a dedicated amplifier themselves.
This service will be a start/run/stop service.  We could have the equivalent of `make test`, but running on any environments.

What do you think?

ndegory:
related to #301, I have a local branch where I'm trying to solve this issue, we can discuss about it @freignat91

freignat91:
ok!

bquenin:
I agree with François, after discussion, we ended up with the following table:

| | `amplifier` running locally  | `amplifier` running as a service in the swarm | 
|----|----|----|
| `swarm` running locally | This where we are today | This is our target for local swarm tests| 
| `swarm` running remotely | Not covered today and not required | This is our target for remote swarm tests | 

Also note that our tests would also be running a service in the swarm, which is not the case today.

PR #479, opened 2016-11-18
Dockerizing integration tests

bquenin:
Related to #410 

Dockerizing integration tests.

## Verification
    $ make install
    $ amp pf stop
    $ amp pf pull
    $ amp pf start
    $ make test


Makefile
```diff
@@ -196,17 +197,34 @@ build-image:
...
 		go test $$pkg ; \
 	done
 
 test-integration:
+	@docker service rm amp-integration-test || true
+	@docker build -f Dockerfile.test -t appcelerator/amp-integration-test .
+	@docker service create --network amp-infra --name amp-integration-test --restart-condition none appcelerator/amp-integration-test
+	@n=""; \
+	while [[ $${n} == "" ]] ; do \
+		n=`docker ps -qf 'name=amp-integration'`; \
```
generalhenry: This trick only works in a single node setup.  Which is likely fine for now.  docker 1.13 will be adding support for docker run within swarm overlay networks which would solve the issue.

generalhenry (changes requested):
Looks great, no troubles running it locally.  I needs a write up of the changes for the change log, and the docs need to be updated to reflect it (there are references to localhost:<port> in the docs)

Makefile
```diff
@@ -196,17 +197,34 @@ build-image:
...
+		go test $$pkg ; \
+	done
+
 test-unit:
 	@for pkg in $(UNIT_TEST_PACKAGES) ; do \
 		go test $$pkg ; \
 	done
 
 test-integration:
+	@docker service rm amp-integration-test || true
```
JosephGJ: This is slightly misleading. Even when the test runs correctly, it returns an error. 

`Error response from daemon: service amp-integration-test not found`

JosephGJ (changes requested):
You need to get the latest changes and rebase to the master. Other than the small comments, this looks really good.

Makefile
bquenin: Yes you're right. I thought about this as an issue.
I didn't know that docker 1.13 will add this feature, but it will definitely solve the issue.
Maybe we should open an issue to keep track of it.

Makefile
bquenin: Yeah, I'll change the output message

Makefile
ndegory: why don't you use `docker service ps amp-integration` ? That would not be limited to the local node. You won't get the logs but at least you can loop on it to wait for the exited status.
Another trick would be to force the node on which you execute the task, with a constraint.

bquenin: @ndegory `docker service ps amp-integration-test` returns the task id, and I need the container id to get the logs. Here is how it works currently:
- starts the `amp-integration-test` service
- get the container id
- `docker logs -f $CONTAINER_ID` to get the output of the tests, which is cool because it's blocking and exits when the container exits.
- get the exit code of the container and exit with this code in the Makefile to report the test exit status

Hopefully, it's clearer with the explanation.

bquenin:
@JosephGJ I rebased on the master

Makefile
ndegory: @bquenin if you need the log then the only solution is to use the constraint:

```
nodeId=$(docker node ls  | grep '*' | awk '{print $1}')
docker service create --constraint "node.id == $nodeId" --network amp-infra --name amp-integration-test --restart-condition none appcelerator/amp-integration-test
```

bquenin: You're right, I'll add a node constraint.

freignat91 (approved):
LGTM

JosephGJ (approved):

generalhenry (approved):

Issue #477, opened 2016-11-18
AMP CLI colors are not terminal theme aware


ndegory:
The AMP CLI colors are designed for terminal with a dark theme (background is full black).
It's not properly displayed in other teams, in particular the [dark-solarized](http://ethanschoonover.com/solarized) theme.
color.FgHiBlack is lost in the background.

freignat91:
Added cmd-theme in .amp.yaml, PR#495

PR #482, opened 2016-11-18
firstMessage color change

neha-viswanathan:
Ref Issue #477 

The firstMessage was printed in `white` color, which was not clearly visible on light-solarized theme. It is now changed to `magenta`.
Tested the colored output on both light and dark solarized themes. It seems like the color `black` (info logs - `Pulling ...` ) appears consistent across both the themes. The above mentioned change is the only what I felt could be modified.
screen shot 2016-11-18 at 3 11 47 pm
screen shot 2016-11-18 at 3 13 01 pm

### DIFF
--- a/cmd/amp/platform_manager.go
+++ b/cmd/amp/platform_manager.go
@@ -63,7 +63,7 @@ type ampService struct {
 }
 
 var (
-	colWhite   = 0
+	colMagenta = 0
 	colInfo    = 1
 	colWarn    = 2
 	colError   = 3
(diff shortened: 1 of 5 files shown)


generalhenry (approved):

ndegory:
@neha-viswanathan I don't have the same behaviour.
My terminal (iterm2) is configured with the dark-solarized theme, and the terminal emulation setting is set to xterm (I tried also with xterm-256, same result).

See the result below, I only see the text when I select it:
image

can you please describe your settings?

neha-viswanathan:
@ndegory I used the osx-terminal.app-colors-solarized theme for terminal. However, I did get the same issue you face while using the dark solarized theme with iTerm. We have to figure out a way to use a different color scheme.

PR #493, opened 2016-11-23
Add amplifier-gateway to platform

generalhenry:
Adds amplifier-gateway to `amp platform`

To test:
`make install`
`amp platform pull`
`amp platform start`
`curl amplifier-api.local.atomiq.io/v1/log`


bquenin (approved):
LGTM

PR #495, opened 2016-11-23
add cmd color themes

freignat91:
Related to #477 

Have amp commands color themes to adapt colors to different terminals L&F.
Default is the "magenta" theme created by Neha, but it doesn't look good on Ubuntu  default terminal which are magenta dark by default.
Added one named "dark" to be back to the original colors (white in place of magenta)

to change it, update $HOME/.amp.yaml file adding this line for instance:
cmdTheme: dark

For now only 2 themes (default or dark) 

Test:
- make install
- start/stop amp: see the regular lines are magenta
- update .amp.yml as explained
- start/stop amp: see the regular lines are white

It's possible to add new theme updating cmd/amp/platform_manager.go

### DIFF
--- a/cmd/amp/platform_manager.go
+++ b/cmd/amp/platform_manager.go
@@ -64,8 +64,9 @@ type ampService struct {
 	user            bool
 }
 
+var currentColorTheme = "default"
 var (
-	colMagenta = 0
+	colRegular = 0
 	colInfo    = 1
 	colWarn    = 2
 	colError   = 3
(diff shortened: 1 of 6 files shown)


cmd/amp/platform_manager.go
```diff
@@ -615,6 +616,51 @@ func (s *ampManager) close() {
...
+				return strings.Trim(param[1], " ")
+			}
+		}
+	}
+	return ""
+}
+
+func (s *ampManager) setColors() {
+	theme := s.getTheme()
+	if theme == "white" {
```
ndegory: I think the name of the theme should not be based on one of the color of the theme, but should match the terminal theme on which it displays best.
So here white should be black.

ndegory (commented):
good idea to rename the column without the color, as it could change with the theme. The naming of the theme should be different though.

cmd/amp/platform_manager.go
freignat91: Any idea for this name?

freignat91: done, renamed "dark", PR description updated too

cmd/amp/platform_manager.go
```diff
@@ -615,6 +616,51 @@ func (s *ampManager) close() {
 	//s.docker.Close()
 }
 
+func (s *ampManager) getTheme() string {
+	homedir, err := homedir.Dir()
+	if err != nil {
+		return ""
+	}
+	file, err := os.Open(path.Join(homedir, ".amp.yaml"))
```
generalhenry: why manually check the config?

freignat91: yes, you're right, i updated the code to use the regular amp config

bquenin (approved):
LGTM

generalhenry:
You can also do 
`amp config cmdTheme dark` to switch to dark
and `amp config cmdTheme default` to switch back

generalhenry (approved):

PR #504, opened 2016-11-25
Disable refresh on index

bquenin:
Disable forced refresh on index for elasticsearch to reduce CPU load

## Verification
    $ amp pf stop
    $ make install
    $ ./shrink.sh local
    $ amp pf start --local
    $ make test


generalhenry (approved):

freignat91 (approved):
LGTM

PR #508, opened 2016-11-29
Removing duplication of nats cluster id definition

bquenin:
## Verification
    $ amp pf stop
    $ make install
    $ ./shrink local
    $ amp pf start --local
    $ make test


freignat91 (approved):
LGTM

PR #509, opened 2016-11-29
Removed duplication of configuration

bquenin:
Configuration of the amp swarm services was all over the place. I put it in a single file and update references.

## Verification
    $ amp pf stop
    $ make install
    $ ./shrink.sh local
    $ amp pf start --local
    $ make test


freignat91 (approved):
LGTM

PR #526, opened 2016-12-06
Fix cross compilation for binaries distribution

### LINKED ISSUE #523: Make dist doesn't work anymore on master branch
target build-cli-linux of Makefile is not working anymore:
exec: "gcc": executable file not found in $PATH

this message is related to the container building the binaries.
same for make build

ndegory:
fixes #523 

- use image appcelerator/gotools2:1.0.0

how to test:
- make dist


freignat91 (approved):
LGTM, tested ok

Issue #425, opened 2016-11-07
REST Gateway


subfuzion:
One of the reasons that helped influence the decision to switch to gRPC had to do with the [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) project. Being able to exploit the benefits of protocol buffers and gRPC was compelling, but having the ability to be able to generate a reverse proxy that could map REST API operations to gRPC, plus provide [Swagger](http://swagger.io/) was a benefit that was also a part of the consideration.

Providing a REST API is important for consumers for which we don't provide an auto-generated stub.

This issue is to first determine if it is feasible to generate a JavaScript stub that is fully compatible with our API (including streaming APIs) that can be used to build a web UI; then, if necessary, to leverage `grpc-gateway` to generate and expose a REST API service as part of amp that can be used as an alternative.

generalhenry:
Pure browser JS answer: http://stackoverflow.com/a/35166533/491845
The good news is that the browsers are beginning to support http2 trailers, but it'll be years until there's all the support needed for seamless grpc in the browser.

The midway point of partial grpc in the browser is in progress, but it's at the proposal phase. https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md

In the mean time the grpc-gateway gets us immediately to the midway point (no BiDirectional support, though that is being worked on with websockets).

For grpc-gateway support we need to:
- [x] add grpc gateway runtime
- [x] update https://github.com/appcelerator/docker-protoc with gateway and swagger plugins
- [x] update https://github.com/appcelerator/amp/blob/master/hack/proto.go with gateway and swagger output
- [x] add route annotations to our .protos
- [ ] combine swagger output
- [ ] add swagger into our docs/examples
- [x] create an amplifier-gateway
- [x] add amplifier-gateway to the make file
- [x] add amplifier-gateway to the swarm script
- [x] set HAProxy to route traffic to the gateway
- [x] add example usecase

PR #527, opened 2016-12-06
Add grpc-gw usage example

generalhenry:
ref: https://github.com/appcelerator/amp/issues/425

Add a example of using the amplifier-gateway api to build a ui

To test:
`./shrink.sh local`
`amp pf start --local`
`cd examples/ui`
`./deploy.sh`
open http://www.gw-ui.local.atomiq.io/stacks/


chrisccoy (approved):
:+1: works like a charm.

chrisccoy:
FYI- You may get  a protobuf error:
```
Error during: amp up, reason: rpc error: code = 2 desc = proto: can't skip unknown wire type 7 for stack.Stack
```
 if you do not perform a `make install` prior to the deploy step

JosephGJ (approved):
LGTM. Had to open link in a chrome window rather than safari.

generalhenry:
re: safari
I built it using new browser features, and safari is missing one of them (fetch) it's easy to polyfill but for now the browser support matrix is [chrome 55+] though it should also be working in firefox, opera and edge.  (safari is sadly the new IE)

bquenin (approved):
LGTM, tested on linux with Chrome 55

Issue #532, opened 2016-12-08
panic: amp platform start --local -f


subfuzion:
amp was already running; checked out a new branch to review a PR and attempted to restart.

```
amp (storage-api-cli)$ amp pf start --local -f
Starting AMP platform
Force mode: on
Service telegraf-agent removed
Service amplifier removed
Service amplifier-gateway removed
Service elasticsearch removed
Service registry removed
Service amp-log-worker removed
Service kapacitor removed
Service pinger-pinger removed
Service haproxy removed
Service nats removed
Service influxdb removed
Service amp-ui removed
Service amp-agent removed
Service telegraf-haproxy removed
Service etcd removed
Service grafana removed
Service pinger-haproxy removed
Service pinger-haproxy image appcelerator/haproxy:1.0.1@sha256:27dc67585bd8d603dc60ba20c336eb6be89faba40fc22ee90f11896a047fe29a doesn't exist
Force mode: service pinger-haproxy is considered as ready
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x107d5]

goroutine 1 [running]:
panic(0x6d1600, 0xc4200160f0)
        /usr/local/go/src/runtime/panic.go:500 +0x1a1
main.(*ampManager).createService(0xc4202d1a20, 0xc420189b80, 0x0, 0x0)
        /Users/tony/projects/go/src/github.com/appcelerator/amp/cmd/amp/platform_manager.go:467 +0x2f5
main.(*ampManager).start(0xc4202d1a20, 0xc4202914c0, 0x0, 0x0)
        /Users/tony/projects/go/src/github.com/appcelerator/amp/cmd/amp/platform_manager.go:122 +0x404
main.startAMP(0xb7af60, 0xc4203d5440, 0x0, 0x2, 0x1, 0xc420122268)
        /Users/tony/projects/go/src/github.com/appcelerator/amp/cmd/amp/platform_start.go:60 +0x2a2
main.glob..func7(0xb7af60, 0xc4203d5440, 0x0, 0x2)
        /Users/tony/projects/go/src/github.com/appcelerator/amp/cmd/amp/platform_start.go:15 +0x49
github.com/appcelerator/amp/vendor/github.com/spf13/cobra.(*Command).execute(0xb7af60, 0xc4203d5200, 0x2, 0x2, 0xb7af60, 0xc4203d5200)
[...]

PR #533, opened 2016-12-09
fix amp start -f panic

freignat91:
Fix the #532 

it's not easy to reproduce the same error condition, to do so
- make install
- start a service, not an infrastructure one
- remove its image or not (both cases can be tested)
- start amp -f (if amp is not alreday started)
- restart amp force mode: amp pf start -f
- you don't got panic anymore.


ndegory (approved):
LGTM, tests passing on Mac OS

PR #536, opened 2016-12-09
registry command

### LINKED ISSUE #535: AMP registry not reachable remotely
AMP CLI registry command is not working correctly when using a remote CLI:
- --server option is not used by the registry commands
- --domain is only used by the push command (and not the ls command)
- lookup errors while pushing on a custom domain name

ndegory:
fixes #535 

- ```domain``` option applies to all registry commands
- leverage the Docker API for push options
- updated documentation with usage help

how to test:
- make clean install
- create a cluster with multiple nodes (you can use appcelerator/amp-swarm-deploy for that)
- create an alias on route53 to point to an instance of the cluster (domain.com), with * subdomain CNAME
- amp registry --domain domain.com ls # should return an empty array
- amp registry --domain domain.com push an_image # should tag and push to the cluster registry
- amp registry --domain domain.com ls # should return an array with the image just pushed

### DIFF
--- a/cmd/amp/registry.go
+++ b/cmd/amp/registry.go
@@ -1,16 +1,20 @@
 package main
 
 import (
+	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/appcelerator/amp/api/client"
+	distreference "github.com/docker/distribution/reference"
+	"github.com/docker/docker/api/types"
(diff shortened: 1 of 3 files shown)


freignat91 (approved):
LGTM and tested ok
The PR is ok, but i going to open an issue on the --domain argument usage #540

PR #566, opened 2016-12-13
fix panic

### LINKED ISSUE #563: Should not get a panic for supplying invalid options
Our tools should not panic when supplied an invalid command or option. Currently, however, invalid options for `amp` do. For example, the following (mispelled) option causes a panic.

    $ amp service create --helpp

The CLI should simply display "Unrecognized option: helpp" and exit (1).

generalhenry:
closes: https://github.com/appcelerator/amp/issues/563

To test:
`make install`
`amp service create --helpp`

### DIFF
--- a/cmd/amp/main.go
+++ b/cmd/amp/main.go
@@ -96,7 +96,7 @@ func main() {
 		cli.Exit(1)
 	}
 	if err := cmd.Execute(); err != nil {
-		if AMP.Verbose() == false {
+		if verbose {
 			fmt.Printf("Error during: amp %s, reason: %v\n", cmd.Name(), err)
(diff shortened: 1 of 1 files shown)


cmd/amp/main.go
```diff
@@ -96,7 +96,7 @@ func main() {
 		cli.Exit(1)
 	}
 	if err := cmd.Execute(); err != nil {
-		if AMP.Verbose() == false {
+		if verbose {
```
subfuzion, line 99: `== false` ?  How did that happen? :p

subfuzion:
Thanks @generalhenry 👍

PR #582, opened 2016-12-14
added editorconfig, fixed style exceptions

### LINKED ISSUE #571: editorconfig
Go provides standard formatting support, but it would be nice to have an [.editorconfig](http://editorconfig.org/) in the repo for the editors that support it so that formatting appears consistent for not just go, but also

    * yaml files (`.yaml|.yml`)
    * protobuf files (`*.pb`)
    * shell scripts (`*.sh`)

generalhenry:
closes: https://github.com/appcelerator/amp/issues/571

I tried my best to simply pave the cow paths.  I did not add any rules for .md since we're not at all consistent.


.editorconfig
```diff
@@ -0,0 +1,33 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
```
ndegory, line 8: isn't that in conflict with the default OS end of line choice? Git already manages text files depending on the OS, checking in or out text files with LF or CRLF based on the context. I'd prefer the approach of adding file patterns in a .gitattributes file to specify which one should be considered as text, and leave the default OS mechanism.

.editorconfig
```diff
@@ -0,0 +1,33 @@
...
+insert_final_newline = true
+
+# Matches multiple files with brace expansion notation
+# Set default charset
+[*.{js,css,html}]
+charset = utf-8
+
+# 2 space indentation
+[*.{js,json,yml,yaml,proto,sh,html,css}]
```
ndegory, line 17: let's also add our scripts without extension. I can only think about hack/build

ndegory (changes requested):

.editorconfig
generalhenry: I think it might be easier to just rename hack/build to hack/build.sh

.editorconfig
generalhenry: It's in conflict with windows default, which is the point.  I'm not sure I see the advantage of adding .gitattributes when this line of the .editorconfig should do the trick.  The final result is `lf` in the repo.

ndegory: not all editors have a plugin for .editorconfig, whereas the .gitattributes will be anyway be applied by git.
And it's more natural to allow the editor to edit the file in the native line ending.

.editorconfig
ndegory: one or the other is fine.

.editorconfig
subfuzion: I personally don't have a problem with adding this per se. It's fairly standard to add this to `.editorconfig`, as you can see from a sampling of major oss projects [here](https://github.com/editorconfig/editorconfig/wiki/Projects-Using-EditorConfig).

.editorconfig
subfuzion: I disagree. Library sh and bash scripts should have the `.sh` extension, but executable scripts should not. This is a subjective style issue, but I'm old school on this, I guess (even though I haven't always been 100% consistent myself). The only convenience for adding the extension has to do with using editors that require the extension for syntax highlighting. The [Google Shell Style Guide](https://google.github.io/styleguide/shell.xml#File_Extensions) agrees. Bottom line, if the file has a shebang line and the executable bit set, then don't add an extension. If the script is meant to be invoked by another script and is not executable standalone, then add the extension.

Makefile
```diff
@@ -110,38 +110,38 @@ install:
 	@go install $(LDFLAGS) $(REPO)/$(CMDDIR)/$(GATEWAY)
 
 build:
-	@hack/build $(CLI)
-	@hack/build $(SERVER)
-	@hack/build $(AGENT)
-	@hack/build $(LOGWORKER)
-	@hack/build $(GATEWAY)
+	@hack/build.sh $(CLI)
```
subfuzion: I prefer to not add `.sh`. I realize it's a style issue, but I already commented my thoughts on this above. I know everyone isn't 100% consistent out there, including Docker itself. They add the extension under their `hack` directory, and yet have dozens of shell scripts that don't under `hack/make`. Bottom line, unless not having the extension is frustrating everyone with editors that key off the extension for syntax highlighting (I don't have that problem with vim), then let's standardize on leaving it off as I had it originally.

subfuzion (changes requested):
@generalhenry Looking good overall. My only request is to leave off the `.sh` extension.

Makefile
generalhenry: Ok, I'll remove the commit at adds the extension.

generalhenry:
'.sh' reverted

.editorconfig
ndegory: no problem, anyway nobody in her right mind is developing on Windows.

ndegory (approved):

subfuzion (approved):

PR #583, opened 2016-12-14
issue 573-576: grpc codes error for stacks and stats

freignat91:
related to #573 and #576 (also #591)
add grpc errors for stack and stats

test:
- make install
- ../shrink.sh local
- amp pf start -v --local
- make test

rules applies for all grpc error issues to be consistent:
- if docker error: code.Internal
- If ETCD creation/update/delete error: code.Internal
- if state machine creation failed: code.Internal
- if state machine inconstancy: code.FailedPrecondition
- if stack file error or arg error: code.InvalidArgument
- if internal timeout: code.DeadlineExceeded
- if resource not found: code.NotFound
- if resource creation already exist: code.AlreadyExists

### DIFF
--- a/api/rpc/stack/stack.go
+++ b/api/rpc/stack/stack.go
@@ -23,6 +23,8 @@ import (
 	units "github.com/docker/go-units"
 	"github.com/golang/protobuf/proto"
 	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 )
 
 const stackRootKey = "stacks"
@@ -59,6 +61,7 @@ func (s *Server) Up(ctx context.Context, in *StackFileRequest) (*StackReply, err
 	}
 
 	//start the stack
+	//start the stack
 	startRequest := StackRequest{
 		StackIdent: r.StackId,
 	}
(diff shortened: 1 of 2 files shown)


api/rpc/stack/parse.go
```diff
@@ -68,7 +70,7 @@ func ParseStackfile(ctx context.Context, in string) (*Stack, error) {
 	var stack = &Stack{}
 	specs, err := parseStack([]byte(in))
 	if err != nil {
-		return nil, err
+		return nil, grpc.Errorf(codes.InvalidArgument, "%v", err)
```
generalhenry: The parsing is being done client side currently, not that these hurt there.

subfuzion (changes requested):
@freignat91 Looks good. Please resolve conflicts so I can merge. Thanks.

api/rpc/stack/parse.go
freignat91: oh yes, excellent, i didn't see that it has been moved client side. Good! 
I updated back the pr for parse.go

subfuzion (approved):

PR #596, opened 2016-12-16
amp --config-file now display the configuration file in use (if any)

### LINKED ISSUE #557: amp --config-file with no argument should display config file path

bquenin:
Fix #557 

## Verification
    $ make install

:warning: Backup your configuration file if you have one!

    $ rm -f ~/.config/amp/amp.yaml
    $ amp --config-file
    No configuration file used (using default configuration)
    $ mkdir -p ~/.config/amp
    $ echo "{
      "CmdTheme": "",
      "Github": "",
      "Images": null,
      "Port": "",
      "ServerAddress": "127.0.0.1:8080",
      "Target": "",
      "Verbose": false
    }" > ~/.config/amp/amp.yaml
    $ amp --config-file
    /home/bq/.config/amp/amp.yaml

Restore your configuration file if needed :)

### DIFF
--- a/cmd/amp/main.go
+++ b/cmd/amp/main.go
@@ -7,6 +7,7 @@ import (
 	"github.com/appcelerator/amp/api/client"
 	"github.com/appcelerator/amp/cmd/amp/cli"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


cmd/amp/main.go
```diff
@@ -85,10 +96,11 @@ func main() {
 	RootCmd.SetUsageTemplate(usageTemplate)
 	RootCmd.SetHelpTemplate(helpTemplate)
 
-	RootCmd.PersistentFlags().StringVar(&configFile, "config", "", "Config file (default is $HOME/.config/amp/amp.yaml)")
+	RootCmd.PersistentFlags().StringVar(&configFile, "config", "$HOME/.config/amp/amp.yaml", "Config file")
```
ndegory: Problem is, it breaks the use case when the user uses an env variable for XDG_CONFIG_HOME (not sure who would do that), not a big deal but that means that we have some dead code.

ndegory: I've made the test, it behaves as expected, so no problem after all!

bquenin: ok good, can you change your review then ?

ndegory: actually it was just a comment, not a review. But I'll approve.

ndegory (approved):
LGTM

subfuzion:
@bquenin LGTM. I'll merge after you resolve the conflicts. Thanks.

subfuzion (changes requested):
@bquenin LGTM. I'll merge after you resolve the conflicts. Thanks.

subfuzion (commented):
Wondering if we should detect Windows and choose a file path more appropriate for that platform (than "$HOME/.config/amp/amp.yaml") or is this going to resolve okay and also not be too weird (`.config/`)? Just asking because I'm not a Windows user, but can we get confirmation? (Forgot to leave this as a line comment, but it applies to main.go:99).

bquenin:
@subfuzion I'm no windows user either, so I can't really say

ndegory:
@hagantzer?

subfuzion (approved):
@bquenin Let's leave it the way it is for now. We probably want to revisit this and any other potential path-based idiosyncrasies at some point (or else wait until someone complains ;) ).

PR #600, opened 2016-12-16
stats integration test robustness (fixes #599)

### LINKED ISSUE #599: Stats integration tests randomly fail
The stats integration tests uses a regexp to look for a container (amp), but the container names also contain random strings (like kapacitor.1.ljnd**amp**wvlkh), which can lead to false positive, eventually failing the test.

ndegory:
fixes #599 

- more precise regexp
- de-capitalize error messages

how to test:
- make clean install-cli
- amp pf start
- make test

No way to trigger the same cause, but at least we can check that the tests are still passing most of the time.


neha-viswanathan (approved):

bquenin (approved):
LGTM (linux)

Issue #548, opened 2016-12-13
amp version


subfuzion:
`amp version` should work similarly to `docker version`. It should display version information about the client environment, and if it is able to connect to `amplifier`, it should also display version information about the server environment.

PR #608, opened 2016-12-21
Amp version command - more useful info

JosephGJ:
ref: #548 

Added more useful info to `amp version` command. Based off the `docker version` command. Will display AMP info and Amplifier info when platform is running, only AMP info is displayed if platform is stopped.

More can be added in future.

```
AMP:
 Version:       v0.4.0-dev
 Build:         90dbda1c
 ConfigAddr:    127.0.0.1:8080
 Go version:    go1.7.4
 OS/Arch:       darwin/amd64

Amplifier:
 Version:       v0.4.0-dev
 Build:         unknown
 Port:          :50101
 Go version:    go1.7.4
 OS/Arch:       darwin/amd64
```


generalhenry (approved):

api/client/amp.go
```diff
@@ -53,6 +53,8 @@ func (l logger) Println(args ...interface{}) {
 
 // Configuration is for all configurable client settings
 type Configuration struct {
+	Version            string
```
ndegory: Should Version and Build be part of the configuration? They will be displayed with the ```amp config``` command, and will be settable (is it a word?) with ```amp config Version value```.

api/rpc/version/version.proto
```diff
@@ -0,0 +1,29 @@
+syntax = "proto3";
+
+package version;
+
+import "google/api/annotations.proto";
+
+service Version {
+  rpc List (ListRequest) returns (ListReply) {
+    option (google.api.http) = {
+      get: "/v1/version",
```
ndegory, line 10: about the .pb.gw.go, please have a look at the Makefile, the PROTOGWTARGETS are not dynamically set, you should add the new files in this list.

api/client/amp.go
JosephGJ: I've removed these as they're unnecessary.

api/rpc/version/version.proto
JosephGJ: I've added `version.pb.gw.go` to the PROTOGWTARGETS.

JosephGJ:
Amplifier's `build` field has been removed (for now) as it is pulling `unknown` from the `Dockerfile` rather than the build from the Makefile.

subfuzion (approved):
I gave you my feedback in person before the holidays -- nice work, @JosephGJ 👍

PR #613, opened 2016-12-30
[refactor] moved amp confighelper

subfuzion:
Moved `amp/confighelper` to `amp/cli/confighelper`, and moved the `SaveConfiguration` function that was in `amp/cli/cli` to match up with the `InitFunction`.

## Verification

    $ make install-cli
    $ amp config # should display config
    $ amp config Verbose true
    $ amp config Verbose # verify true
    $ amp config Verbose false
    $ amp config Verbose # verify false

### DIFF
--- a/cmd/amp/main.go
+++ b/cmd/amp/main.go
@@ -58,7 +58,7 @@ var (
 // All main does is process commands and flags and invoke the app
 func main() {
 	cobra.OnInitialize(func() {
-		InitConfig(configFile, Config, verbose, serverAddr)
+		cli.InitConfig(configFile, Config, verbose, serverAddr)
 		if addr := RootCmd.Flag("server").Value.String(); addr != "" {
 			Config.ServerAddress = addr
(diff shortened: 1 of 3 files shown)


cmd/amp/main.go
```diff
@@ -67,7 +67,7 @@ func main() {
 		}
 		AMP = client.NewAMP(Config)
 		if AMP.Verbose() == false {
-			RootCmd.SilenceErrors = true
+			;RootCmd.SilenceErrors = true
```
generalhenry: Stray semi?

subfuzion: Yes, missed this here, actually fixed it in a subsequent PR, though. Anyway, will fix here, thanks.

freignat91 (approved):
Great, then configHelper is now importable and could be used also by ampadm.
I'll suppress the specific apmadm confighelpter to use this one in another PR

PR #624, opened 2017-01-03
Add make rule to print all the make rules

subfuzion:
The Makefile has quite a few rules at this point, and they're getting hard to remember. This adds a rule to print all the rules in sorted order, which can be a bit quicker than opening and inspecting the Makefile.

    $ make rules


freignat91 (approved):
LGTM
tested ok

PR #627, opened 2017-01-03
fix default value for adminserveraddress

freignat91:
add default value for adminServerAddress in config

Test:
- make install
- rm ~/.config/amp/amp.yaml
- amp config Verbose true (for instance)
- verify the adminServerAddress value is well: 127.0.0.1:31315


bquenin (approved):
LGTM

PR #630, opened 2017-01-03
Prevent amp logs from crashing when Verbose flag is set

### LINKED ISSUE #629: amp logs crashes when Verbose flag is set

bquenin:
Fix #629 

## Verification
    $ make install
    $ make TAG=local build-image
    $ amp pf pull
    $ amp pf start --local
    $ amp config Verbose true
    $ amp logs

Should not crash


freignat91 (approved):
LGTM
tested ok

Issue #646, opened 2017-01-05
Prerequisite check for ampadm doesn't work


freignat91:
adm-server detects the elasticsearch prerequisite issue correctly, but tries to fix it in the container context, which fails.

ndegory:
to be moved in the bootstrap agent, it's not a container and it's run as root

PR #648, opened 2017-01-05
Don't try to fix sysctl values when in a container

freignat91:
related to issue #646 

fix the prerequisite detection error:

test:
- make TAG=local build-image
- ampadm cluster start --local
- ampadm pf start
return an error showing the prerequisite issue (sysctl -n vm.max_map_count < 262144)
(of course, if the node already has a max map count setting of at least 262144, the error won't occur)

### DIFF
--- a/cmd/adm-server/servercore/amp-service.go
+++ b/cmd/adm-server/servercore/amp-service.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"os/user"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


ndegory:
looks like we should add the change in the bootstrap command, since we're root.

subfuzion (approved):
LGTM

subfuzion:
@freignat91 Thanks for the update. Just so you know, if you say "closes #646" or "fixes #646" in the description, then the issue will get automatically closed when the PR is merged. 👍

Issue #656, opened 2017-01-07
Add web-client Go function example


alexellis:
As a way of contributing something I wanted to add a web-client Go function example. It will go off to the Docker Hub API and find the amount of repositories a user or org. has.

The test syntax would work like this:

```
# echo appcelerator | ./webclient-demo-function
The organisation or user appcelerator has 34 repositories on the Docker hub.

# echo alexellis2 | ./webclient-demo-function
The organisation or user alexellis2 has 97 repositories on the Docker hub.
```

Testing through the Docker image after running `make image`:

```
# echo alexellis2 | docker run -i appcelerator/amp-webclientdemo-function:latest
The organisation or user alexellis2 has 97 repositories on the Docker hub.
```

I have used "demo-function" as a template including the Dockerfile/Makefile.

PR #657, opened 2017-01-07
Add new function example with web-client request to Docker hub #656.

alexellis:
Signed-off-by: Alex Ellis <AlexEllis2@gmail.com>

### What I did

As a way of contributing something I wanted to add a web-client Go function example. It will go off to the Docker Hub API and find the amount of repositories a user or org. has.

### How I did it

Created a new example based upon amp-demo-function.

### How to verify it

See #656 

### Description for the changelog

Add amp demo function that makes web request.

### DIFF
--- a/examples/functions/webclient-demo-function/.gitignore
+++ b/examples/functions/webclient-demo-function/.gitignore
@@ -1,0 +1,1 @@
+webclient-demo-function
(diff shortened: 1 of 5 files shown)


subfuzion (approved):
This is great, @alexellis -- thanks for contributing! 👍

PR #659, opened 2017-01-09
Revert modification on make install

### LINKED ISSUE #658: 'make install' takes too long time (more than 30 secs)

bquenin:
Fix #658 

## Verification
    $ make clean install
    should take ~10 seconds
    $ make install
    should take ~2 seconds

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,36 +141,29 @@ clean: proto-clean bin-clean
 
 install: install-cli install-server install-agent install-log-worker install-gateway install-fn-listener install-fn-worker install-adm-server install-adm-agent install-ampadm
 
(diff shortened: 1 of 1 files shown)


ndegory:
I see the benefit of removing the dependencies to the sources and systematically running the go compilation, as the go compiler notices that there's no need to compile, we don't lose time.
But, we lose the dependency check on the proto files, which means it doesn't trigger the proto compilation if the files are not up to date, which is an issue.
Let's see if we can optimize the dependency check instead...

ndegory:
something's broken in the definition of the pb.gw.go file list, some files are not real candidates and therefore are always missing, which triggers the proto compilation all the time, which takes a lot of time. Addressed in #664.

Issue #660, opened 2017-01-09
AMP Teams


subfuzion:
This is an epic pertaining to defining and implementing the amp permissions model around organizations, teams, and resources.

Subsequent issues will be opened for implementing the storage data access layer, protobuf types and grpc services, identity provider, authentication and authorization, CLI command support, and tests.

The effort will iterate and build on the model shown in the following diagram:
![model](https://cloud.githubusercontent.com/assets/159738/21780566/ef7f5d1e-d660-11e6-8ff4-8f0aa2359cbb.png)

PR #661, opened 2017-01-09
[v0.6.x] Add account API and CLI stubs

generalhenry:
ref: https://github.com/appcelerator/amp/issues/660

Adds definitions for Users, Organizations, Teams and Permissions and operations on them.


api/rpc/account/account.proto
```diff
@@ -0,0 +1,248 @@
...
+      body: "*"
+    };
+  };
+  rpc Verify (VerificationRequest) returns (google.protobuf.Empty) {
+    option (google.api.http) = {
+      post: "/v1/account/{name}/verification/{code}",
+      body: "*"
+    };
+  };
+  rpc CreateOrganization (OrganizationRequest) returns (google.protobuf.Empty) {
```
chrisccoy, line 21: Shouldn't we return the generated Id as part of the response?

cmd/amp/account.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	// Interactive
+	verifyCmd = &cobra.Command{
+		Use:   "verify [CODE]",
+		Short: "verify email using code",
+	}
+
+	assumeRoleCmd = &cobra.Command{
+		Use:   "assume [ORGANIZATION] [TEAM]",
```
subfuzion: Only need to switch organization for now. If you're on multiple teams with access to a resource, your grant will be the highest level of access available to you.

cmd/amp/account.go
```diff
@@ -0,0 +1,116 @@
...
+		Short: "list users, optionally filter by organation and team",
+	}
+
+	listOrganizationsCmd = &cobra.Command{
+		Use:   "list organizations",
+		Short: "list organizations",
+	}
+
+	listTeamsCmd = &cobra.Command{
+		Use:   "list teams [ORGANIZATION]",
```
subfuzion, line 61: I think we really only need to list the versions of the team you're currently switch to.

api/rpc/account/account.proto
generalhenry: I'm trying to avoid IDs for now. So far none of the APIs use IDs other than resource IDs.  The names are unique (or ORG_NAME+TEAM_NAME in the case of teams)

cmd/amp/account.go
```diff
@@ -0,0 +1,116 @@
...
+		Short: "list teams by organization",
+	}
+
+	listPermissionsCmd = &cobra.Command{
+		Use:   "list permissions [ORGANIZATION] [TEAM]",
```
subfuzion, line 66: This should list for all the resources or else just the specified resource the teams the user belongs to that has access to that resource as well as the level of access.

cmd/amp/account.go
```diff
@@ -0,0 +1,116 @@
...
+		Short: "edits account information",
+	}
+
+	deleteCmd = &cobra.Command{
+		Use:   "delete [name]",
+		Short: "deletes account",
+	}
+
+	addOrganizationMembersCmd = &cobra.Command{
+		Use:   "add organation members [ORGANIZATION] [MEMBERS...]",
```
subfuzion: Should be for the current organization. Do we want to call this "invite" since users don't get added until they accept?

subfuzion (changes requested):
Great stake in the ground! 👍

chrisccoy:
What's up with the branch name? underscores are so 2000 ;-)

cmd/amp/account.go
generalhenry: We don't have a notification/accept system defined so far.  We should create some user stories around it.

api/rpc/account/account.proto
subfuzion: Agree. The "ID" is the user name that was specified, which must be unique. We don't need to surface the internal id.

cmd/amp/account.go
```diff
@@ -0,0 +1,116 @@
+package main
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var (
+	// Interactive
+	loginCmd = &cobra.Command{
```
chrisccoy: Need to change because it is already being used:
```
cmd/amp/login.go:25: loginCmd redeclared in this block
	previous declaration at cmd/amp/account.go:12
```

subfuzion: Caught by the CI build: https://travis-ci.org/appcelerator/amp/builds/190429445

subfuzion (approved):
LGTM

subfuzion:
Ignoring failed CI build ("no space left on the device") error.

PR #662, opened 2017-01-10
[v0.6.x] User stories to account command

neha-viswanathan:
Added User stories to account commands.


subfuzion:
@neha-viswanathan Please resolve conflict. Thanks.

subfuzion:
Let's remove the user stories from code and add the long descriptions. We can put the user stories in the doc. Also, I suppose we should stick to the conventional user story format (As a <user>...). Otherwise, thanks @neha-viswanathan

PR #664, opened 2017-01-10
fix proto files compilation

### LINKED ISSUE #658: 'make install' takes too long time (more than 30 secs)

ndegory:
fixes #658 

The dynamic list for *.pb.gw.go targets was broken, the resulting list contained files that were never generated, triggering systematically the proto compilation, loosing a lot of time.

- fixing the dynamic proto targets
- combining the explicit rules
- removing unneeded lines

## How to compare
Timing below are on a Mac, YMMV.
You can notice a big difference in the total cpu time.

### Before modification

```
~/Development/src/github.com/appcelerator/amp(master ✔) make clean
~/Development/src/github.com/appcelerator/amp(master ✔) time make install-cli ; time make install-cli
make install-cli  9.91s user 3.74s system 14% cpu 1:36.78 total
make install-cli  9.45s user 3.98s system 20% cpu 1:07.01 total

```
### After the modification

```
~/Development/src/github.com/appcelerator/amp(master ✗) make clean
~/Development/src/github.com/appcelerator/amp(master ✗) time make install-cli ; time make install-cli
make install-cli  9.12s user 2.73s system 33% cpu 34.871 total
make install-cli  4.27s user 1.19s system 100% cpu 5.417 total

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,3 @@
-  
 .PHONY: all clean proto-clean bin-clean install install-host fmt simplify check version build-image run proto proto-host rules
 .PHONY: build build-cli-linux build-cli-darwin build-cli-windows build-server build-server-linux build-server-darwin build-server-windows
 .PHONY: dist dist-linux dist-darwin dist-windows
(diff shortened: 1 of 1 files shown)


bquenin (approved):
LGTM (linux)

subfuzion:
@ndegory At some point a while ago, I used *.proto files as the prerequisites for other build rules, the rationale being if if any proto file was updated, then we needed to run protoc on it. At some point it seems that it was removed and I had commented about this previously. I'm just taking a quick cursory look, but it seems that this has gotten more complicated and the dependency is on *.pb.go files, which in reality depend on *.proto files. I want to take a closer look at this, and which rules list depend on these prerequisites. @bquenin As a reminder, I don't want PRs merged without at least two LGTMs on it, and for the time being, one of those being me. Thanks.

bquenin:
@subfuzion Sorry about that, I was eager to make compilation time back to normal again.

ndegory:
@subfuzion please don't blame on Bertrand, I pushed him in merging it. But you're right, I should have waited for 2 LGTMs.

About the dependency, it is restored, there a dynamic variable for the *.proto source files, another for the *.pb.go targets and another for the *.pb.gw.go targets. Then there's an explicit rule for the compilation of these targets, and all install-* targets depend on these targets plus their own go source files.
So basically, we have something like that:

```
 *.proto --explicit rule--> *.pb.go-----------
              --explicit rule--> *.pb.gw.go--|
cmd/amp/**.go -------------------------------------> amp CLI target
api/rpc/**.go ---------------------|
vendor/* --------------------------|
```

PR #667, opened 2017-01-10
[v0.6.x] Account model

chrisccoy:
Work In Progress, but wanted some feedback on the initial approach


api/rpc/account/account.proto
```diff
@@ -246,3 +248,37 @@ message PermissionRequest {
...
   string resource_id = 4;
 }
+
+// Storage Types
+//This is the storage object
+
+message AccountEntry {
+  int64  id=1;
+  string name =2;
+  int64  type=3;
```
generalhenry: A bool would work for now.

api/rpc/account/account.proto
```diff
@@ -246,3 +248,78 @@ message PermissionRequest {
...
+  google.protobuf.Timestamp create_dt=5;
+}
+message ResourceSettingsEntry {
+  int64 id=1;
+  int64 resource_id=2; // FK->ResourceEntry.id
+  string key=3;
+  string value=4;
+  google.protobuf.Timestamp create_dt=5;
+}
+message TeamEntry {
```
generalhenry: team needs description

api/rpc/account/account.proto
chrisccoy: are suggesting something like "isUser"?

api/rpc/account/account.proto
```diff
@@ -246,3 +248,78 @@ message PermissionRequest {
...
+  int64 id=1;
+  int64 org_account_id=2; //FK->AccountEntry.id (type=org)
+  string name=3;
+  google.protobuf.Timestamp create_dt=4;
+}
+message TeamMemberEntry {
+  int64 id=1;
+  int64 user_account_id=2; //FK->AccountEntry.id (type=user)
+  int64 team_id=3; //FK->TeamEntry.id
+  int64 added_by_user_id=4; //FK->AccountEntry.id (type=user)
```
generalhenry: I'm not sure we need this here.  I like the idea.

api/rpc/account/account.proto
generalhenry: Yes.

subfuzion: No. Not isUser: `is_user`.

api/rpc/account/account.proto
chrisccoy: removed, we can add back if we need the audit capability

api/rpc/account/account.proto
chrisccoy: Added desc field

api/rpc/account/account.proto
chrisccoy: Changed this to be bool

api/rpc/account/account.proto
```diff
@@ -246,3 +248,78 @@ message PermissionRequest {
...
+// Storage Types
+//This is the storage object
+
+message AccountEntry {
+  int64  id=1;
+  string name =2;
+  bool  is_user=3; // (true=user, false=org)
+  string email =4;
+  string pw_hashcode =5;
+  bool is_verfified=6;
```
subfuzion: sp: is_verified

api/rpc/account/account.proto
```diff
@@ -246,3 +248,78 @@ message PermissionRequest {
...
   string resource_id = 4;
 }
+
+// Storage Types
+//This is the storage object
+
+message AccountEntry {
+  int64  id=1;
+  string name =2;
+  bool  is_user=3; // (true=user, false=org)
```
subfuzion: The default value of false should not imply the account is an org. I prefer a protobuf enum for this:

enum type {
  USER = 0;
  ORGANIZATION = 1;
}

subfuzion (commented):
I don't think we need to append *Entry to every entity to indicate it represents a single instance. I think it's fine to have `Account`, `Resource`, `Permission`, etc. We can pluralize collection messages or distinguish when it definitely improves clarity.

subfuzion (changes requested):
For the time being we can use UUID v1 for IDs, but we will eventually want an ID generation service (with some sense of being time ordered, maybe flake IDs or 64-bit [coreos/etcd IDs](https://github.com/coreos/etcd/blob/master/pkg/idutil/id.go).

chrisccoy:
Should I wrap the int64 into a seperate UUIDv1 message type or just leave it be?

subfuzion:
Let's just use string for now. String keeps the uuid human-readable, plus we don't have to be concerned with endian order, like I presume if we used bytes. I don't see a need to wrap the uuid in a dedicated message type; I'd just call the field id and type it as a string for the time being.

api/rpc/account/account.proto
```diff
@@ -171,7 +173,7 @@ message AccountRequest {
   string name = 1;
 }
 
-message Account {
+message AccountInfo {
```
subfuzion: I don't think you need to rename this to AccountInfo. For one, I don't think the data resources should be in the rpc package bundled with the service api resources, but all in another package under one flat namespace (like `amp/schema`, so when imported, looks like schema.Account, not to be confused with rpc.Account).

subfuzion (changes requested):

api/rpc/account/account.proto
subfuzion: Oops! You renamed it so not to conflict with the other Account message. Sorry, my bad. Let me chew on this a moment....

chrisccoy: :+1:  I like the idea of using a separate package for schemas, but was not sure how we wanted to organize this

subfuzion: Maybe put it under `data/schema`...

subfuzion: I don't see a need for AccountInfo.

api/rpc/account/account.proto
```diff
@@ -173,7 +173,7 @@ message AccountRequest {
   string name = 1;
 }
 
-message AccountInfo {
+message Account {
```
subfuzion, line 176: Hate to tell you this, but Account isn't needed here. @generalhenry already created an Account type for the service: https://github.com/appcelerator/amp/blob/v0.6.x/api/rpc/account/account.proto#L174

api/rpc/account/account.swagger.json
```diff
@@ -708,33 +708,21 @@
     "accountAccount": {
```
subfuzion, line 708: See above comment, but this should go away, too.

data/schema/account/account.proto
```diff
@@ -0,0 +1,77 @@
+syntax = "proto3";
```
subfuzion: Since you have this in the `schema` package (which I think is good), you might as well put this (and all the other types you create) under the flat `data/schema` directory (ie, no need for an `account` subdirectory).

data/schema/account/account.proto
```diff
@@ -0,0 +1,77 @@
...
+import "google/protobuf/timestamp.proto";
+
+// Account Schema Storage Types
+
+enum AccountType {
+  USER=0;
+  ORGANIZATION=1;
+}
+message Account {
+  string  id=1;
```
subfuzion: Nitpicking here about spaces (but then again, Go itself is quite opinionated, so we should be used to that) :p  You have two spaces before `id`, and none at all around the assignment operator itself. Should look like:

    string id = 1;

The [protobuf style guide](https://developers.google.com/protocol-buffers/docs/style) isn't explicit about it, but that's how it shows them, and that's consistent with all of our other proto files...

subfuzion (changes requested):
Thanks @chrisccoy ... getting there ...

data/schema/account/account.proto
bquenin: I'm going to fix the formatting issue

subfuzion (approved):
I think this is a good point to merge now and start working on implementation. 👍

PR #673, opened 2017-01-11
[v0.6.x] Add api stubs to account

generalhenry:
Adds api stubs to account with validation

### DIFF
--- a/api/rpc/account/account.go
+++ b/api/rpc/account/account.go
@@ -1,0 +1,357 @@
+package account
+
+import (
+	"fmt"
+	"net/mail"
+
+	context "golang.org/x/net/context"
+
+	pb "github.com/golang/protobuf/ptypes/empty"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"gopkg.in/hlandau/passlib.v1"
+)
+
+const codeLength = 8
+const hash = "$s2$16384$8$1$42JtddBgSqrJMwc3YuTNW+R+$ISfEF3jkvYQYk4AK/UFAxdqnmNFVeUw2gUVXEMBDAng=" // password
+
+// Validate validates SignUpRequest
+func (r *SignUpRequest) Validate() (err error) {
+	if r.Name == "" {
+		return grpc.Errorf(codes.InvalidArgument, "user name is mandatory")
+	}
+	address, err := mail.ParseAddress(r.Email)
+	if err != nil {
+		return grpc.Errorf(codes.InvalidArgument, err.Error())
+	}
+	r.Email = address.Address
+	if len(r.Password) < 8 {
+		return grpc.Errorf(codes.InvalidArgument, "password too weak")
+	}
+	return
+}
+
+// Validate validates VerificationRequest
+func (r *VerificationRequest) Validate() (err error) {
+	if len(r.Code) != codeLength {
+		return grpc.Errorf(codes.InvalidArgument, "invalid verification code")
+	}
+	return
+}
+
+// Validate validates OrganizationRequest
+func (r *OrganizationRequest) Validate() (err error) {
+	if r.Name == "" {
+		return grpc.Errorf(codes.InvalidArgument, "organization name is mandatory")
+	}
+	address, err := mail.ParseAddress(r.Email)
+	if err != nil {
+		return grpc.Errorf(codes.InvalidArgument, err.Error())
+	}
+	r.Email = address.Address
+	return
+}
+
+// Server is used to implement account.AccountServer
+type Server struct{}
+
+// SignUp implements account.SignUp
+func (s *Server) SignUp(ctx context.Context, in *SignUpRequest) (out *SessionReply, err error) {
+	out = &SessionReply{}
+	err = in.Validate()
+	if err != nil {
+		return nil, err
+	}
+	_, err = passlib.Hash(in.Password)
+	if err != nil {
+		return nil, grpc.Errorf(codes.Internal, "hashing error")
+	}
+	out.SessionKey = in.Name
+	return
+}
+
+// Verify implements account.Verify
+func (s *Server) Verify(ctx context.Context, in *VerificationRequest) (out *pb.Empty, err error) {
+	out = &pb.Empty{}
+	err = in.Validate()
+	if err != nil {
+		return nil, err
+	}
+	fmt.Println(in.Code)
+	return
+}
+
+// CreateOrganization implements account.CreateOranization
+func (s *Server) CreateOrganization(ctx context.Context, in *OrganizationRequest) (out *pb.Empty, err error) {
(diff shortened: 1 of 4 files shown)


api/rpc/account/account.go
```diff
@@ -0,0 +1,307 @@
+package account
+
+import (
+	"fmt"
+	"net/mail"
+
+	context "golang.org/x/net/context"
+
+	google_protobuf1 "github.com/golang/protobuf/ptypes/empty"
```
subfuzion: I know there are samples on the Internet using this convention because that's what protoc generates for `*.pb.go` file imports, ie, `google_protobuf1`, `google_protobuf2`, etc), but I don't think we need to follow that convention for human generated code. I'm fine with either of the following:

```
import (
        .  "github.com/golang/protobuf/ptypes/empty"    //->   Empty
        // or
        pb "github.com/golang/protobuf/ptypes/empty"    // -> pb.Empty
)
```
I'm fine with the dot import form simply because [empty.proto](https://github.com/google/protobuf/blob/master/src/google/protobuf/empty.proto) only exports the single Empty message anyway.

api/rpc/account/account.go
```diff
@@ -0,0 +1,307 @@
...
+func (s *Server) SignUp(ctx context.Context, in *SignUpRequest) (*SessionReply, error) {
+	if in.Name == "" {
+		return nil, grpc.Errorf(codes.InvalidArgument, "user name is mandatory")
+	}
+	address, err := mail.ParseAddress(in.Email)
+	if err != nil {
+		return nil, grpc.Errorf(codes.InvalidArgument, err.Error())
+	}
+	if len(in.Password) < 8 {
+		return nil, grpc.Errorf(codes.InvalidArgument, "password too weak")
```
subfuzion: I'm a little hesitant about validation logic implemented in server methods instead of as methods on the message types, and the potential effort across the board of keeping validation logic in sync. Ex:

```
func (m *SignupRequest) Validate() error {
    ...
}

if err := in.Validate(); err != nil {
        return nil, err;
}
```

While the advantage might not be high for any particular service/message, there is probably a benefit from consistency in using this as a general pattern. More complicated Validate functions might need to be parameterized for specific validation cases.

api/rpc/account/account.go
generalhenry: I went with 'pb'

api/rpc/account/account.go
generalhenry: I did a few as examples, I think converting the rest can be a follow up.  I'd like to get this in soon so the cli and ui hook ups can start.

api/rpc/account/account.go
```diff
@@ -0,0 +1,357 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"gopkg.in/hlandau/passlib.v1"
+)
+
+const codeLength = 8
+const hash = "$s2$16384$8$1$42JtddBgSqrJMwc3YuTNW+R+$ISfEF3jkvYQYk4AK/UFAxdqnmNFVeUw2gUVXEMBDAng=" // password
+
+// Validate validates SignUpRequest
+func (r *SignUpRequest) Validate() (err error) {
```
subfuzion: Don't really see a compelling need for named error return value in validators. It's basically ignored and requires an explicit terminal return statement that's just noise.

api/rpc/account/account.go
```diff
@@ -0,0 +1,357 @@
...
+	if r.Name == "" {
+		return grpc.Errorf(codes.InvalidArgument, "user name is mandatory")
+	}
+	address, err := mail.ParseAddress(r.Email)
+	if err != nil {
+		return grpc.Errorf(codes.InvalidArgument, err.Error())
+	}
+	r.Email = address.Address
+	if len(r.Password) < 8 {
```
subfuzion: Should be a const, even though we'll want to be able to supply configurable (and stronger) password policies in the future.

api/rpc/account/account.go
```diff
@@ -0,0 +1,357 @@
...
+		if err != nil {
+			return nil, grpc.Errorf(codes.InvalidArgument, err.Error())
+		}
+	}
+	if in.NewPassword != "" {
+		_, err := passlib.Verify(in.Password, hash)
+		if err != nil {
+			return nil, grpc.Errorf(codes.Unauthenticated, err.Error())
+		}
+		if len(in.NewPassword) < 8 {
```
subfuzion: These kinds of things should be in the validators. Validators can be parameterized if necessary to limit validation scope depending on operation. Validators can chain calls to fundamental Validate functions, like ValidatePassword, or perhaps validators types, like PasswordValidator with a Validate method. Not trying to be too hard on this PR, but since this will likely be the model for all the resources to follow, we probably want to get it right.

subfuzion (changes requested):

api/rpc/account/account.go
generalhenry: You always need a terminal return statement if you have a return signature.

without the named return:
``` go
func (r *VerificationRequest) Validate() error {
	if len(r.Code) != codeLength {
		return grpc.Errorf(codes.InvalidArgument, "invalid verification code")
	}
	return nil
}
```

subfuzion: Yes, I shouldn't call it noise so much as it obfuscates the fact that the named return is really irrelevant since we have specific returns already. In other words, `return nil` is clearer to me than the empty `return`. (I've contributed my [fair share](https://github.com/appcelerator/amp/graphs/contributors) of go code to the project; I do know how return works!  :p )

api/rpc/account/account.go
```diff
@@ -0,0 +1,311 @@
...
+			return nil, grpc.Errorf(codes.Internal, "hashing error")
+		}
+	}
+	return
+}
+
+// DeleteAccount implements account.DeleteAccount
+func (s *Server) DeleteAccount(ctx context.Context, in *AccountRequest) (out *pb.Empty, err error) {
+	out = &pb.Empty{}
+	if in.Name == "" {
```
subfuzion, line 121: I like that you moved all the validators out to their own file. Regarding using validators, it seems like we should be consistent in applying the pattern. Thus, we probably want a Validate method on AccountRequest as well for standardization, even though the validation is so simple.  The same principle would apply to all the following request messages, such as TeamRequest, where the same validation logic keeps getting applied (we may want to create a validation context or enum parameter for CRUD-type operations that validate different subsets of the message fields if we don't want to create distinct message types).

generalhenry: I was wanting to push those to a follow up https://github.com/appcelerator/amp/issues/677

subfuzion: Gotcha, ok thanks. The rest LGTM.

subfuzion (approved):

Issue #591, opened 2016-12-15
Standardize output format using tabwriter


neha-viswanathan:
The outputs of different commands are formatted differently. Only some are using `tabwriter` package (like: `topic` and `kv`)

freignat91:
done for new cluster cli commands, it stays stats and stacks to move on it,
For amp pf command (monitor), we can wait the cluster cli is validated

freignat91:
waiting until #583 is merged to fix this issue

subfuzion:
@freignat91 #583 was merged.

PR #675, opened 2017-01-11
Use tab writer

freignat91:
related to #591

use tabwriter for stack ls and am pf monitor commands

test:
- execute amp pf monitor (see no change regarding the previous version without tabwrite)
- create several stacks
- execute: amp stack ls (see no change regarding the previous version without tabwrite)

regression tests:
make test

### DIFF
--- a/cmd/amp/platform_manager.go
+++ b/cmd/amp/platform_manager.go
@@ -19,6 +19,7 @@ import (
 	"github.com/docker/docker/client"
 	"github.com/fatih/color"
 	"golang.org/x/net/context"
+	"text/tabwriter"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


subfuzion (changes requested):
Thanks for the PR, @freignat91. My thoughts at the moment is that I wanted a more comprehensive CLI UI package for standardizing our output formatting and colors, etc. We also want to ensure we can disable terminal ansi codes in case the command is not running in tty mode, ie, the command is piping output (which admittedly pertains more to logs and stats than to the platform commands, but I can see long list output getting piped to less, grep, etc). I'd like to make sure we get a few others to chime in as to whether we want to accept the PR as is and fix later vs postpone. I'd prefer to postpone (plus this isn't a priority for the current team permission model milestone).

generalhenry:
I think the next step is a single command that has a table, color, ansi codes etc . . . that we perfect before replicating it across out CLI.  Making that kind of CLI improvement is out of scope for v0.6.x unless we find that we need it for the account commands.

subfuzion (approved):
Merging since it will be good to get the enhancements. Just keep in mind that we'll need to refactor the CLI output itself, per my previous comments.

Issue #621, opened 2017-01-02
Base AMP images on Alpine 3.5


ndegory:
Alpine 3.5 is out, many AMP infra images are based on Alpine 3.4, they should rebased on 3.5.

- [x] influxdb (influxdb-1.1.2)
- [x] telegraf (telegraf-1.1.2)
- [x] kapacitor (kapacitor-1.1.1)
- [x] grafana (grafana-4.1.1)
- [x] elasticsearch (5.1.1-3)
- [x] etcd (3.1.0-rc.1-1)
- [x] haproxy (1.0.4)
- [x] amp (PR #682)
- [x] update references in AMP CLI sources

ndegory:
Haproxy image based on Alpine is part of #699, will leave it there.

PR #682, opened 2017-01-12
Alpine 3.5

### LINKED ISSUE #663: Remove python dependency in AMP images
The appcelerator alpine image includes the python libs because of the envtpl script used in several AMP images (influxdata/*, grafana). Removing it would result in smaller images.

ndegory:
ref #621 
close #663

- build AMP image based on Alpine 3.5
- use newly built AMP infra images with Alpine 3.5

test with:
```
$ make clean install
$ make TAG=local build-image
$ amp pf pull && amp pf start --local && make test
```


subfuzion (approved):
👍

Issue #679, opened 2017-01-11
Account cli basics


generalhenry:
ref: https://github.com/appcelerator/amp/issues/660

Add interactive flows, input validation and basic output to account cli methods

Issue #678, opened 2017-01-11
Account type mocks


generalhenry:
ref: https://github.com/appcelerator/amp/issues/660
Follow up to https://github.com/appcelerator/amp/pull/673

Add basic mocks to the api

For example the Login api checks a password against a single hash and returns a dummy session token.

We should create a set of account in memory and have the api respond with them.

PR #689, opened 2017-01-12
[v0.6.x] Added account commands - WIP

neha-viswanathan:
Ref: #678, #679 

Added CLI implementation of `account` command - WIP


cmd/amp/account.go
```diff
@@ -40,198 +55,759 @@ var (
 	switchRoleCmd = &cobra.Command{
 		Use:   "switch [ORGANIZATION]",
```
subfuzion: This shouldn't be an optional parameter. And switch is for ACCOUNT, not ORGANIZATION. You might switch to the account that is the primary account directly associated with your account id (ie, your individual account), or you might switch to an organization account.

cmd/amp/account.go
```diff
@@ -40,198 +55,759 @@ var (
...
 		Short: "Switch primary organization",
+		Long:  `Switch command changes the primary organization of a resource to the specified organization.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return switchRole(AMP, cmd, args)
+		},
 	}
 
 	createOrganizationCmd = &cobra.Command{
-		Use:   "create organization [NAME] [EMAIL]",
+		Use:   "createorg [NAME] [EMAIL]",
```
subfuzion: I don't like the way these are coming out ... `createorg`, `listteam`, etc. We need to think through a better command-noun-verb-attribute model, but I would think something more like: `amp account create --organization` or `amp account create --type=organization` (default `type=user`)... Let's discuss more.

subfuzion (changes requested):

cmd/amp/account.go
neha-viswanathan: Working on it. Will put it up as a separate PR.

cmd/amp/account.go
neha-viswanathan: I have added the command usage in the doc. Waiting for consensus from the team.

subfuzion: @neha-viswanathan I've been reviewing and taking a cut at it.

PR #691, opened 2017-01-12
[v0.6.x] Add basic mocks to some routes

generalhenry:
gets things rolling on: https://github.com/appcelerator/amp/issues/678


api/rpc/account/mock.go
```diff
@@ -0,0 +1,52 @@
...
+
+import "github.com/appcelerator/amp/data/schema"
+
+var accounts = []*schema.Account{
+	{
+		Id:         "0",
+		Name:       "generalhenry",
+		Type:       schema.AccountType_USER,
+		Email:      "hallentilford@axway.com",
+		PwHashcode: hash,
```
subfuzion: I should have commented before, but let's refactor the field name (it's so idiomatic to just say password hash): PwHashCode -> `PasswordHash`.

generalhenry:
I'm going to convert the inline data access code into an account data package, so the account data package can then be converted to using a real backing store.

api/rpc/account/account.go
```diff
@@ -27,6 +29,15 @@ func (s *Server) SignUp(ctx context.Context, in *SignUpRequest) (out *SessionRep
...
 	}
+	account := &schema.Account{
+		Id:         strconv.Itoa(len(accounts)),
+		Name:       in.Name,
+		Type:       schema.AccountType_USER,
+		Email:      in.Email,
+		PwHashcode: hash,
+		IsVerified: false,
+	}
+	accounts = append(accounts, account)
```
subfuzion: You should be going through an interface that uses the mock data (and ultimately actual data) to give you `accounts` ... and that applies to all the usages of mock data. This would look like how we create an [instance of the Storage interface](https://github.com/appcelerator/amp/blob/master/api/server/server.go#L116) with our [etcd implementation](https://github.com/appcelerator/amp/blob/master/data/storage/etcd/store.go#L38). In this case, you would create an instance of the account access code using the mock data.

subfuzion (changes requested):

generalhenry:
Next step, create data interface.

chrisccoy:
Should we go ahead and merge this to [0.6x] and then start doing incremental PR's from there?

PR #694, opened 2017-01-13
[v0.6.x] Account command - signup

JosephGJ:
ref: #679 

Implements the `signup` command on the CLI. User enters their username, email and password which is validated using `validators.go`. The current validations are:

- Username must not be empty
- Email must pass the `ParseAddress` function of the `net/mail` package
- Password must pass the `safe.Check` and have at least Medium strength

To test
```
$ make install
$ amp pf start --local
$ amp account signup
```


chrisccoy:
:+1: The only suggestion from me would be to provide some better guidance around the password criteria. 

`error in getting password: password is too weak`

The user must currently guess as to the password rules.

JosephGJ:
A password error now returns a message to specify a password of good strength. 

There was also a minor error with the password, wherein if the user interrupted the sequence it would output an rpc error, now it returns `error: interrupted`.

These have been fixed.

JosephGJ:
Modified the command interactivity to align with the Account CLI UX within the drive. Password is not requested on signup and will be set via the url sent in the verification email or with the verify command.

PR #695, opened 2017-01-13
Updated long message for fn command

neha-viswanathan:
Updated long message for `function` [alias: `fn`] command

### DIFF
--- a/cmd/amp/function.go
+++ b/cmd/amp/function.go
@@ -17,6 +17,7 @@ var (
 	functionCmd = &cobra.Command{
 		Use:     "function",
 		Short:   "function operations",
+		Long:    `Function command manages all function-related operations.`,
 		Aliases: []string{"fn"},
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			return AMP.Connect()
@@ -26,6 +27,8 @@ var (
 	createFunctionCmd = &cobra.Command{
 		Use:   "create NAME IMAGE",
 		Short: "Create a function",
+		Long: `The create command registers a function with the specified name and image.
+	If successful, a function id is returned.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return createFunction(AMP, cmd, args)
 		},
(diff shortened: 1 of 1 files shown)


cmd/amp/function.go
```diff
@@ -26,6 +27,8 @@ var (
 	createFunctionCmd = &cobra.Command{
 		Use:   "create NAME IMAGE",
 		Short: "Create a function",
+		Long: `Create command registers a function into ETCD, with the specified name and image.
```
subfuzion: The create command registers a function with the specified name and image. If successful, a function ID is returned.

cmd/amp/function.go
```diff
@@ -34,6 +37,7 @@ var (
 	listFunctionCmd = &cobra.Command{
 		Use:   "ls",
 		Short: "List functions",
+		Long:  `List command displays all the functions available in ETCD.`,
```
subfuzion: The list command displays all registered functions.

cmd/amp/function.go
```diff
@@ -42,6 +46,7 @@ var (
 	removeFunctionCmd = &cobra.Command{
 		Use:   "rm FUNCTION",
 		Short: "Remove a function",
+		Long:  `Remove command unregisters the specified function id from ETCD.`,
```
subfuzion: The remove command unregisters the specified function.

cmd/amp/function.go
```diff
@@ -42,6 +46,7 @@ var (
 	removeFunctionCmd = &cobra.Command{
 		Use:   "rm FUNCTION",
```
subfuzion: FUNCTION should really be something `FUNC-NAME-OR-ID` but I don't think we support name right now (or even ID fragments, like the other commands. To make it clear, we should probably say FUNC-ID for now.

subfuzion (changes requested):

cmd/amp/function.go
neha-viswanathan: Done.

cmd/amp/function.go
neha-viswanathan: Done.

cmd/amp/function.go
neha-viswanathan: Done.

cmd/amp/function.go
neha-viswanathan: Done.

subfuzion (approved):
👍

PR #702, opened 2017-01-19
Account api etcd

chrisccoy:
WIP: I wanted to review this against the mock branch to make sure we are on the same page. I made a few changes, particularly moving the interface and impl modules under its own package. If we plan to proliferate this pattern, we will need this level of isolation.


subfuzion (approved):
@chrisccoy Merging for now, but per conversation yesterday I don't want any energy getting spent on mocks going forward. Let's just use etcd.

chrisccoy:
@subfuzion 10-4 on mocks

PR #704, opened 2017-01-23
Docker gotools

subfuzion:
Collection of Go tools in an alpine image. Used by the amp toolchain. Among other things, this is specifically for building alpines images at the moment. We may want to generate ubuntu binaries again down the road.


ndegory (approved):
LGTM

PR #705, opened 2017-01-23
Fix Makefile glide rules

### LINKED ISSUE #133: Fix file owner on Linux when running containerized tools on mounted volume
@chrisccoy identified an issue that occurs on Linux (but not Docker for Mac): after running `make install-deps`, which runs glide in a container as root to update vendor dependencies by mounting the project directory, files are owned by `root:root`.

In fact, this error occurs for any files generated by containerized tools when writing on the mounted volume. This causes a variety of issues since the user can no longer remove the files without running sudo, such as when switching to a branch that doesn't have them, or updating vendor dependencies, etc. Files affected include all the generated protobuf `*.pb.go` files, the mounted glide cache directories (`$HOME/.glide`, `./.glide`), and `amp` and `amplifier`.

The expected behavior is that files have ownership attributes corresponding to the current UID:GID when running any of the containerized tools with make.

subfuzion:
Fixes #133 

Fixes permission issue for glide cache in home directory in container by mapping the cache to /tmp/glide and most importantly (thanks @ndegory) ensuring that the `~/.glide` on the host exists or else the mount will fail.

# Verification

    $ make install-deps
    $ make install-deps (AGAIN)
    # VERIFY on a LINUX system that you don't have any permission issues and nothing in `~/.glide/cache` or `./vendor` is owned by `root`.


chrisccoy (approved):
Works for me. nothing owned by root

ndegory:
Tests passed on Mac OS  ✓
Tests passed on Linux ✓

ndegory (approved):
LGTM

PR #706, opened 2017-01-23
docker-dind

subfuzion:
Convenience image based on latest Docker in Docker with experimental flag set and a few network utilities (like iproute2, dig and drill).

### DIFF
--- a/images/dind/README.md
+++ b/images/dind/README.md
@@ -1,0 +1,37 @@
+# dind
+
+[On Docker Hub](https://hub.docker.com/r/subfuzion/dind/). Cloned from [subfuzion/docker-dind](https://github.com/subfuzion/docker-dind).
+Convenience image based on latest Docker in Docker with experimental flag set and
+a few network utilities (like iproute2, dig and drill).
+
+Why is this useful? Right now if you want to start dind with the experimental flag
+you have to explicitly supply it as follows:
+
+    $ docker run --detach --privileged --name=dind docker:dind --experimental
+    $ docker exec -it dind sh
+    / # docker version
+    Client:
+     Version:      1.13.0
+     API version:  1.25
+     Go version:   go1.7.3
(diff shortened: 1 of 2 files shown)


images/dind/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+docker build -t dind .
```
ndegory: shouldn't it be appcelerator/dind?

images/dind/README.md
```diff
@@ -0,0 +1,37 @@
+# dind
+
+[On Docker Hub](https://hub.docker.com/r/subfuzion/dind/). Cloned from [subfuzion/docker-dind](https://github.com/subfuzion/docker-dind).
```
ndegory: we should create an appcelerator/dind repo on Docker hub

images/dind/README.md
```diff
@@ -0,0 +1,37 @@
...
+     Go version:   go1.7.3
+     Git commit:   49bf474
+     Built:        Wed Jan 18 16:20:26 2017
+     OS/Arch:      linux/amd64
+     Experimental: true
+
+You can see from the last line that the daemon was started in experimental mode.
+With this image, however, experimental mode is the default:
+
+    $ docker run --detach --privileged --name=dind subfuzion:dind
```
ndegory: and change the cmd to use appcelerator/dind instead of subfuzion:dind

ndegory (changes requested):

images/dind/build.sh
subfuzion: Yes, it was already removed, forgot to push.

images/dind/README.md
subfuzion: Yes, as discussed, will leave that up to you. Thanks.

ndegory (approved):

PR #708, opened 2017-01-23
[v0.6.x] Account command - verify

JosephGJ:
ref #679 

Implements the `verify` command on the CLI. User enters the code they were sent on their verification email and then enters their username and creates their password which are validated by `validators.go`. The validations will be:

- Code must be exactly 8 characters long.
- Username must match an existing username
- Password must be at least 8 characters long and contain 3 unique sets of characters (lowercase, uppercase, number, symbol).

Currently, only the code and password can be verified.

To test
```
make install
amp pf start --local
amp account verify
```


api/rpc/account/account.proto
```diff
@@ -146,7 +146,8 @@ message SignUpRequest {
 
 message VerificationRequest {
   string name = 1;
-  string code = 2;
+  string password = 2;
```
subfuzion, line 155: Since we should only be transmitting a hash of the password, we should specify an appropriately sized fixed length byte array (probably 12 bytes, but you and @generalhenry should confirm and coordinate on this).

subfuzion (changes requested):
LGTM other than one comment for resolution on using a fixed length byte array for storing the password hash. @generalhenry, please merge once you're also satisfied.

api/rpc/account/account.proto
generalhenry: Why would we shift hashing to the client side?

generalhenry: Discusion: 
http://security.stackexchange.com/questions/53594/why-is-client-side-hashing-of-a-password-so-uncommon

If the password hash is sent in place of the password then the password hash becomes in effect a plain text password stored in the database.  Which means if the database if compromised then the accounts can be accessed.

It's also possible to double hash (client side and server side) but that only helps if the user reuses the password and either the tls or the server is compromised.

subfuzion: We should never trust what is supplied by a random client, I'm only saying there is no need for **our own** client to send a plaintext password -- even over TLS. Our server shouldn't receive plaintext passwords, which we can mitigate by requiring clients to send a fixed byte length hashcode. However after the server receives it, we would want to follow best practices with respect to hashing before storing in our system. That being said, maybe it's not worth the extra effort, especially right now; we can certainly revisit this later as part of a security practices review.

No need to obsess it at the moment since we do need to ensure that the backend hashes regardless, but keep in mind that TLS only protects the password during transport, it doesn't protect it against breaches on the server. Since we use haproxy for SSL termination, HTTPS will get sent as unencrypted HTTP to our services (which we might be able to mitigate with the `--opt encrypted` for the frontend overlay network), and in any case the service receives a plaintext password, which leaves the door open to a potential breaches that somehow might obtain passwords in transit or stored in proxy logs, etc.

generalhenry: As for string vs byte I don't see a way to specify a fixed length either way: https://developers.google.com/protocol-buffers/docs/proto3#scalar I would stick with string since our hashing returns a string.

subfuzion: Ah, good point. We can't specify the length of either bytes or strings in the protobuf definion, only validate it at the receiver.

PR #710, opened 2017-01-24
[v0.6.x] Message queuer interface

bquenin:
Add an interface to access message queuer

## Verification
    $ amp pf stop
    $ make TAG=local build-image
    $ amp pf start --local
    $ make test


pkg/mq/interface.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Close connection
+	Close() error
+
+	// Publish
+	Publish(subject string, data []byte) error
+	PublishAsync(subject string, data []byte, ah AckHandler) (string, error)
+
+	// Subscribe
+	Subscribe(subject string, mh MsgHandler, opts ...SubscriptionOption) (Subscription, error)
```
generalhenry: Do you plan on supporting a channel style interface and/or sync style interface as well?

bquenin: Not at the moment, but if requested I can give it a shot

pkg/mq/interface.go
```diff
@@ -0,0 +1,163 @@
...
```
subfuzion: @bquenin I do think there's value in also having the ability to subscribe to and publish raw bytes, but I'd like an interface like the one I did for etcd storage that uses protocol buffer messages (see [here](https://github.com/appcelerator/amp/blob/master/data/storage/interface.go)). I would prefer that the interface explicitly call that out, ex: `PublishRaw()`, `SubscribeRaw`, and let `Publish()` and `Subscribe` be used for `proto.Message`. Or we can have `PublishProto()` and `SubscribeProto()`. That will certainly be our own default case, but that might make sense if we decide to also add `PublishJSON()`, `SubscribeJSON()`, etc, to support various strongly typed APIs in addition to raw bytes...

api/rpc/logs/logs.go
```diff
@@ -72,7 +71,7 @@ const (
 type Server struct {
 	Es            *elasticsearch.Elasticsearch
 	Store         storage.Interface
-	NatsStreaming ns.NatsStreaming
+	MessageQueuer mq.Interface
```
subfuzion: No offense, but `MessageQueuer` is pretty unwieldy ;).  I would just call it `MQ` and if you want to make things consistent, you can rename `Store` here to `KV`.

pkg/mq/interface.go
subfuzion: Do you plan to add [queue](https://github.com/nats-io/go-nats-streaming#queue-groups/) support in a separate PR?

api/rpc/logs/logs.go
```diff
@@ -154,8 +153,8 @@ func (s *Server) Get(ctx context.Context, in *GetRequest) (*GetReply, error) {
 func (s *Server) GetStream(in *GetRequest, stream Logs_GetStreamServer) error {
 	log.Println("rpc-logs: GetStream", in.String())
 
-	sub, err := s.NatsStreaming.GetClient().Subscribe(amp.NatsLogsTopic, func(msg *stan.Msg) {
-		entry, err := parseProtoLogEntry(msg.Data)
+	sub, err := s.MessageQueuer.Subscribe(amp.MQLogsTopic, func(msg []byte) {
```
subfuzion: MQLogsTopic ... I don't think the `MQ` part is needed, and `Topic` is part of the Kafka legacy. I think it would be better to rename to `LogsQueue` (at least once you add queue support in addition to pub/sub, as mentioned in another comment in this review).

subfuzion (changes requested):
Thanks for working on this @bquenin. Just a few comments for now...

pkg/mq/interface.go
bquenin: It's implemented here: https://github.com/appcelerator/amp/pull/710/files/74ea62ba211280cfcf4173a7eb13249e0684a260#diff-3d5bb591c85059b41e5cf90eef60af85R21

subfuzion (approved):
Awesome, thanks! 👍

PR #714, opened 2017-01-25
[v0.6.x] Account command - login

JosephGJ:
ref #679

Implements the login command on the CLI. User enters their registered username and password which are validated by `validators.go`. The validations will be:

- Username must not be empty

Both Username and Password will be validated more thoroughly when the backend is implemented

To test
```
make install
amp pf start --local
amp account login
```


generalhenry (commented):
Resolve the conflicts and it should be good

JosephGJ:
fixed.

Just to note, the command will throw an rpc error atm as the hashing is not implemented correctly.

generalhenry (approved):

PR #715, opened 2017-01-25
amp infra stack yml files for tests

freignat91:
# Do  not merge this PR, it's for test only:

docker compose files sample for amp infra stacks:
core, function, log, monitor (admin on going)
to start them:

- make install
- make TAG=local build-image
- export ampVersion=local
- launch amp pf monitor, to monitor the services
- docker stack deploy ampcore -c ampcore.yml
- docker stack deploy ampfunction -c ampfunction.yml
- docker stack deploy amplog -c amplog.yml
- docker stack deploy ampmonitor -c ampmonitor.yml
- use docker stack rm ampxxx to rmeove a stack

ampcore should be started first and stopped last (it's not controlled right now)

amplifier image should be rebuild to suppress dependencies on elasticsearch and influxdb, so log and monitoring don't work anymore. The connection to these two database will be dynamic in the future.

ampcore.var will contain all the version variables and will be used in future version.

### DIFF
--- a/ampadmin.yml
+++ b/ampadmin.yml
@@ -1,0 +1,47 @@
+version: "3"
+
+services:
+
+    server:
+        image: appcelerator/amp:${ampVersion:-latest}
+        command: ["adm-server"]
+        deploy:
+            mode: replicated
+            replicas: 1
+            labels:
+                io.amp.role: "amp.InfrastructureRole"
+                io.amp.mapping: "amplifier:tcp:31315:31315"
+            placement:
+                constraints: [node.role == manager]
+        volumes:
+            - /var/run/docker.sock:/var/run/docker.sock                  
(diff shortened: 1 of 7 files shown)


subfuzion:
@freignat91 Actually I would like to add these. How about you change the PR to merge this with the `stacksv3` branch instead of `master`?

freignat91:
@subfuzion  if i do so, the amplifier won't be able to connect to InfluxDB and elasticsearch.
Because the ampcore.yml works only if amplifier doesn't connect to these two.
I'm about to create a true PR with the stack 1.13 (begining of next week, i think) don't you think it's better to wait?

freignat91:
all is in stacksv3 branch now, this PR can be closed

PR #717, opened 2017-01-25
[v0.6.x]Password change for Account command

neha-viswanathan:
Ref: #679
Added password change for account command

To test:
```
amp account password-change
```


cmd/amp/account.go
```diff
@@ -193,3 +228,61 @@ func getEmailAddress() (email string, err error) {
...
+		color.Set(color.FgRed, color.Bold)
+		fmt.Println("Password is mandatory. Try again!")
+		color.Unset()
+		fmt.Println("")
+		return getPwd()
+	}
+	err = account.CheckPasswordStrength(password)
+	if err != nil {
+		color.Set(color.FgRed, color.Bold)
+		fmt.Println("Password must be at least 8 characters long. Try again!")
```
generalhenry: It should print the error text it got from CheckPasswordStrength

api/rpc/account/account.proto
```diff
@@ -259,3 +265,10 @@ message PasswordResetRequest {
   string username = 1;
   string email = 2;
 }
+
+message PasswordChangeRequest {
+  string username = 1;
+  string existingPassword = 2;
+  string newPassword = 3;
+  string confirmPassword = 4;
```
generalhenry: confirmPassword should only be part of the UI not the API

neha-viswanathan: Yes. Makes sense. I have updated it.

cmd/amp/account.go
```diff
@@ -186,6 +196,29 @@ func getCode() (code string) {
...
+	getConfirmNewPwd(newPwd)
+	request := &account.PasswordChangeRequest{
+		Username:         username,
+		ExistingPassword: existingPwd,
+		NewPassword:      newPwd,
+	}
+	client := account.NewAccountServiceClient(amp.Conn)
+	_, err := client.PasswordChange(context.Background(), request)
+	if err != nil {
+		return err
```
JosephGJ: `return fmt.Errorf("server error: %v", err)`

This should be consistent with the other commands.

JosephGJ (approved):
LGTM

PR #723, opened 2017-01-27
[v0.6.x] Account command - forgot-login

JosephGJ:
ref: #679

Implements the forgot-login command on the CLI. User enters their email, which is validated using `validators.go`, and is then sent an email with their corresponding username . The current validations are:

 - Email must pass the ParseAddress function of the net/mail package

To test
```
$ make install
$ amp pf start --local
$ amp account forgot-login
```


api/rpc/account/account.proto
```diff
@@ -20,6 +20,11 @@ service AccountService {
       body: "*"
     };
   };
+  rpc ForgotLogin (ForgotLoginRequest) returns (google.protobuf.Empty) {
+    option (google.api.http) = {
+      get: "/v1/account/{email}"
```
generalhenry: This means we need to make emails unique and make emails invalid user/org names.

It's likely easier overall to make the url more unique.

'/v1/findAccount/' with the email in the url params (less likely to end up encoded wrong)

JosephGJ: Fixed with requested change.

PR #725, opened 2017-01-30
Refactor make

subfuzion:
This PR containerizes the build toolchain (except for make itself) and cleans up related scripts and the makefile. One of the main goals is to be in full control of the build environment and be able to build a containerized cli to access amp.

The main rules for this PR are:

    $ make -f Makefile.refactor.make install-deps
    $ make -f Makefile.refactor.make protoc
    $ make -f Makefile.refactor.make build-cli

The `build-cli` rule depends on glide and protoc targets, so that should be all that you need to run.

The end result of this is that the following image for the CLI is built: `appcelerator/amp`:

To run the CLI:

    $ docker run -it --rm appcelerator/amp

This PR is an interim step. The next PR will run make itself in the container, as well as the other tools, instead of spinning up separate containers for each.


hack/amptools
```diff
@@ -11,9 +11,26 @@ export BUILD="${BUILD:-$(git rev-parse HEAD | cut -c1-8)}"
 export OWNER="${OWNER:-appcelerator}"
 export REPO="${REPO:-github.com/$OWNER/amp}"
 
-docker image list amptools | grep -q amptools || $(dirname $0)/image-for-local-user
+TOOLS=appcelerator/amptools:latest
```
ndegory: will need a version, can be done in another PR
this way we can also tag the local amptools image with same version, and the rebuild of this image can be triggered when this version is absent (with `latest` once the image is built we won't rebuild it even if the appcelerator/amptools image has been updated)

hack/image-for-local-user
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
ndegory: file should be removed, not used anymore

images/amptools/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM appcelerator/gotools
```
ndegory: not sure about the state of `latest` for the gotools image. Version 1.3.0 is the stable one.

ndegory (approved):
LGTM

PR #727, opened 2017-01-31
[v0.6.x] Account api data interface

generalhenry:
Adds data interface to account with etcd implimentation


chrisccoy (approved):
:+1: 
Only commenting on Account support, since those are the funcs under test.

api/rpc/account/account.go
```diff
@@ -23,10 +34,17 @@ func (s *Server) SignUp(ctx context.Context, in *SignUpRequest) (out *SessionRep
 	if err != nil {
 		return nil, err
 	}
-	_, err = passlib.Hash(in.Password)
+	hash, err := passlib.Hash(in.Password)
```
JosephGJ: So, we are setting the password in the signup command? I thought we would set the password via the email verification link / verify command? If we are setting the password in the signup command, `account.proto` will need to be modified as the field no longer exists in `SignUpRequest`.

generalhenry: We are no longer setting password on signup, I'll update.

chrisccoy:
@generalhenry not sure with this build. My branch is rebased from this one and is working. I think the build just needs a reboot here?

chrisccoy:
@generalhenry seems like the merge had a conflict. I originally thought it was just the intermittent build issues, but on closer inspection seems to have a conflict

chrisccoy:
I attemtped to resolve the conflict, but unfortunately it goes a bit deeper:
```
# github.com/appcelerator/amp/api/rpc/account
api/rpc/account/account.go:37: undefined: hash
api/rpc/account/account.go:56: undefined: hash
api/rpc/account/account.go:95: in.Email undefined (type *PasswordChangeRequest has no field or method Email)
api/rpc/account/account.go:125: in.Password undefined (type *OrganizationRequest has no field or method Password)
api/rpc/account/account.go:129: out.SessionKey undefined (type *empty.Empty has no field or method SessionKey)
Makefile:149: recipe for target 'install-cli' failed
make: *** [install-cli] Error 2
```

bquenin:
I think we can close this PR for now

PR #726, opened 2017-01-31
Run make in container

subfuzion:
This PR provides the support to run make in a container to build the cli. Among other thing, it optimizes the use of `find` in the makefile to significantly improve performance with a mounted volume in the container.

    $ docker build -t appcelerator/gotools images/gotools
    $ docker build -t appcelerator/amptools images/amptools
    $ hack/amptools sudo make -f Makefile.refactor.make build-cli
    $ docker run -it --rm appcelerator/cli

Part of a work in progress.

PR #728, opened 2017-01-31
Run make in container - doesn't need sudo anymore

ndegory:
Follows #726 

- when running a build task, there's no need for an explicit sudo, the Makefile can add it if a docker command is executed (the only place where sudo is needed)
- tagged version of appcelerator/gotools and appcelerator/amptools

how to test:
- hack/amptools make -f Makefile.refactor.make build-cli # no more sudo, should build the cli but not the vendors (they should be already up to date)
- rm ./api/rpc/service/service.pb.go
- hack/amptools make -f Makefile.refactor.make proto # should compile the missing proto file
- rm -rf vendor
- hack/amptools make -f Makefile.refactor.make build-cli # should install vendors and build the CLI
- docker run -it --rm appcelerator/amp info # should display version and build number (31f985d1)

### DIFF
--- a/Makefile.refactor.make
+++ b/Makefile.refactor.make
@@ -95,10 +95,16 @@ CLIBINARY=$(CLI).alpine
 CLIIMG := appcelerator/amp
 CLITARGET := $(CMDDIR)/$(CLI)/$(CLIBINARY)
 CLISRC := $(shell find ./cmd/amp -type f -name '*.go')
+# to run the docker command with sudo, set the with_sudo variable
+ifdef with_sudo
+DOCKER_CMD := sudo docker
+else
+DOCKER_CMD := docker
+endif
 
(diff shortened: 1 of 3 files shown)


ndegory:
one thing I noticed, not sure when it has been introduced:
after a make -f Makefile.refactor.make clean-protoc and hack/amptools sudo make -f Makefile.refactor.make protoc, the swagger files are slightly changed (the format property disappears).

subfuzion:
Ok, I don't consider swagger to be a reason to hold this up right now. It's possibly due to us using a later versions of protoc and grpc-gateway or possibly a misconfiguration issue, but this is something @generalhenry can evaluate separately and open an issue around it, if necessary.

subfuzion (approved):
Good update! 👍

PR #730, opened 2017-02-02
Refactor make build-server, all

subfuzion:
Adds rule for `build-server`, supports `make build` and just `make`. Updates `clean* rules`.

Now generates separate images for `amp` and `amplifier`. The intention is to lay the groundwork here to have `amp` (cli) run in a container attached to the same overlay network that `amplifier` is running in the platform swarm.

## Verification

Touch various *.proto and *.go files in related source directories (api, cmd, config, test), touch `glide.yaml` or remove `vendor`, etc., and run `make -f Makefile.refactor.make` for default `all` rule, or explicitly test `build` or various other rules in the Makefile.

Ex:

    $ hack/amptools make -f Makefile.refactor.make clean
    $ hack/amptools make -f Makefile.refactor.make

### DIFF
--- a/cmd/amplifier/Dockerfile
+++ b/cmd/amplifier/Dockerfile
@@ -1,0 +1,3 @@
+FROM alpine
+COPY amplifier.alpine /usr/local/bin/amp
+ENTRYPOINT [ "amplifier" ]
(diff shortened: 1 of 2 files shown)


cmd/amplifier/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM alpine
+COPY amplifier.alpine /usr/local/bin/amp
```
ndegory: should be /usr/local/bin/amplifier

ndegory (changes requested):

ndegory (approved):

PR #734, opened 2017-02-03
Stack 1.13

freignat91:
Related to #733 

#738 should be merged first

refactor amp stack to use Docker 1.13 way of managing them.

to test:

- amp pf stop amp (stop amp old way)
- make install (clean before could be required)
- make TAG=local build-image
- amp pf start --local all

- - amp pf monitor (in the separated console)
- - amp pf start --local -> start only ampcore
- - amp logs -> error 
- - amp pf start amplog --local (wait for stack started)
- - amp logs --infra -> a lot of logs...
- - amp pf stop amplog
- - amp stats -> error
- - amp pf start ampmonitor --local (wait for stack started, and wait a while to have some data)
- - amp stats -> a lot of metrics
- - amp pf stop ampmonitor
- - amp pf start ampfunction --local
- - cd examples/functions/demo-function
- - make image
- - amp fn create test appcelerator/amp-demo-function
- - cat Makefile | curl fn.ampfunction.local.appcelerator.io/test --data-binary @-
- -> function executed
- - amp pf stop ampfunction
- - cd ../../.. (back to amp home)
- - amp pf start ampadmin --local (wait stack is started)
- - amp node ls -> see your host information
- - amp node count -> see your stopped container
- - amp node purge --container -> container deleted
- - amp node count -> see you don't have stopped container anymore
- - amp pf stop ampadmin
- - amp stack up test -c api/rpc/stack/test_samples/sample-03.yml (wait stack is started)
- - curl www.test.local.appcelerator.io/ping -> pong
- - amp stack ps test  -> list task
- - amp stack services test -> list services
- - amp stack rm test
- - amp stack ps ampcore -> see it's working also for infra stack
- - amp stack services ampcore -> same

autmated tests:
- amp pf stop (wait all is down)
- amp pf start all --local -v  (wait all is up)
- make test

-

PR #735, opened 2017-02-03
Updated vendor

subfuzion:
This PR should be merged before #734 

Clean vendor update after running through and testing in containerized toolchain. The PR also includes updates to generated swagger files. As mentioned as part of a previous PR, these should be verified by @generalhenry. If the wrong version of grpc-gateway is getting used, an issue should be opened separately.

## Verification

    $ hack/amptools make -f Makefile.refactor.make cleanall
    $ hack/amptools make -f Makefile.refactor.make build


glide.lock
```diff
@@ -256,7 +260,7 @@ imports:
 - name: github.com/spf13/jwalterweatherman
   version: 33c24e77fb80341fe7130ee7c594256ff08ccc46
 - name: github.com/spf13/pflag
-  version: a9a634f3de0a7529baded7ad6b0c7467d5c6eca7
+  version: 5644820622454e71517561946e3d94b9f9db6842
```
ndegory, line 263: this change breaks the CLI build. a9a634f is newer than 5644820, and has the fields that are missing in the older version (vendor/github.com/spf13/cobra/bash_completions.go:370: flag.NoOptDefVal undefined (type *pflag.Flag has no field or method NoOptDefVal))

ndegory: will be fixed in next PR

ndegory (approved):

PR #737, opened 2017-02-04
Vendoring for Stack 1.13

ndegory:
Updated vendors for next PR (#734)

test with:

hack/amptools make -f Makefile.refactor.make clean install-deps


subfuzion:
Thanks @ndegory. As you know, I began working yesterday to resolve our vendor conflicts / build issues. It took a while and a lot of testing, but it should be good now. I submitted #738 for the updates, and also went ahead and also added the 3 vendor dependencies @freignat91 needs for #734, so we don't need to test the update more than once.

PR #738, opened 2017-02-05
Resolve vendor conflicts

### LINKED ISSUE #739: Vendor conflicts interfere with glide update and cause build errors
There are a number of issues when running `glide update` and building. One of the recurring issues is the version conflict caused by `docker/docker` pinning to an older version of pflag.

subfuzion:
Fixes #739 

This PR needs to be merged before #734.

This PR resolves various vendor conflicts and fixes the build. The fix involved a combination of reordering packages in `glide.yaml` and pinning a number of packages (eg, `spf13`) to specific latest SHAs (since they don't use semver).

It ensures `glide-cleanall` also removes the generated `glide.lock` file, and improves the `clean-protoc` rule performance by restricting directories.

    $ hack/amptools make -f Makefile.refactor.make cleanall build

This PR also adds the 3 vendor dependencies required by #734 (github.com/xeipuuv/gojsonschema, github.com/xeipuuv/gojsonreference, github.com/xeipuuv/gojsonpointer).


subfuzion:
Thanks @ndegory. As you know, I began working yesterday to resolve our vendor conflicts / build issues. It took a while and a lot of testing, but it should be good now. I submitted #738 for the updates, and also went ahead and the 3 vendor dependencies required by #734, so we don't need to test the update more than once.

Makefile.refactor.make
```diff
@@ -48,7 +48,7 @@ clean-glide:
 
 .PHONY: cleanall-glide
 cleanall-glide: clean-glide
-	@rm -rf .glide
+	@rm -rf .glide glide.lock
```
ndegory, line 51: the issue with the cleanup of the glide.lock, is that you always end up with a modified workspace: even if the pinned commits are the same, the lock file is updated with a new timestamp.

ndegory (approved):
LGTM, the dependency install works from the first time

Makefile.refactor.make
subfuzion: @ndegory Yes, I'm aware. When I was testing, I wanted a way to remove glide.lock as part of make without having to rm manually; since it is an autogenerated file, it goes with the philosophy of cleanall to remove all of our generated artifacts (even though regenerating `glide.lock` isn't quite the same thing). After a cleanall, I intentionally wanted to force `glide.lock` to be updated (as if I had run `update-deps`). Without `glide.lock` being present, this forces glide to effectively do an update, which will recreate `glide.lock` pinned to the latest versions that satisfy the requirements of `glide.yaml`. I'm aware of your point, however, since we potentially end up with an updated glide.lock because *only* the internal glide.lock timestamp (not file timestamp) has changed, and nothing else. The alternative is to not delete `glide.lock` and just run `update-deps` when we wanted it updated, which is probably the behavior we will want to revert back to eventually anyway. There are only two reasons we have these rules in the makefile: 1) insulate make from the actual vendor tool we're using; 2) deal with the temporary hack around removing `vendor/github.com/docker/docker/vendor/golang.org/x/net/trace`. The reality is that the developer has to know to use `glide get` to add a new package which essentially causes a `glide update`, which doesn't fire the fix for reason #2 in any case -- so it's still necessary to run `update-deps` again.... Man, vendoring is a PITA....

Issue #91, opened 2016-08-28
Update Makefile


subfuzion:
Finish dockerizing Makefiles so no dependencies need to be installed on the client just to build. 

Remaining dependencies:
- [x] Go
- [x] glide
- [x] golint
- [x] make

subfuzion:
With the #108, the majority of developer tasks can now be accomplished with a dockerized toolchain. A developer needs to have Docker installed, and currently also needs to have `make` installed. Currently the server (`amplifier`) and the CLI (`amp`) can now be built without installing Go. Go is still needed, however, to run tests and for a few other tools, like `gofmt`.

subfuzion:
Done.

PR #741, opened 2017-02-06
ampmake

subfuzion:
Relates to #91

This is an initial version of `ampmake`, a binary that does the job of `hack/amptools` without requiring make on the host system. It runs a containerized version of make to run the rest of the containerized toolchain.

This is not a complete replacement for `hack/amptools` yet, and it won't work yet for Docker for Windows. This is just a foundation for further iteration and improvements.

For now the intention is to just copy the appropriate binary under the bin directory to a place in your path. The binaries have intentionally been included for now as part of this PR, although eventually they will probably be a part of an amp tools release.

    bin/
        darwin/amd64/ampmake
        linux/amd64/ampmake

Of course, if you already have go, you can just install `ampmake` to your GOPATH (`go install github.com/appcelerator/amp/cmd/ampmake`).

This version of `ampmake` uses the makefile that is currently under development (`Makefile.refactor.make`, not `Makefile`).

## Benefit

The main benefit is that the only prerequisite for a developer to contribute to amp is Docker itself. The binary wraps the usage of make in a container. To rebuild the `ampmake` itself requires the ability to execute a bash shell script, but this can also be executed in a container.

## Usage

    $ bin/<os>/<arch>/ampmake <rule>

    # example:
    $ bin/darwin/amd64/ampmake build

    # or copy the binary to your path
    $ ampmake build

## Rebuilding

    $ cmd/ampmake/build    # updates targets under bin

## Verification

    $ ampmake clean build


JosephGJ (approved):
LGTM. Tested on Mac.

ndegory (approved):
LGTM

PR #747, opened 2017-02-09
Moving config folder under pkg folder

bquenin:
Little refactoring to avoid having code spreading everywhere


JosephGJ (approved):
LGTM

subfuzion:
Hey @bquenin Thanks, this was actually bothering me and I was planning to refactor as well. I'm not crazy about importing `**/config` just to reference `amp.*` platform configuration constants and it's not clear to me why the NATS Streaming client is under the same `pkg` package. I'm not sure where the best place is for this, but let's chew on it just a bit more.

subfuzion (changes requested):
Hey @bquenin Thanks, this was actually bothering me and I was planning to refactor as well. I'm not crazy about importing `**/config` just to reference `amp.*` platform configuration constants and it's not clear to me why the NATS Streaming client is under the same `pkg` package. I'm not sure where the best place is for this, but let's chew on it just a bit more.

bquenin:
Hi @subfuzion,
The idea is to mimic most of golang projects such as docker (https://github.com/docker/docker/tree/master/pkg). It seems that most of them put their "helper" source code under `pkg`.
The main idea being avoid having a lot of root folder entries containing code.

subfuzion:
@bquenin Sorry, I guess I phrased it wrong (inverted) since I understand the idea of utilities being under `pkg`; what I meant is I don't see how this belongs in the same package as things that are general utility code like the NATS client. This is pretty specific to the amplifier platform runtime environment.

bquenin:
@subfuzion sure, it can be moved elsewhere, the idea was to move it away from the root folder

subfuzion:
I thought about suggesting either `api/runtime` or `cmd/amplifer`, although to be honest, those don't feel quite right to me either. That's why I suggested we chew on it a bit more...

bquenin:
Sure, no rush :)

subfuzion:
@bquenin Let's go ahead and do as you propose to reduce top level clutter, with the proviso that we look for a better home later.

subfuzion (approved):
@bquenin Let's go ahead and do as you propose to reduce top level clutter, with the proviso that we look for a better home later.

PR #750, opened 2017-02-13
Latest vendor updates

subfuzion:
The latest glide install fails. This updates `glide.lock` and vendor dependencies to the latest and successfully installs.

## Verification

    $ ampmake cleanall build


neha-viswanathan (approved):
Tested on Mac.

PR #767, opened 2017-02-21
Fixes oauth service and vendor build

### LINKED ISSUE #766: Broken build
Looks like the GitHub client has been updated to require context arguments at the indicated locations.

```
ampmake cleanall build

# github.com/appcelerator/amp/api/rpc/oauth
api/rpc/oauth/oauth.go:104: not enough arguments in call to client.Authorizations.Delete
api/rpc/oauth/oauth.go:140: not enough arguments in call to client.Authorizations.GetOrCreateForApp
api/rpc/oauth/oauth.go:157: not enough arguments in call to client.Users.Get
api/rpc/oauth/oauth.go:161: not enough arguments in call to client.Organizations.IsMember
Makefile.refactor.make:136: recipe for target 'cmd/amplifier/amplifier.alpine' failed
make: *** [cmd/amplifier/amplifier.alpine] Error 2
```

subfuzion:
Fixes https://github.com/appcelerator/amp/issues/766

This fix also repairs `glide.lock` and vendor using the `ampmake` toolchain.

Verification

    $ ampmake cleanall build


ndegory (approved):
LGTM, tested with ```make -f Makefile.refactor.make cleanall build```.
I had a newer version for github.com/opencontainers/runc, not an issue.

subfuzion:
@ndegory Thanks, although I prefer we test with `ampmake` as requested since it's supposed to be a part of our toolchain. Even though all it does is wrap make, if any issues arise using `ampmake`, it's good to discover that during verification as well.

subfuzion:
@ndegory I should also point out -- more importantly -- that if you don't build using `ampmake`, then the binaries you generate won't be the correct alpine binaries required for the images that get built. For performance reasons right now, the makefile doesn't cross-compile the targets. `ampmake` runs the go compiler in an `amptools` alpine container, which does generate alpine binaries. It doesn't particularly matter in this case if all you want to do is ensure that the build succeeds, but it's worth mentioning...

PR #777, opened 2017-03-16
Account email sms

freignat91:
### What I did

Adding email and sms senders

### How I did it

The email sender uses SendGrid
The sms sender uses Twilio

Update ampmake to mount ~/.config in /root/.config
Update amptools to mount ~/.config in /root/.config 
Update makefile to add amplifier.yaml, before amplifier build and remove it after
Move amplifier.yaml in /root/.config/amp in the amplifier image
This file contains private information as SendGrid apiKey or twilio accountId
(this file should be replaced by docker secret on next version)

Move amplifier config from cmd/amplifier/server/config.go to pkg/config/config.go, because of import infinite loop in rpc services when they use mail module (which uses config, which used rpc, ...)

### How to verify it
Not testable for now without account features

### Description for the changelog

- add email sender using SendGrid
- add sms sender using Twilio


bquenin (approved):
LGTM

PR #785, opened 2017-03-17
Local Swarm cluster deployment with Infrakit

ndegory:
Swarm cluster deployment for AMP based on InfraKit.

## Bootstrap

The AMP CLI will bootstrap a Swarm cluster composed of DockerInDocker nodes.
This PR is limited right now to:
- amp start, creates a 5 nodes cluster (m1, m2, m3 and 2 workers with random names), with an infrakit container managing it (self healing) helped by a docker instance plugin. An extra container is created to map a list of ports from the host to the swarm. It creates also a registry service for making our images available to the Swarm,
- amp stop, destroys the cluster and the infrakit containers

Part of the infrakit deployment is the creation of an overlay network (ampnet), that is needed for the deployment of the amplifier stack.

## Verification

Add the appropriate `bin` subdirectory to your path. For example:

    $ export PATH=$PWD/bin/darwin/amd64:$PATH

Build server and CLI images:

    $ ampmake build

Create a local swarm-enabled cluster (`amp start`):

    $ amp start

Start amplifier:

To start amplifier, you will need to push the `amplifier` image to the amp swarm registry. To do this, you need to tag it first, then push it:

Then deploy the amplifier stack:

    $ docker run -it --rm --network=hostnet -v $PWD/stacks:/stacks docker --host=m1 stack deploy -c /stacks/amplifier.stack.yml amplifier

Verify services have started (will take some time as images are pulled):

    $ alias mgr='docker run -it --rm --network=hostnet docker --host=m1'
    $ mgr service ls

Start pinger

    $ mgr service create -p=3000:3000 --name=pinger appcelerator/pinger

Scale pinger

    $ mgr service scale pinger=2

Stop pinger

    $ mgr service rm pinger


neha-viswanathan:
I get this error message on the amplifier service -`No such image: appcelerator/amplifier:local`

ndegory:
Thank you @neha-viswanathan, you're right the nodes were not using the local registry, it is now fixed

subfuzion:
@ndegory Do you want to update the PR description regard port publishing?

subfuzion (commented):
* Right now if we don't do a graceful shutdown (`amp stop`), the `amp-bootstrap` volume won't be removed and this prevents a subsequent `amp start` from succeeding. We will want to automatically remove the volume as part of start, or if there is a reason why this shouldn't be the default behavior, then at least provide an option (eg `amp start --force` or `amp start --clean`) to remove the volume, and display this information to the user when the basic `amp start` fails.

* We discussed that the size of the cluster should *include* the proxy worker for local port publishing. However, on second thought, I like that you are not counting it since this is an incidental aspect of supporting local development.

* We need to consider how we leverage infrakit itself for creating of the registry node so that it too benefits from infrakit self-healing. Right now this is basically what my former bootstrap script did, which doesn't support availability.

subfuzion (approved):
@ndegory Nice work. We have added a great deal of complexity compared to the previous bootstrap version, but on the other hand we get the benefit of demonstrating the same approach for cluster management locally as remotely, along with the self-healing aspects that made InfraKit appealing in the first place. This was a great effort between you and @JosephGJ!  👍

PR #813, opened 2017-03-25
Update tools

subfuzion:
Updated gotools grpc to 1.2.0; updated dependent amptools, ampmake; rebuilt proto files.


ndegory:
@subfuzion do you mind if I merge the build scripts you added in gotools and amptools in the images/build-images.sh? (the push action is missing in the latter)

PR #821, opened 2017-03-26
Update amptools to docker 17.03

subfuzion:
Another update for amptools, this one to bump the Docker version.


subfuzion:
CI build fails due to lint checks, as expected right now.

PR #823, opened 2017-03-27
CLI refactoring

subfuzion:
This is a big cleanup for the CLI codebase, which had gotten a bit unwieldy over time.

    $ PATH=bin/<GOOS>/<GOARCH>:$PATH
    $ ampmake build-cli
    $ amp --help


cmd/amp/main.go
```diff
@@ -27,279 +15,39 @@ var (
...
-	}
-	if cmd.Flag("password").Changed {
-		password = cmd.Flag("password").Value.String()
-	} else {
-		password = getPassword()
+	// Initial configuration (before processing environment and flags)
+	config = &cli.Configuration{
+		Version: Version,
+		Build:   Build,
+		Address: "m1:50101",
```
bquenin, line 27: I think we should have a way to override the server address.

cmd/amp/commands.go
```diff
@@ -0,0 +1,48 @@
+package main
+
+import (
+	"github.com/appcelerator/amp/cli"
+	"github.com/appcelerator/amp/cli/command/version"
+	"github.com/spf13/cobra"
+)
+
+type Opts struct {
```
bquenin: `cmd/amp/commands.go:9:6:warning: exported type Opts should have comment or be unexported (golint)
`

Makefile
```diff
@@ -102,10 +102,12 @@ AMPBOOTIMG := appcelerator/$(AMP)-bootstrap:$(AMPTAG)
 AMPTARGET := $(CMDDIR)/$(AMP)/$(AMPBINARY)
 # TODO: add api/client back to project
 #AMPDIRS := api/client $(CMDDIR)/$(AMP) tests
-AMPDIRS := $(CMDDIR)/$(AMP) tests
+AMPDIRS := $(CMDDIR)/$(AMP) cli tests
 AMPSRC := $(shell find $(AMPDIRS) -type f -name '*.go')
 
 $(AMPTARGET): $(CMDDIR)/$(AMP)/Dockerfile $(GLIDETARGETS) $(PROTOTARGETS) $(AMPSRC) $(AMPBOOTDIR)/$(AMPBOOTEXE)
+	@echo "Compiling $(AMP) source(s):"
+	@echo $?
```
bquenin, line 110: Probably development helper that should be removed?

Makefile
```diff
@@ -142,6 +144,8 @@ AMPLDIRS := cmd/$(AMPL) api data tests
 AMPLSRC := $(shell find $(AMPLDIRS) -type f -name '*.go')
 
 $(AMPLTARGET): $(GLIDETARGETS) $(PROTOTARGETS) $(AMPLSRC)
+	@echo "Compiling $(AMPL) source(s):"
+	@echo $?
```
bquenin, line 148: Probably development helper that should be removed?

cli/logger.go
```diff
@@ -0,0 +1,80 @@
...
+	out     *OutStream
+	verbose bool
+}
+
+func init() {
+	// Creates an instance of Logger for grpc logging
+	// WARNING: the grpc logger can only be set during init()
+	// https://godoc.org/google.golang.org/grpc/grpclog#SetLogger
+	// TODO: set verbose to false after testing
+	grpclog.SetLogger(Logger{out: NewOutStream(os.Stdout), verbose: true})
```
bquenin, line 25:
```
// TODO: set verbose to false after testing
grpclog.SetLogger(Logger{out: NewOutStream(os.Stdout), verbose: true})
```

cli/required.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func NoArgs(cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		return nil
+	}
+	return fmt.Errorf("unexpected argument: %s\nSee '%s --help'", args[0], cmd.CommandPath())
+}
+
+//func NoArgs() func(*cobra.Command, []string) error {
```
bquenin: Should it be removed ?

cmd/amp/main.go
freignat91, line 27: Perhaps better to separate address and port in configuration. it could help later, if we have several ports used on the same address

freignat91 (approved):
LGTM

ndegory (approved):
LGTM

cmd/amp/main.go
subfuzion: Yes, agree. It will be restored as an option and environment variable, just like before. This just provides an initial value for the config default.

cmd/amp/commands.go
subfuzion: Thanks. I just started doing the lint checks and can also see the CI build results. I still need to clean up the PR, squash, and won't expect this merged until the CI build passes.

Makefile
subfuzion: I actually just recently added it because it *really* helps to confirm what's getting built while hacking on it. The long silent pause wasn't helping either, and unless we want to suppress all the docker output as well, it's not like these are actually silent commands right now. I'd like to keep these here for the time being. We can revisit it again in a couple of weeks.

Makefile
subfuzion: See my response above.

cli/logger.go
subfuzion: We won't be finished testing for another two weeks. ;)

cli/required.go
subfuzion: Yes, but I'm not done with this file yet. It's the last remaining wip. Among other things, I need to add more argument validators. I'll clean this up as part of my linting and squashing.

cmd/amp/main.go
subfuzion: Sure. I had host and port before. But maybe we can hold off on updating this until we actually need to. I'm thinking right now that this is just for the default well known address for the server.

JosephGJ (approved):
LGTM

Makefile
bquenin: Sure

Makefile
bquenin: ok

cli/required.go
bquenin: ok

cmd/amp/commands.go
bquenin: ok

cmd/amp/main.go
bquenin: ok

cli/logger.go
bquenin: ok, I wasn't sure if it was intentional or not

neha-viswanathan (approved):

bquenin (approved):
Awesome job @subfuzion :+1: !

PR #831, opened 2017-03-28
Mount user directory in ampwrapper to get access to the token

### LINKED ISSUE #830: amp login issue
After successful amp signup/verified, the amp login works but does keep the session for the next command

amp user verify --token eyJhbG...
Your account has now been activated.
$ amp login
username: freignat2
password: qweqweqwe
Welcome back, freignat2!
$ amp stats
credentials required

bquenin:
Fixes #830


freignat91 (approved):
tested ok

cmd/ampwrapper/main.go
```diff
@@ -14,9 +16,15 @@ var (
...
 func init() {
+	homedir, err := homedir.Dir()
+	if err != nil {
+		panic(err)
+	}
+
 	dockerArgs = []string{
 		"run", "-i", "--rm", "--name", "ampcli",
 		"-v", "/var/run/docker.sock:/var/run/docker.sock",
+		"-v", fmt.Sprintf("%s/.config/amp:/root/.config/amp:rw", homedir),
```
subfuzion: Since our config command allows you to save settings, you should remove `:rw` from the mount ... but you might want to confirm if that works correctly on Linux right now since I think we still have an outstanding issue with setting the user correctly.

subfuzion (changes requested):

subfuzion (approved):

PR #834, opened 2017-03-28
Use dupl linter

neha-viswanathan:
Use of [dupl](https://github.com/mibk/dupl#dupl-) for detecting clones in backend code.
Since some of the functions have similar implementation, `// nolint : dupl` has been added to suppress the warnings.

To test -
`ampmake lint`

### DIFF
--- a/api/rpc/account/account.go
+++ b/api/rpc/account/account.go
@@ -201,6 +201,7 @@ func (s *Server) ForgotLogin(ctx context.Context, in *ForgotLoginRequest) (*empt
 }
 
 // GetUser implements account.GetUser
+// nolint : dupl
 func (s *Server) GetUser(ctx context.Context, in *GetUserRequest) (*GetUserReply, error) {
 	// Get the user
 	user, err := s.Accounts.GetUser(ctx, in.Name)
@@ -215,6 +216,7 @@ func (s *Server) GetUser(ctx context.Context, in *GetUserRequest) (*GetUserReply
 }
 
 // ListUsers implements account.ListUsers
+// nolint : dupl
 func (s *Server) ListUsers(ctx context.Context, in *ListUsersRequest) (*ListUsersReply, error) {
 	users, err := s.Accounts.ListUsers(ctx)
 	if err != nil {
@@ -297,6 +299,7 @@ func (s *Server) CreateOrganization(ctx context.Context, in *CreateOrganizationR
 }
 
 // AddUserToOrganization implements account.AddOrganizationMember
+// nolint : dupl
 func (s *Server) AddUserToOrganization(ctx context.Context, in *AddUserToOrganizationRequest) (*empty.Empty, error) {
 	if err := s.Accounts.AddUserToOrganization(ctx, in.OrganizationName, in.UserName); err != nil {
 		return &empty.Empty{}, convertError(err)
@@ -312,6 +315,7 @@ func (s *Server) AddUserToOrganization(ctx context.Context, in *AddUserToOrganiz
 }
 
 // RemoveUserFromOrganization implements account.RemoveOrganizationMember
+// nolint : dupl
 func (s *Server) RemoveUserFromOrganization(ctx context.Context, in *RemoveUserFromOrganizationRequest) (*empty.Empty, error) {
 	if err := s.Accounts.RemoveUserFromOrganization(ctx, in.OrganizationName, in.UserName); err != nil {
 		return &empty.Empty{}, convertError(err)
@@ -336,6 +340,7 @@ func (s *Server) ChangeOrganizationMemberRole(ctx context.Context, in *ChangeOrg
 }
 
 // GetOrganization implements account.GetOrganization
+// nolint : dupl
 func (s *Server) GetOrganization(ctx context.Context, in *GetOrganizationRequest) (*GetOrganizationReply, error) {
 	organization, err := s.Accounts.GetOrganization(ctx, in.Name)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


data/accounts/store.go
```diff
@@ -46,6 +46,7 @@ func secureUser(user *User) *User {
 	return user
 }
 
+// nolint : dupl
 func (s *Store) getUser(ctx context.Context, name string) (user *User, err error) {
 	if user, err = s.rawUser(ctx, name); err != nil {
```
subfuzion, line 51: Since we're doing code cleanup, there's nothing gained here by using named return values. In fact, most of the usage in these files are not that consistent (a lot of `nil`s getting returned anyway). I recommend getting rid of  them.

data/storage/etcd/store.go
```diff
@@ -254,12 +254,14 @@ func (s *etcd) CompareAndSet(ctx context.Context, key string, expect proto.Messa
...
 // Watch implements storage.Interface.Watch.
+// nolint : dupl
 func (s *etcd) Watch(ctx context.Context, key string, resourceVersion int64, filter storage.Filter) (storage.WatchInterface, error) {
 	key = s.prefix(key)
 	return s.watch(ctx, key, resourceVersion, filter, false)
 }
 
 // WatchList implements storage.Interface.WatchList.
+// nolint : dupl
 func (s *etcd) WatchList(ctx context.Context, key string, resourceVersion int64, filter storage.Filter) (storage.WatchInterface, error) {
```
subfuzion, line 265: `Watch` and `WatchList` like bona fide duplicates to me except for the boolean flag on `s.watch()`. Seems like the duplicate code should be factored.

api/rpc/account/account.go
```diff
@@ -215,6 +216,7 @@ func (s *Server) GetUser(ctx context.Context, in *GetUserRequest) (*GetUserReply
 }
 
 // ListUsers implements account.ListUsers
+// nolint : dupl
```
subfuzion, line 219: I'm curious if we should try adjusting the threshold size (minimum token sequence) to eliminate so many of these false positives (see [DuplThreshold](https://github.com/alecthomas/gometalinter/blob/master/config.go#L184)).

subfuzion (changes requested):
See inline comments.

api/rpc/account/account.go
neha-viswanathan: I tried setting `--dup-threshold` for `dupl` in the Makefile to a different value. It seems to work.
`--enable=dupl --dupl-threshold=100`

data/accounts/store.go
bquenin: The named return values are used to avoid declaring variables in the function body.

api/rpc/account/account.go
bquenin: I think the default value is ok. If we set the threshold too high, we're probably going to miss actual duplication. We have a lot of duplicates in test code and data storage which makes sense considering the nature of the code.

data/storage/etcd/store.go
bquenin: The actual duplicated code has been factored in `s.watch()`.

subfuzion: I see. Yes, agree ... not worth adding a potentially ambiguous boolean parameter to `Watch` just to eliminate `WatchList`.

api/rpc/account/account.go
subfuzion: Yes, but I would argue that's the reason for tuning. If you know the nature of the code base is going to trigger a lot of false positives, you tweak the threshold a bit. I do think dupl is overly aggressive.

data/accounts/store.go
subfuzion: I know what named return values are for.  ;)  Converting `=` to `:=` handles declaration. My point is about consistency, as well as using named return values in a way that doesn't really increase comprehension. Not to bikeshed the issue, but we should be more consistent.

subfuzion: Lines 50 - 57

```
func (s *Store) getUser(ctx context.Context, name string) (user *User, err error) {
	if user, err = s.rawUser(ctx, name); err != nil {
		return nil, err
	}
	if user == nil {
		return nil, UserNotFound
	}
	return user, nil
}
```

Converting `=` to `:=`

```
func (s *Store) getUser(ctx context.Context, name string) (*User, error) {
	if user, err := s.rawUser(ctx, name); err != nil {
		return nil, err
	}
	if user == nil {
		return nil, UserNotFound
	}
	return user, nil
}
```
I'm not particularly a huge fan (especially of naked returns), but named return values can be convenient. I just think within the very same files at least there ought to be consistency. The previous similar function on Line 30 (`func (s *Store) rawUser(ctx context.Context, name string) (*User, error) {`)  flagged as a duplicate doesn't use named return values.

bquenin:
```
if user, err := s.rawUser(ctx, name); err != nil {
		return nil, err
	}
```

This doesn't work since the `user` variable is only defined in the `if` scope, which is why, by convenience over consistency, I used a named return value.

If we absolutely need to get rid of names return values, then this is the alternative:
```
func (s *Store) getUser(ctx context.Context, name string) (*User, error) {
	var user *User
	var err Error
	if user, err = s.rawUser(ctx, name); err != nil {
		return nil, err
	}
	if user == nil {
		return nil, UserNotFound
	}
	return user, nil
}
```

api/rpc/account/account.go
bquenin: Up to you, yet I think it's normal to have false positive on the first run.

PR #838, opened 2017-03-28
Remove --no-cache from Makefile

subfuzion:
Restoring previous behavior since this is slowing local builds a bit too much.


bquenin (approved):
LGTM (linux)

ndegory (approved):
LGTM, but we should use a tag for the base image of amplifier (alpine:3.5 instead of alpine) to limit the risk with cached images.

PR #841, opened 2017-03-29
Exclude cmd/amp from lint check for now

neha-viswanathan:
Exclude cmd/amp from lint check for now (until the refactored CLI is ready)


subfuzion (approved):
CI build passes now! 👍

Issue #840, opened 2017-03-29
end to end tests in Travis builds


ndegory:
Update the travis yaml configuration to run a tests with:
- local deployment (docker in docker)
- amplifier tagged and pushed to the local registry
- deploy the amplifier stack to the cluster
- CLI connection to amplifier

Can be inspired by the verification steps of PR #785 (originally #774) and the Travis updates of PR #759

PR #774, opened 2017-03-14
amp refactoring and updates for Docker 1.13

subfuzion:
This PR adds the `amp start` command for bootstrapping a local swarm. It introduces `ampwrapper`, which is a binary for executing a containerized version of the CLI.

## ampmake - containerized make

The amp toolchain is completely containerized and this includes `make`. `ampmake` is a wrapper for executing `make` inside of an alpine container. Even if you already have `make` on your system, you should use `ampmake` since the project makefile leverages a fully containerized toolchain and will ensure binaries are created for the correct alpine-based targets before building any images.

At the present time, the makefile that is used by `ampmake` is `Makefile.refactor.make`.

`ampmake` should be built with the `build` script located under `cmd/ampmake`.

## amp - CLI image

The `amp` executable itself (not `ampwrapper`, explained later) needs to be rebuilt when either the CLI or the `local-bootstrap` script is updated. As part of rebuilding the CLI, the Docker image will also be rebuilt. This will update the local `appcelerator/amp:local` image stored in the local Docker image cache. It is not necessary to update `ampwrapper` when updating the CLI since all the wrapper does is tell Docker to run the a container using the CLI image.

To build it, run:

    $ ampmake build-cli

or

    $ampmake clean-cli build-cli

## ampwrapper - CLI executable

This is a wrapper for the CLI, which runs as a Docker container. The reason for the wrapper is to insulate users from an unwieldy Docker command to run an amp command as well as future updates. For example:

    $ docker run -it --rm  --name ampcli -v /var/run/docker.sock:/var/run/docker.sock --network hostnet appcelerator/amp:local start

The Docker socket is mounted because the `start` command uses Docker to create a number of containers to start a local cluster. It uses the `hostnet` bridge network because the shared network is used to refer to the nodes by hostname while creating the swarm.

By using the wrapper, the command becomes:

    $ amp start

The wrapper source is located under `cmd/ampwrapper`. To build it, run:

    $ ampmake build-cli-wrapper

This will generate executables for Linux and Mac (darwin) and place them under `bin`.

## local-bootstrap

The `local-bootstrap` script is used to start a local cluster of nodes and then initialize Docker swarm mode. One of the nodes will be used to host a registry, the other nodes will provide managers and workers for the swarm.

The nodes are actually just Docker containers; to be able to use them to join a swarm, each container runs its own instance of a Docker engine using a Docker-in-Docker.

The script is parameterized to create a configurable number of managers and workers. By default, it will create three managers and two workers.

[...]

PR #842, opened 2017-03-29
Update travis.yml

neha-viswanathan:
Ref: #774, #785, #840  

Updated Travis Build to include the following -
- Lint check
- Create local swarm (with InfraKit) 
- Start amplifier - tag and push to local registry 
- Ensure the container/nodes/swarm are ready before deploying amplifier stack
- Deploy amplifier stack

### DIFF
--- a/hack/ctrcheck
+++ b/hack/ctrcheck
@@ -1,20 +1,25 @@
 #!/bin/bash
 
 usage() {
-    echo "usage: ctrcheck CMD [FILTER] [TIMEOUT]"
-    echo "example: ctrcheck docker ps --filter "name=jdoe" 60"
+    echo "usage: ctrcheck NAME [FILTER] [TIMEOUT]"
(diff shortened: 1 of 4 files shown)


hack/ctrcheck
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+usage() {
+    echo "usage: ctrcheck docker ps [FILTER] [TIMEOUT]"
+    echo "example: ctrcheck docker ps --filter "name=jdoe" 60"
```
subfuzion: @neha-viswanathan The flexibility of supplying a custom filter could be useful, but when we chatted about this, I was just thinking something simple, more along the lines of `ctrcheck NAME [STATUS] [TIMEOUT]`, where status could default to `running` and timeout could default to `300` (5m) (we can tune the value based on travis performance).

subfuzion (approved):
LGTM

PR #843, opened 2017-03-30
[Monitoring] preparation step

freignat91:
New monitoring, preparation step: 
- remove influxDB from amplifier config
- make amplifier connect to elasticsearch only when needed to do not have stating dependency between amplifier and monitoring stacks
- have a minimal monitoring stack for now: elasticsearch, kibana, nats


freignat91:
to be move under new-monitoring

PR #847, opened 2017-03-30
[Monitoring] Add ampbeat targets to the Makefile

bquenin:
This PR adds `ampbeat` targets in the Makefile by mimicking what already exists for other binaries, namely:
 - `build-beat`: Build the alpine binary and then build the docker image
 - `rebuild-beat`: Does the same, expect it cleans the binary first
 - `clean-beat`: Cleans the binary

Note the `ampbeat` binary is only a skeleton and is not integrated to the monitoring stack yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ clean-protoc:
 .PHONY: clean cleanall
 # clean doesn't remove the vendor directory since installing is time-intensive;
 # you can do this explicitly: `ampmake clean-deps clean`
-clean: clean-protoc clean-cli clean-server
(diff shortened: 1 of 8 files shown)


freignat91 (approved):
LGTM

Makefile
```diff
@@ -151,11 +151,36 @@ build-server: $(AMPLTARGET)
 
 rebuild-server: clean-server build-server
 
-.PHONY: clean-cli
+.PHONY: clean-server
```
subfuzion, line 154: Good catch! 😊

subfuzion (approved):

PR #849, opened 2017-03-30
[Monitoring] Add amp-agent targets to the Makefile

freignat91:
This PR adds amp-agent targets in the Makefile by mimicking what already exists for other binaries, namely:

- build-agent: Build the alpine binary and then build the docker image
- rebuild-agent: Does the same, expect it cleans the binary first
- clean-agent: Cleans the binary

Note the amp-agent code/binary hasn't been updated for now

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,8 @@ clean-protoc:
 .PHONY: clean cleanall
 # clean doesn't remove the vendor directory since installing is time-intensive;
 # you can do this explicitly: `ampmake clean-deps clean`
-clean: clean-protoc clean-cli clean-server clean-beat
+
+clean: clean-protoc clean-cli clean-server clean-beat clean-agent
 cleanall: clean cleanall-deps
 
 # =============================================================================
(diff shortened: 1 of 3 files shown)


bquenin (dismissed):
LGTM

subfuzion (approved):

bquenin (commented):
Just thinking out loud, but should we get rid of the `amp` prefix while we're at it? something like `monitoring-agent` or `mon-agent` or `agent` or something else ?
Or keep it as it is today

subfuzion:
@bquenin I thought about that as well when reviewing the PR earlier, but decided not to mention it to @freignat91 for now. I figure we'll do this anyway to migrate amp -> atomiq and it might as well be part of one coordinated update. I actually was going to ask you if you wanted to tackle this since I noticed your attention to detail is quite good when renaming/refactoring things. Let's merge this PR for him, but let's talk about renaming tomorrow. I definitely don't like all the hyphens anyway!

subfuzion (approved):

PR #855, opened 2017-03-31
[Monitoring] Update proto files 

freignat91:
Update proto files for logs and stats to prepare to move on new amp-agent and amp-beat
Adjust current code to do not break the build


bquenin (approved):
LGTM

cmd/amp-agent/core/logs.go
```diff
@@ -55,7 +55,7 @@ func openLogsStream(ID string, lastTimeID string) (io.ReadCloser, error) {
 // Use elasticsearch REST API directly
 func getLastTimeID(ID string) string {
 	request := strings.Replace(elasticSearchTimeIDQuery, "[container_id]", ID, 1)
-	req, err := http.NewRequest("POST", conf.elasticsearchURL+"/amp-logs/_search", bytes.NewBuffer([]byte(request)))
+	req, _ := http.NewRequest("POST", conf.elasticsearchURL+"/amp-logs/_search", bytes.NewBuffer([]byte(request)))
```
subfuzion: Why ignore error?

subfuzion (changes requested):
LGTM, just one question about ignoring errors.

cmd/amp-agent/core/logs.go
freignat91: This is just to do not have compile warning, this code is going to be updated and then we'll handle error.

freignat91: It's to do not have build warning, because in the original code there was a bad use of the err var. All the amp-agent code is going to be updated deeply, this part will disappear next pr because amp-agent no longer be connected to elasticsearch

freignat91: actually handle the error to be clean

subfuzion:
The Travis CI build passed, although there seems to be an issue with the status being reported for the CI check that will need to be investigated. However, this shouldn't hold up this PR, so disregarding CI check for now and performing an administrator merge.

PR #868, opened 2017-04-03
Refactoring amplifier configuration

bquenin:
## How to test
    $ ampmake build
    $ amp cluster create
    $ alias mgr='docker run -it --rm --network=hostnet -v $PWD/stacks:/stacks docker --host=m1'
    $ mgr stack deploy -c /stacks/amplifier.stack.yml amplifier

Make sure amplifier is up and running


cmd/amplifier/server/configuration.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// ReadConfig reads the configuration file
+func ReadConfig(config *Configuration) error {
+	// Add matching environment variables - will take precedence over config files.
+	viper.AutomaticEnv()
+
+	// Add default config file search paths in order of decreasing precedence.
+	viper.SetConfigName("amplifier")
+	viper.AddConfigPath("/usr/local/bin/")
```
subfuzion: Configuration files shouldn't be stored in bin; I recommend: `/etc/atomiq/`.

pkg/config/amp.go
```diff
@@ -2,49 +2,26 @@ package amp
 
 import "time"
 
+// AMP platform default values
 const (
-	// AmplifierDefaultEndpoint is the default amplifier endpoint
-	AmplifierDefaultEndpoint = "amplifier:50101"
+	AmplifierDefaultPort          = ":50101"
+	AmplifierDefaultPublicAddress = "local.appcelerator.io"
+	EmailDefaultSender            = "amp.appcelerator@axway.com"
```
subfuzion: Should not have a hardcoded email sender (if we need it, the domain should default to `atomiq.io`)

subfuzion (changes requested):
Looks good, just a couple of comments right now.

pkg/config/amp.go
bquenin: It's the email sender used for outgoing emails if none is provided, I'm going to change it to `amp@atomiq.io`

cmd/amplifier/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM alpine
 RUN apk --no-cache add ca-certificates
 COPY amplifier.alpine /usr/local/bin/amplifier
-COPY amplifier.yaml /root/.config/amp/amplifier.yaml
+COPY amplifier.yaml /etc/default/amplifier.yaml
```
subfuzion: See other comment about moving config under `/etc`.

cmd/amplifier/server/configuration.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// ReadConfig reads the configuration file
+func ReadConfig(config *Configuration) error {
+	// Add matching environment variables - will take precedence over config files.
+	viper.AutomaticEnv()
+
+	// Add default config file search paths in order of decreasing precedence.
+	viper.SetConfigName("amplifier")
+	viper.AddConfigPath("/etc/default/")
```
subfuzion: @bquenin Sorry, not sure what happened, but I had updated my comment before saving the review, but I had several comments opened and apparently my final recommendation was lost. Please change to `/etc/atomiq`. Thanks!

bquenin: No worries!

cmd/amplifier/server/configuration.go
```diff
@@ -0,0 +1,46 @@
...
+
+// ReadConfig reads the configuration file
+func ReadConfig(config *Configuration) error {
+	// Add matching environment variables - will take precedence over config files.
+	viper.AutomaticEnv()
+
+	// Add default config file search paths in order of decreasing precedence.
+	viper.SetConfigName("amplifier")
+	viper.AddConfigPath("/etc/atomiq/")
+	if err := viper.ReadInConfig(); err != nil {
```
subfuzion, line 36: One concern with this -- and the reason I abandoned this approach -- is that settings can be specified in the config file that actually shouldn't be loaded (such as version and build compile time constants) and environment variables shouldn't be used at all (nor should they override) credential settings. Credentials themselves shouldn't be in the global config but stored under home directory in a directory with permissions set to 700 and file(s) set to 600. That doesn't have to be done with this PR as long as we do get it done. I'll create an issue for it.

subfuzion (approved):
LGTM overall with the note that we will still have to improve configuration support.

PR #871, opened 2017-04-03
add stats api

freignat91:
Add stats api
Tests pending the update to the CLI and test scripts.


api/rpc/stats/stats.go
```diff
@@ -74,12 +76,252 @@ func (s *Stats) doesElasticsearchServiceExist(ctx context.Context) bool {
...
+	ranges, ok := result.Aggregations.Terms("group")
+	if !ok {
+		return nil, fmt.Errorf("Request error 'group' not found")
+	}
+	ret := &StatsReply{}
+	for _, buck := range ranges.Buckets {
+		var found bool
+		histo, found := buck.Histogram("histo")
+		if !found {
+			panic("sub agg histo not found")
```
subfuzion: Is panic something we want to be doing here? Not a microservice yet, so this currently will stop the whole `amplifier` process. I'm not quite clear, but is this even something that benefits from restarting the process?

subfuzion (changes requested):

api/rpc/stats/stats.go
freignat91: sure, it's a mistake. fixed it

subfuzion (approved):
LGTM 👍

PR #873, opened 2017-04-03
Refactor service startup

bquenin:


freignat91 (approved):
LGTM

subfuzion (approved):
PR looks great, @bquenin. Only real comment is that I had already mentioned to you that I had created a docker wrapper over the weekend for the swarm support I needed. Slightly different approach, but close enough, however, and I don't want to block you, so I'll merge your PR now and handle the refactoring with my PR when ready. Also, among the other concerns I mentioned to you yesterday about the config package, I forgot to mention the concerns over how we have amplifier-specific stuff in  `pkg/`, such as amplifier settings in `pkg/config`. It's really not a part of this PR except it's getting used and I'm getting worried about things getting more baked in (just after we went through a big refactoring effort). I think we're abusing `pkg`, at least as far as the way the rest of the community uses it as a "go-gettable" set of reusable library packages. It might be good for us to tackle this sooner than later, I'm thinking...

PR #869, opened 2017-04-03
Cluster ID

ndegory:
## content of the PR
- Cluster creation returns a cluster id to stdout and sets exit status code to `0` on success or `1` on error.
- The registry node as well as the swarm nodes are now all labeled with `atomiq.clusterid=CLUSTERID` to identify all the cluster nodes
- The registry is now also destroyed when the cluster is destroyed (but the registry volume is still preserved)
- `-i CLUSTERID` sets a specific id to use for the cluster on creation
- `-s CLUSTERID` provides a exit status code: `0` if cluster has converged, `1` if not
- `-l CLUSTERID` provides a list of the nodes in the cluster
- `-d CLUSTERID` destroys the cluster

## verification

```
$ ./bootstrap/bootstrap 2>/dev/null
414fbb29-057a-4827-894b-e9bb2c14ff93
$ docker container ls --format 'table {{.Names}}\t{{.Status}}\t{{.Label "atomiq.clusterid"}}\t{{.Label "infrakit.group"}}'
NAMES                    STATUS              CLUSTERID                              GROUP
m2                       Up 26 minutes       414fbb29-057a-4827-894b-e9bb2c14ff93   amp-manager
laughing_brahmagupta     Up 26 minutes       414fbb29-057a-4827-894b-e9bb2c14ff93   amp-worker
m3                       Up 27 minutes       414fbb29-057a-4827-894b-e9bb2c14ff93   amp-manager
amp-proxy                Up 27 minutes       414fbb29-057a-4827-894b-e9bb2c14ff93   amp-proxy
m1                       Up 27 minutes       414fbb29-057a-4827-894b-e9bb2c14ff93   amp-manager
instance-plugin-docker   Up 27 minutes
infrakit                 Up 27 minutes
registry                 Up 27 minutes       414fbb29-057a-4827-894b-e9bb2c14ff93
$ ./bootstrap/bootstrap -l  414fbb29-057a-4827-894b-e9bb2c14ff93
414fbb29-057a-4827-894b-e9bb2c14ff93 amp-manager m2 55930ea7c968
414fbb29-057a-4827-894b-e9bb2c14ff93 amp-manager m3 22eb1c063b7c
414fbb29-057a-4827-894b-e9bb2c14ff93 amp-manager m1 7ffc31eec36e
414fbb29-057a-4827-894b-e9bb2c14ff93 amp-worker laughing_brahmagupta 5bc6f9dc8c88
[...]

PR #874, opened 2017-04-03
Add local cluster creation blocking option and fix status

ndegory:
See #869 for other details.

- `boostrap -b TIMEOUT` makes the command block for a maximum of seconds (`TIMEOUT`) until the cluster is ready. If the command times out, the exit code is non zero.
- `bootstrap -s CLUSTERID` is now more reliable


subfuzion (approved):
LGTM

PR #875, opened 2017-04-04
Update version command for amplifier

neha-viswanathan:
## To test

    $ ampmake build
    $ amp cluster create
    $ docker run -it --rm --network=hostnet -v $PWD/stacks:/stacks docker --host=m1 stack deploy -c /stacks/amplifier.stack.yml amplifier
    $ alias mgr='docker run -it --rm --network=hostnet docker --host=m1'
    $ mgr service ls
Make sure amplifier is up and running

```
    $ amp version
    amp:
     Version:       v0.6.0-dev
     Build:         f8f8ed86
     Address:       m1:50101
     Go version:    go1.8
     OS/Arch:       linux/amd64

    amplifier:      
     Version:       v0.6.0-dev
     Build:         e42f33ff
     Go version:    go1.8
     OS/Arch:       linux/amd64
```

### DIFF
--- a/cli/command/version/version.go
+++ b/cli/command/version/version.go
@@ -7,11 +7,14 @@ import (
 	"github.com/appcelerator/amp/cli"
 	"github.com/docker/docker/pkg/templates"
 	"github.com/spf13/cobra"
+	"github.com/appcelerator/amp/api/rpc/version"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
 )
 
 type Version struct {
 	Client *ClientVersionInfo
-	Server interface{} // should be:  *ServerVersionInfo from rpc/version/version.proto ...
+	Server *version.Info
 }
 
 // AmplifierOK Checks if AMP is connected to Amplifier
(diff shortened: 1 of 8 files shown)


cli/command/version/version.go
```diff
@@ -38,7 +40,7 @@ var versionTemplate = `amp:
 amplifier:      {{if .IsConnected}}
  Version:       {{.Server.Version}}
  Build:         {{.Server.Build}}
- Address:       {{.Server.Address}}
+ Port:          {{.Server.Port}}
```
subfuzion: Not sure `Port` is needed. If we can connect to the service from the cli, then we already know the port. It could possibly be interesting to know if the port is different because of port mapping, but that probably isn't really needed (and potentially confusing). Probably better to remove it and add later if we realize we do need it.

subfuzion (changes requested):
Please resolve merge conflicts.

bquenin (approved):
LGTM (tested on linux)

subfuzion (approved):
LGTM

PR #880, opened 2017-04-04
User signup

bquenin:
Don't forget that you need a valid SendGrid token in your amplifier configuration file located here: `~/.config/amp/amplifier.yaml`

## How to test
    $ ampmake build
    $ hack/dev
    $ amp user signup


subfuzion (approved):
@bquenin It's failing for me because amplifier isn't connecting to etcd. However, the PR itself looks good and since the failure is related to a known issue that we're working on, I'm going to merge this and the related account PRs. We'll do more thorough testing after the etcd issue is resolved and open subsequent issues if problems with the account commands arise then.

PR #885, opened 2017-04-05
[New monitoring] update stats-api

freignat91:
Final version so far of the stats api.
Can't be tested for now without the stats cli (next pr)
Manager current and historic requests with all metrics


bquenin (approved):
LGTM

PR #892, opened 2017-04-05
CI build deploy updates

subfuzion:
Modify build order, services, increase service check timeout, dump service logs if failure. It displays service status every 30 seconds while waiting for services to come up.


subfuzion:
Updates related to configuration and CI build logging working as expected, although all services not starting successfully (see [build log](https://travis-ci.org/appcelerator/amp/builds/219088863).

PR #893, opened 2017-04-05
Reuse of shared ikt template code

JosephGJ:
ikt `config.tpl` file code is moved to separate files so that there is better reusability.


ndegory:
@JosephGJ local deployment is failing: "invalid character '}' looking for beginning of object key string"
Please update the templates.

JosephGJ:
Fixed. A stray comma.

bootstrap/config.aws-aws.tpl
```diff
@@ -15,65 +21,18 @@
...
-              },{{ end }}
-              "SecurityGroupIds": [ "{{ ref "/aws/securitygroupid" }}" ]
-            },
-            "Tags": {
-              "Name": "{{ ref "/aws/stackname" }}-manager",
-              "Deployment": "Infrakit",
-              "Role" : "manager"
-            }
-          }
+          {{ include "instace-aws.tpl" "config.aws-aws.tpl" }}
```
ndegory: s/instace-aws/instance-aws/

bootstrap/flavor-swarm-manager.tpl
```diff
@@ -0,0 +1,15 @@
+"Plugin": "flavor-swarm/manager",
```
ndegory: can we make these included files real JSON files (with the leading brace)?

bootstrap/config.aws-aws.tpl
ndegory: to get a cleaner repo, all the included files should be in a subdirectory (include), and the include statement could be {{ include "include/instance-aws.tpl" }}

ndegory: what is the use of the second argument? I don't see how it impacts the result.

bootstrap/instance-terraform.tpl
```diff
@@ -0,0 +1,15 @@
+"Plugin": "instance-terraform",
+"Properties": {
+  "type": "aws_instance",
```
ndegory: this file should probably be named instance-terraform-aws.tpl or we'll have conflicts once we're ready to add other cloud providers with the same plugin.

ndegory (changes requested):

bootstrap/config.aws-aws.tpl
JosephGJ: Fixed.

JosephGJ: Done.

JosephGJ: Originally, it was for giving context to the template file so that local variables could be passed to it before it was rendered. Since we are using globals now, this is obsolete.

bootstrap/flavor-swarm-manager.tpl
JosephGJ: Done.

bootstrap/instance-terraform.tpl
JosephGJ: Done.

subfuzion:
Going to chill this in the icebox for @ndegory / @JosephGJ until there's more time to fully vet.

ndegory:
@JosephGJ  we should refresh this PR based on the work done in infrakit PR 473

PR #895, opened 2017-04-06
mount /lib/modules in alpine

bquenin:
Check travis


subfuzion (approved):
Awesome! Nice way to bring it in over the finish line after a long marathon today, Bertrand!! Thanks!! 👍

PR #896, opened 2017-04-06
Reserved port range for exposed services

ndegory:
- dind swarm enabled in .travis.yml instead of mounting /lib/modules (tests are working again on Mac OS)
- open range 50101 to 50199


bquenin:
The RFC 6335 is explaining this (https://tools.ietf.org/html/rfc6335#section-8.1.2):

Ports in the Dynamic Ports range (49152-65535) have been specifically set aside for local and dynamic use and cannot be assigned through IANA. Application software may simply use any dynamic port that is available on the local host, without any sort of assignment. On the other hand, application software MUST NOT assume that a specific port number in the Dynamic Ports range will always be available for communication at all times, and **a port number in that range hence MUST NOT be used as a service identifier**.

ndegory:
I agree with the RFC guy, we should move to ports below 2^15, but it also impacts amplifier and the cli code.

bquenin (approved):
LGTM

subfuzion:
@bquenin @ndegory I'm aware of the RFC. We picked this port range specifically due to Axway's network restrictions (all outgoing ports are blocked except for a few critical ports and ports in the dynamic port range (or maybe all ports above 2^15, I've heard conflicting information)). We are only going to need establish and standardize one port in the assignable range for amp traffic, and once we're ready to designate it, we can ask Axway to open that port up. The other ports are just for our convenience right now to expose on localhost (like kibana) for testing. For those kinds of HTTP uses, we'll use 443 when we're ready and add back our reverse proxy service for forwarding traffic. If we want to decide on a port for amp now and ask Axway to open it up, we can do that. We'll want to choose a port in the user port range (1024-49151) that we can officially register with IANA as an amp service name.

PR #907, opened 2017-04-07
Infrakit Image with the Docker instance plugin

ndegory:
Development version of InfraKit with the Docker instance plugin integrated: one less container to run.


subfuzion (approved):
👍

PR #909, opened 2017-04-07
Cluster deployment on AWS

ndegory:
This PR adds the missing elements for AMP stack deployment on AWS.

- symlink for hack/deploy: hack/deploy-aws. The script is now able to deploy locally or on AWS
- the terraform deployment now has several outputs: the cluster id, the leader manager ip and all the ips of the bootstrapped instances.

to test:

- you should have aws credentials set in your home directory (~/.aws/credentials). Please refer to http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html#credentials-file-format.
- an aws keypair should be available in this region.
- add a terraform.tfvars in the bootstrap/terraform-aws folder with a content similar to
```
aws_name = "TAG"
aws_region = "us-west-2"
bootstrap_key_name = "AWS-KEYPAIR-ID"
aws_profile = "default"
infrakit_config_base_url = "https://raw.githubusercontent.com/appcelerator/amp/cluster-aws/bootstrap"
```
The key name should exist in AWS EC2 in this region and you should have the private key in your possession.
- run `hack/deploy-aws`
- while the script is running, the cluster id and the manager leader ip are printed
- add the leader IP (:5000) as an insecure registry on your host, restart the engine
- if the script is still running and has not tried to push images yet, keep waiting, else run `./hack/deploy-aws $clusterid`
- it should deploy the stacks on the cluster and validate the availability of nats and elasticsearch
- get the dns name of one of the node, and connect on port 50106, kibana should appear.


hack/deploy
```diff
@@ -14,13 +14,92 @@ readonly clustercheck="$SP/clustercheck"
 readonly curlcheck="$SP/curlcheck"
 readonly swarmcheck="$SP/swarmcheck"
 readonly servicescheck="$SP/servicescheck"
+readonly terraformdir="$SP/../bootstrap/terraform-aws"
+readonly stacks_to_deploy="amp1 amp2 amp3"
```
subfuzion: Do you need this (`stacks_to_deploy`)?

hack/deploy
```diff
@@ -14,13 +14,92 @@ readonly clustercheck="$SP/clustercheck"
...
+readonly terraform="docker run -v $HOME/.aws:/root/.aws:ro -v $terraformdir:/data --workdir=/data hashicorp/terraform"
+
+# IP range that should have access to the docker remote API
+readonly rapi_cidr="$(curl -sf ifconfig.co)/32"
```
subfuzion: For what it's worth, you can make the curl call faster by just querying the ip: `ifconfig.co/ip`.

hack/deploy
```diff
@@ -37,9 +116,32 @@ pushimage() {
   ok
 }
 
+
+echo "$(deployment_target) deployment"
```
subfuzion: A bit concerned about blossoming complexity of this script, but at least as a style issue, let's avoid executable code between function declarations (ref: Google's bash style guide for [Function Location](https://google.github.io/styleguide/shell.xml?showone=Function_Location#Function_Location)).

hack/deploy
```diff
@@ -14,13 +14,92 @@ readonly clustercheck="$SP/clustercheck"
...
+  local)
+    dockerhost=m1
+    registryhost=localhost
+    ;;
+  aws)
+    dockerhost=$($terraform output -state=/data/terraform.tfstate leader_ip)
+    registryhost=$dockerhost
+    # variable used in sub scripts
+    DOCKERHOST=$dockerhost
+    export DOCKERHOST
```
subfuzion: It's a bash script, so `export DOCKERHOST=$dockerhost` is fine.

subfuzion (commented):
Ok, happy to see aws deployment unified to a single starting point, but to be honest, I feel that my deploy script was already too complicated and wasn't happy with it. Now that complexity has increased. I don't like that we have multiple functions performing conditional case checks for `localhost` or `aws`. What happens when we support more provider backends? Really, the script should determine the deployment target then run `cluster.localhost.sh` or `cluster.aws.sh` specialized scripts for specialized cluster management implementation, and then this script should continue and run the generic smoke tests. I can live with it between now and our milestone deadline as long as we commit to a less brittle strategy afterwards, but we want to look for opportunities to reduce complexity, not increase it, as we proceed.

subfuzion:
Local deploy failed:

```
$ hack/dev
. . .
final service replica smoke check failed (exit 1)
clean up
destroy amp-manager-f573e897-7aa0-4516-a195-42ee91039e97 initiated
destroy amp-worker-f573e897-7aa0-4516-a195-42ee91039e97 initiated
destroy amp-proxy-f573e897-7aa0-4516-a195-42ee91039e97 initiated
infrakit has been destroyed
registry has been destroyed
client certificates have been removed
configuration volume deleted
```

Ran `hack/dev` again:

```
$ hack/dev
local deployment
cluster is not healthy, please fix the issue and try again (exit 1)
clean up
client certificates have been removed
no configuration volume to delete
```

Expected second deployment to succeed.

hack/deploy
ndegory: not anymore, I'll remove it

subfuzion (changes requested):
Local deployments fail, despite running `./bootstrap/bootstrap -d` or manually removing all containers and volumes, etc.

The issue is that `hack/dev` now fails, probably because it's supplying a fixed cluster id. I can run `hack/deploy` directly.

ndegory:
@subfuzion comments have been taken into account and local deployment with the hack/dev script is now working.

hack/swarmcheck
```diff
@@ -12,9 +15,18 @@ usage() {
...
 SECONDS=0
+typeset -i nc
+typeset -i ready
 while true; do
-    nc=$(docker node ls | awk '{ print $1 }' | wc -w)
-    accepted=$(docker node ls --filter membership=accepted | awk '{ print $1 }' | wc -w)
-    [[ $nc -eq $accepted ]] && exit 0
+    # number of nodes + 1 (header)
+    nc=$($amps node ls | wc -l)
+    if [[ $? -ne 0 ]]; then nc=0; fi
```
subfuzion, line 23: Please leave the commented out version of the previous filter test for `accepted` and a TODO comment to investigate whether this is a docker bug.

subfuzion (approved):
Other than the reservations I've already expressed about increasing script complexity, this looks good for now. Thanks @ndegory

PR #912, opened 2017-04-07
Adding an error if the email API key has not been provided

bquenin:
Amplifier now returns a fatal error is the SendGrid key has not been set.


subfuzion (approved):
LGFN

PR #914, opened 2017-04-07
Fixe monitoring stack name issue

freignat91:
## how to test

- ampmake clean build
- hack/deploy
- amp logs -i

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -2,6 +2,7 @@ package docker
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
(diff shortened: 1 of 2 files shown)


pkg/docker/docker.go
```diff
@@ -46,9 +48,14 @@ func (d *Docker) DoesServiceExist(ctx context.Context, name string) bool {
 	if err != nil || len(list) == 0 {
 		return false
 	}
+	suffix := fmt.Sprintf("_%s", name)
 	for _, service := range list {
-		if service.Spec.Annotations.Name == name {
-			return true
+		if strings.HasSuffix(service.Spec.Annotations.Name, suffix) {
+			for _, name := range service.Spec.Labels {
+				if name == "infrastructure" {
```
subfuzion, line 55: @freignat91 Thanks for your PR, but #922 fixes the connection issue by ensuring we specify service aliases and also placing all the required services in the same stack anyway, even though we deploy them in stages to reduce healthcheck churn. I wouldn't want to hardcode these checks as is into our docker wrapper anyway, but am open to the idea of supplying a regex as well as map of labels that need to match, and we can tackle this then if it turns out we do need this flexibility.

subfuzion (changes requested):

pkg/docker/docker.go
subfuzion: I recommend closing this PR for now, but you can decide if you want to update it or not.

freignat91: ok,i close it

PR #915, opened 2017-04-07
Fix linting and formatting issues

bquenin:
## How to test
    $ ampmake build lint


freignat91 (approved):
Tested ok on my machine, but seems not ok on trevis

freignat91 (changes requested):
Seems not ok on trevis

freignat91 (approved):
LGTM

ndegory (approved):
LGTM

cli/cli.go
```diff
@@ -27,13 +27,10 @@ type Interface interface {
 // cli implements cli.Interface
 type cli struct {
 	Configuration
-	version    string
-	build      string
```
subfuzion, line 31: I don't consider version and build as options. These were read-only through the interface/implementation exposed on cli.Interface and cli.

cli/cobra.go
```diff
@@ -28,7 +28,7 @@ func SetupRootCommand(rootCmd *cobra.Command) {
 	rootCmd.SetHelpCommand(helpCommand)
 
 	rootCmd.PersistentFlags().BoolP("help", "h", false, "Print usage")
-	rootCmd.PersistentFlags().MarkShorthandDeprecated("help", "please use --help")
+	rootCmd.PersistentFlags().MarkShorthandDeprecated("help", "please use --help") // nolint
```
subfuzion: I'm just curious ... what was the lint error we got for this? I prefer the more specific version of nolint that says the lint check to skip.

cli/in.go
```diff
@@ -53,7 +53,7 @@ func (i *InStream) SetRawTerminal() (err error) {
 // RestoreTerminal restores normal mode to the terminal
 func (i *InStream) RestoreTerminal() {
 	if i.state != nil {
-		term.RestoreTerminal(i.fd, i.state)
+		term.RestoreTerminal(i.fd, i.state) // nolint
```
subfuzion: Same here (please specify lint rule being ignored).

subfuzion: If it's about ignoring an error, it's probably better to just be explicit that we don't care; i.e.:  `_ = term.RestoreTerminal(...)`.

cli/logger.go
```diff
@@ -22,57 +22,59 @@ func init() {
 	// WARNING: the grpc logger can only be set during init()
 	// https://godoc.org/google.golang.org/grpc/grpclog#SetLogger
 	// TODO: set verbose to false after testing
-	grpclog.SetLogger(Logger{out: NewOutStream(os.Stdout), verbose: true})
+	grpclog.SetLogger(&Logger{out: NewOutStream(os.Stdout), verbose: true})
```
subfuzion: Converting to references, and converting all the functions to pointer methods made sense for Console, but I don't see a need here. There's only a need for a single instance, it's immutable once created, and it's a tiny struct. The change actually obscures that intention. See [value/pointer methods](https://golang.org/doc/faq#methods_on_values_or_pointers).

cli/required.go
```diff
@@ -6,18 +6,10 @@ import (
...
 
+// NoArgs checks that the command is not passed any parameters
 func NoArgs(cmd *cobra.Command, args []string) error {
 	if len(args) == 0 {
 		return nil
 	}
 	return fmt.Errorf("unexpected argument: %s\nSee '%s --help'", args[0], cmd.CommandPath())
 }
-
-func NoArgsCustom(errstr string) func(*cobra.Command, []string) error {
```
subfuzion: Was this causing an error (other than no comment)? Please don't remove.

cli/time.go
```diff
@@ -5,6 +5,7 @@ import (
 	"time"
 )
 
+// ConvertTime converts unix timestamp to time
```
subfuzion: Consider placing this under `pkg/time`.

subfuzion (changes requested):
@bquenin Thanks for tackling the lint issues. I've requested a few things, but overall looks good.

cli/cli.go
bquenin: Unused code

cli/cobra.go
bquenin: error check

cli/in.go
bquenin: ok

cli/logger.go
bquenin: sure

cli/required.go
bquenin: Unused code

cli/time.go
bquenin: ok

cli/required.go
subfuzion: I can do it, but if you don't mind, please add `// nolint: unused` above the func with your PR.

cli/cli.go
subfuzion: Unfortunately, I did a squash on that part of the history, but it's probably my fault. By adding the fields to Configuration to streamline `NewCLI`, they are now mutable. It's nothing earth-shattering, but I'd prefer that Configuration then extend the same constraint with a [getter](https://golang.org/doc/effective_go.html#Getters), delegated via cli. But since it wasn't you that broke the abstraction, I won't ask you to fix it (unless you want to). ;)

cli/cobra.go
subfuzion: `// nolint: error`

subfuzion (approved):

PR #919, opened 2017-04-07
More ports available for local deployment

ndegory:
When bootstrapping locally, we now have these ports mapped to the host:

- 50101 (amplifier)
- 50102 (available)
- 50103 (available)
- 50104 (available)
- 50105 (available)
- 50106 (Kibana)
- 50180 (available)
- 50190 (available)

how to test:

- hack/deploy
- deploy a stack with a service publishing on one of the available ports and check that it's accessible from the host.


freignat91 (approved):
LGTM

PR #922, opened 2017-04-08
Refactor stacks

subfuzion:
Creates aliases for infrastructure services so that dependent services can connect by the basic service name (eg, `nats`) instead of the stack-prefixed name (eg, `monitoring_nats`).

Modifies existing calls to services to skip the test to see if the service name exists and just attempts to connect to the service, returning an error as appropriate if the connection fails.

In addition, the deploy script now deploys all the amp-related stack files using the name `amp` so that subsequent deploys are treated as `amp` stack updates. The phased approach to loading the stack is just to minimize churn as services that have dependencies on other services are deployed later and have a better chance of being healthy.

Ran fmt as normal and caught other formatting issues that are included with this pr.

## Verification

    $ ampmake build
    $ hack/dev
    $ amp logs -i

Should see logs.

    $ . hack/ampenv
    $ amps service ls

Verify that all the services are healthy and listed as part of the amp stack.


ndegory (approved):
LGTM

PR #926, opened 2017-04-09
Build improvements

subfuzion:
This PR changes the `build-cli` rule to now build the local native `amp` binary and place it under `./bin/GOOS/GOARCH/` (eg, `./bin/darwin/amd64/`).

When you just want to update the CLI for your platform, run the following:

    $ make build-cli

Otherwise, to build all supported cross-compiled versions, run

    $ ampmake buildall-cli

The PR also adds a new quality rule. **Before submitting a pull request, ensure you run the following:**

    $ ampmake check

This ensures that cross-compiled versions of the CLI are built, formats source code, and performs a quick subset of the full lint checks (which the CI server will run).

This PR also

* fixes the build-server rule to ensure that the amplifier config is copied, whether running make locally or from the container (instead of hardcoding `/root` it uses `~` to point to the home directory) and also updates the mount path to *only* the `~/.config/amp` dir, not everything under `~/.config`
* enables color output
* updates the build config
* adds a few other cli-related build rules (`cleanall-cli`, rebuildall-cli`)
* adds a fast lint rule (`lint-fast`)


bquenin (approved):
LGTM (tested on linux)

PR #960, opened 2017-04-12
Update version template, normal color

subfuzion:
Normal color should be the terminal default color.
Version command updated to display "Client" and "Server" sections.

## Verification

```
$ amp version
Client:
 Version:       v0.6.0-dev
 Build:         0768dadf
 Address:       localhost:50101
 Go version:    go1.8
 OS/Arch:       darwin/amd64

Server:      
 Version:       v0.6.0-dev
 Build:         b485f3c7
 Go version:    go1.8.1
 OS/Arch:       linux/amd64
```

Output should be in default terminal color and now says `Client` (instead of `amp`) and `Server` (instead of `amplifier`).

### DIFF
--- a/cli/command/version/version.go
+++ b/cli/command/version/version.go
@@ -2,9 +2,8 @@ package version
 
 import (
 	"bytes"
-	"runtime"
-
 	"fmt"
+	"runtime"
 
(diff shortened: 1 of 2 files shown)


neha-viswanathan (approved):
Tested on dark and light themed terminals.

PR #969, opened 2017-04-13
TLS

ndegory:
- add a SSL termination service in the cluster (dockercloud/haproxy)
- enable it for kibana
- local deployment with self signed certificate (cluster deployment require a valid certificate)
- update the AMI references for AWS deployment

how to test:

    $ hack/deploy
    # then open https://dashboard.local.atomiq.io in your browser. You'll get a warning about the certificate because it's self-signed, check the issuer and the subject, then proceed to the service. You should be able to access the usual dashboard.


subfuzion (approved):
Looks great, @ndegory  Thanks. 👍

PR #971, opened 2017-04-13
Documentation update

neha-viswanathan:
All the existing documentation are now under `docs/legacy` directory. The CLI documentation can be found under `docs`, along with `README.md`.


docs/cluster.md
```diff
@@ -5,12 +5,13 @@ The `amp cluster` command manages all cluster-related operations for AMP.
...
 
-    Usage:	amp cluster [OPTIONS] COMMAND [ARGS...]
+    Usage:	amp cluster COMMAND
 
     Cluster management operations
 
     Options:
-          --help   Print usage
+      -h, --help            Print usage
+      -s, --server string   Specify server (host:port)
```
ndegory, line 14: should be `-s, --server string   Specify server (default "cloud.atomiq.io:50101")`
this is true for all commands, and should be implemented in the CLI.

ndegory (changes requested):

docs/cluster.md
bquenin: Actually, `-s` is a root command option that should only be displayed when using `amp --help` and not be repeated on all commands.

ndegory: Also, we'd like to keep a few md files from legacy, and update the README.md. Some links are broken now.

neha-viswanathan: The docs will be updated soon. For now, all the old files are under `legacy`.

ndegory (approved):

PR #992, opened 2017-04-14
fix registry slowness in local bootstrap

ndegory:
(see https://github.com/dock…er/for-mac/issues/1536)

Use 127.0.0.1 instead of localhost for the registry endpoint

### DIFF
--- a/hack/deploy
+++ b/hack/deploy
@@ -82,7 +82,7 @@ set_deployment_variables(){
   case $(deployment_target) in
   local)
     dockerhost=m1
-    registryhost=localhost
+    registryhost=127.0.0.1
     amps="docker run -t --rm --network=hostnet -v $(dirname $SP)/stacks:/stacks docker --host=$dockerhost"
     stacks_path=/stacks
     domainname=$localdomainname


subfuzion (approved):

subfuzion (approved):
I should have known better and used the loopback interface just to avoid network weirdness. But as noted by [this issue](https://github.com/docker/for-mac/issues/1536), using `127.0.01` does indeed work. I chatted with @stevvooe on slack and it seems that there is a separate issue with our cluster on AWS (possibly a regression) since we're using IP address, not DNS name, for the registry container. Will need further investigation and I'll open an issue on the https://github.com/docker/distribution/ repo, but at least this fixes what would have been a showstopper for anyone using the edge version of Docker for Mac.

subfuzion:
@ndegory I'm kicking myself still for specifying localhost and not the loopback interface. I'm going to be helping with an orchestration workshop at DockerCon and remember hearing this the slack channel last month -- and promptly forgot: https://github.com/jpetazzo/orchestration-workshop/commit/44f40179927ad2e56f056d76123b79246b1e0be0. I'm still concerned about the underlying issue since it affects us on AWS, so we'll definitely want to keep following this.

PR #1006, opened 2017-04-14
Updated ls, rm aliases

neha-viswanathan:
Updated `ls, list` and `rm, remove` aliases for the following commands:
- function
- org, org member
- stack
- team, team member
- user


subfuzion (approved):
👍

PR #1014, opened 2017-04-15
Append default port if not specified in --server flag

neha-viswanathan:
adds default port to --server option

### To test
```
$ amp -s localhost version

Client:
 Version:       v0.6.0-dev
 Build:         e75f63c5
 Server:        localhost:50101
 Go version:    go1.8
 OS/Arch:       darwin/amd64

Server:         
 Version:       v0.6.0-dev
 Build:         33dca96f
 Go version:    go1.8.1
 OS/Arch:       linux/amd64
```

### DIFF
--- a/cmd/amp/commands.go
+++ b/cmd/amp/commands.go
@@ -27,7 +27,6 @@ type opts struct {
 // newRootCommand returns a new instance of the amp cli root command.
 func newRootCommand(c cli.Interface) *cobra.Command {
 	opts := &opts{}
-
 	cmd := &cobra.Command{
 		Use:           "amp [OPTIONS] COMMAND [ARG...]",
 		Short:         "Deploy, manage, and monitor container stacks and functions.",
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"fmt"
 	"io"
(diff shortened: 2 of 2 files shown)


freignat91 (approved):
LGTM, just add how to test in the description

cmd/amp/commands.go
```diff
@@ -53,7 +57,7 @@ func newRootCommand(c cli.Interface) *cobra.Command {
 	cli.SetupRootCommand(cmd)
 
 	cmd.Flags().BoolVarP(&opts.version, "version", "v", false, "Print version information and quit")
-	cmd.PersistentFlags().StringVarP(&opts.addr, "server", "s", "", "Specify server (host:port)")
+	cmd.PersistentFlags().StringVarP(&opts.addr, "server", "s", "cloud.atomiq.io:50101", "Specify server (host:port)")
```
subfuzion: No, please don't add a default here

subfuzion (changes requested):

cmd/amp/commands.go
subfuzion: We only have a one single location for updating the default server to connect to. There should be no default for the flag here.

subfuzion (approved):

PR #1015, opened 2017-04-15
Logout of account after user is removed

### LINKED ISSUE #978: Removing a user who is currently logged in should also log them out
When a user logs in to AMP and tries to delete their own account, the user has to be logged out of the account automatically.

neha-viswanathan:
Resolves #978 
When a user logs in to AMP and tries to delete their own account, the user is logged out of the account.

### To test

```
$ ampmake build
$ make build-cli
$ hack/dev
$ amp user signup --name user1 --email <valid email> --password xxxx
$ amp user verify 7sdk3...
$ amp login --name user1 --password xxxx
$ amp whoami 
user1
$ amp user rm user1
User removed.
$ amp whoami
Error: you are not logged in. Use `amp login` or `amp user signup`.
```

### DIFF
--- a/cli/command/logout/logout.go
+++ b/cli/command/logout/logout.go
@@ -1,11 +1,8 @@
 package logout
 
 import (
-	"fmt"
-
 	"github.com/appcelerator/amp/cli"
 	"github.com/spf13/cobra"
-	"google.golang.org/grpc"
 )
 
 // NewLogoutCommand returns a new instance of the logout command.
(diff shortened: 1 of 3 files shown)


bquenin (approved):
LGTM

JosephGJ:
For future reference, this could cause issues with removing users other than yourself. It may be good to check if the user you are trying to remove is the user currently logged in in your cli.

bquenin:
Actually, you can only remove yourself, not any other user. (Except for the future admin, but that's another story)

Issue #1018, opened 2017-04-16
Wrong links when creating a user on a local cluster


ndegory:
After creating a local cluster, doing a `amp -s localhost:50101 user signup`, the login contains direction on how to login, but redirects to the cloud cluster: `amp verify...` It should explicitely give the -s option. The user won't guess.

bquenin:
Actually, it's impossible for `amplifier` to know if it is running locally or not. That's why we have a `PublicAddress` field in amplifier configuration (https://github.com/appcelerator/amp/blob/master/cmd/amplifier/server/configuration.go#L23).
This field is not used today, but we could set it using an environment variable and make the cluster deployment script set this environment variable accordingly. This way, we could update the templates to use the public address.

ndegory:
@bquenin now that we're about to have the environment variable in amplifier, how are we suppose to set it? Should it be in the stackfile?

bquenin:
No specific requirements on how to set it. It just need to be set by the deployment tool.

subfuzion:
Coping my comments from chat:

This is an easy problem to solve when it's HTTP, but my suggestion would be something similar. I'm surprised if gRPC doesn't tell us the address the client connection used to reach the server. I haven't researched to see if we can add message metadata if it isn't already there. We could always add it to the signup message so we then know what address the client used. That would be a more robust way of handling this in my opinion.

subfuzion:
Where there's a will, there's a way!  👍

PR #1020, opened 2017-04-16
[WIP] Adding environment variable for amplifier public address

bquenin:
ref #1018 

### How to test

    $ go install github.com/appcelerator/amp/cmd/amplifier
    $ export AMP_PUBLICHOST="this.is.a.test"
    $ amplifier
    amplifier (server version: , build: )
    2017/04/15 20:32:44 Configuration file successfully loaded
    Version: 
    Build: 
    Port: :50101
    PublicHost: this.is.a.test
    EtcdEndpoints: [http://etcd:2379]
    ElasticsearchURL: http://elasticsearch:9200
    NatsURL: nats://nats:4222
    DockerURL: unix:///var/run/docker.sock
    DockerVersion: 1.27
    2017/04/15 20:32:44 Connecting to Docker API at unix:///var/run/docker.sock version API: 1.27
    2017/04/15 20:32:44 Connected to Docker API at unix:///var/run/docker.sock
    2017/04/15 20:32:44 Listening on port: 50101

### DIFF
--- a/cmd/amplifier/server/configuration.go
+++ b/cmd/amplifier/server/configuration.go
@@ -10,9 +10,9 @@ import (
 )
 
 const (
-	DefaultPort          = ":50101"
-	DefaultPublicAddress = "local.appcelerator.io"
-	DefaultTimeout       = time.Minute
+	DefaultPort       = ":50101"
+	DefaultPublicHost = "127.0.0.1"
+	DefaultTimeout    = time.Minute
 )
 
 // Config is used for amplifier configuration settings
@@ -20,7 +20,7 @@ type Configuration struct {
 	Version          string
 	Build            string
 	Port             string
-	PublicAddress    string
+	PublicHost       string
 	EtcdEndpoints    []string
 	ElasticsearchURL string
 	NatsURL          string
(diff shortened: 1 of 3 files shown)


cmd/amplifier/server/configuration.go
```diff
@@ -10,17 +10,17 @@ import (
 )
 
 const (
-	DefaultPort          = ":50101"
-	DefaultPublicAddress = "local.appcelerator.io"
-	DefaultTimeout       = time.Minute
+	DefaultPort       = ":50101"
+	DefaultPublicHost = "127.0.0.1"
```
ndegory, line 14: we may have an issue once we're ready to do TLS termination outside of amplifier, if we use a different port, the exposed address won't match

bquenin: This is the "inside the container" port, which is mapped anyway.

bquenin: I can add an environment variable for port as well, or group both address and port into a single value, such as `PublicEndpoint`. Tell me what you think is best.

ndegory: the way it's done most of the time is a URL, when the port is not explicitly set the CLI knows which port to try. We can image that the CLI first try TLS/443, and falls back to non TLS/50101. Unless the scheme and port are explicitly set.

bquenin: gRPC is not really a well known scheme, and port guessing can be confusing for users. I'd rather set a global endpoint as "address:port" for clarity.

ndegory: do we have access to the scheme and authority from the grpc metadata?

bquenin:
Working on another solution for this issue

PR #1021, opened 2017-04-16
Amp user signup takes unique email

### LINKED ISSUE #977: Single email for multiple user accounts
When creating new users, 1 email can be used to create multiple accounts. Can cause issues with the `forgot-login` command, as the command should only return 1 user account name associated with the given email.

JosephGJ:
fixes #977 

`amp user signup` now will not allow you to register with an email that is assigned to an existing account.

To test:
```
$ ampmake build
$ make build-cli
$ hack/dev
$ amp user signup --name=user1 --email=<some email> --password=<some password>
$ amp user signup --name=user1 --email=<the same email> --password=<some password>
error: email is already in use
```


bquenin (changes requested):
Looks good. You just missed the error mapping in https://github.com/appcelerator/amp/blob/master/api/rpc/account/account.go#L24-L47
You need to map your new error to the `return grpc.Errorf(codes.AlreadyExists, err.Error())` block in the switch

bquenin (approved):

PR #1029, opened 2017-04-17
Add amp cluster node ls command

freignat91:
Add "amp cluster node ls" command, listing the nodes on the amp cluster

## test

- ampmake build
- make build-cli
- amp cluster create --localhost
- if needed: amp user signup --localhost
- if needed: amp login --localhost
- amp cluster node ls --localhost


ndegory:
we need a mechanism to protect this feature, it's not meant for all users. We don't want to expose the infrastructure to all users, only the service.
For the node management we should also be able to list the labels (for service scheduling) on the node and add/remove labels depending if you have the right to do so.

subfuzion (approved):
👍

PR #1043, opened 2017-04-21
Update shebangs

### LINKED ISSUE #1042: Hack scripts don't work with other shell than bash
```
$ echo $SHELL
/bin/zsh
$ hack/ampenv
hack/ampenv: 2: hack/ampenv: Bad substitution
```

bquenin:
Fixes #1042 

## How to test

```
$ zsh
$ source hack/ampenv
```


hack/ampenv
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
```
ndegory: this file is supposed to be sourced, so the shebang will be ignored by the script sourcing this file.

bquenin: Good catch

neha-viswanathan (approved):

PR #1054, opened 2017-04-24
Vendor update

### LINKED ISSUE #1053: Vendor issue
It's not possible to add a package in vendor, no matter the package, we got:

```
[ERROR] Error scanning github.com/gogo/protobuf/types: open /go/src/github.com/appcelerator/amp/.glide/cache/src/https-github.com-gogo-protobuf/types: no such file or directory
```

bquenin:
Fixes #1053

## To test
```
ampmake cleanall build
```


freignat91 (approved):
Test it ok adding packages for amp-ui. works well
I need it for amp-ui

cli/out.go
```diff
@@ -59,7 +59,7 @@ func (o *OutStream) GetTtySize() (uint, uint) {
 	}
 	ws, err := term.GetWinsize(o.fd)
 	if err != nil {
-		logrus.Debugf("Error getting size: %s", err)
+		log.Printf("Error getting size: %s\n", err)
```
subfuzion, line 62: 👍

subfuzion (changes requested):
@bquenin Some really good catches in the PR. I only have one concern -- let's discuss in person. It basically has to do with all the sub-packages. I don't think they're necessary; in fact, I think we have been over-using them for a while. I think the basic premise for sub-packages is for grouping packages from multiple VCS locations under one logical package, as explained here: https://github.com/Masterminds/glide#q-why-does-glide-have-the-concept-of-sub-packages-when-go-doesnt. Also, as the final tip mentions here, we should generally try to reference the base package name if that works. Maintaining vendoring health has been a major hassle, but hoping to move in the opposite direction toward more simplification...

bquenin:
I didn't added the subpackages by hand, I just let `glide create` do it for me just for the record. But sure, we can remove them.

subfuzion:
@bquenin Interesting. It's been maybe two months now since I recreated `glide.yaml`, and we've just been updating it since. I guess I wasn't expecting such an explosion of sub-packages. It doesn't seem consistent with its own documentation. I don't know that we need to remove anything that was auto-generated, I'm just so leery about adding to it. Every time we've run into a vendoring problem, I usually ended up resolving it by pruning down, not adding... Gotta love vendoring!

subfuzion (approved):

PR #1061, opened 2017-04-24
Revert to $BASH_SOURCE[0] instead of $0 in hack/ampenv

bquenin:
## To test
```
bash
. hack/ampenv
alias amps
```


subfuzion:
@bquenin I pushed a commit that I've confirmed works for both `sh` and `bash`. Can you please confirm for `zsh`? Thanks!

bquenin:
Unfortunately, it's not working on zsh.
```
$ source hack/ampenv
hack/ampenv:10: division by zero
$ source ./hack/ampenv
./hack/ampenv:10: bad floating point constant
```

hack/ampenv
```diff
@@ -1,4 +1,12 @@
...
+# It attempts to do this in a way that works for sh, bash, and ideally other shells like zsh.
+#
+# It relies on BASH_SOURCE being defined under bash, but does not attempt to access it using
+# bash array syntax (ie, BASH_SOURCE[0]); instead it expands BASH_SOURCE as a simple string,
+# which should provide just the pathname (see:
+# http://mywiki.wooledge.org/BashFAQ/028#Using_BASH_SOURCE).
+
 export AMPCID="f573e897-7aa0-4516-a195-42ee91039e97"
-export AMPHOME="$(dirname "$(cd "$(dirname "$0")" && pwd)")"
+export AMPHOME="$(dirname "$(cd "$(dirname "${BASH_SOURCE:$0}")" && pwd)")"
```
ndegory: zsh doesn't like it:
```
Switched to a new branch 'revert-ampenv'
~/Development/src/github.com/appcelerator/amp(revert-ampenv ✔) echo $SHELL
/bin/zsh
~/Development/src/github.com/appcelerator/amp(revert-ampenv ✔) unalias amps
unalias: no such hash table element: amps
~/Development/src/github.com/appcelerator/amp(revert-ampenv ✔) . ./hack/ampenv
./hack/ampenv:10: bad floating point constant
```

But it you fix the parameter substitution with :- instead of : it works for zsh, and I guess other shells.

bquenin: You should probably commit your fix @ndegory, and let's be done with it!

ndegory (approved):
:LGTM

PR #1068, opened 2017-04-25
version of images in core stacks default to latest

ndegory:
partial fix for #1067 

amplifier, agent, ampbeat, httpfunc and httpexec image versions are now variables, defaulting to `latest`

Since AMP v0.9.0, `amp cluster create -s localhost` will deploy the `latest` tag of these images instead of `local`.
After v0.9.0, to install another tag the hack/dev script will force the tag `local`, unless you use it with the environment variable TAG set to a different value.

## How to test

First, let's make sure the amp CLI 0.9.0 is working as expected
```
# download amp CLI 0.9.0
$ docker rmi appcelerator/amp-bootstrap:1.0.1
$ amp cluster create -s localhost
$ docker exec m1 docker service ls
ID            NAME               MODE        REPLICAS  IMAGE
8vkwk5am0ef5  amp_agent          global      3/3       appcelerator/agent:latest
95ov13bxxsnw  amp_funcexec       replicated  1/1       appcelerator/funcexec:latest
endc6e5r87v4  amp_ampbeat        replicated  1/1       appcelerator/ampbeat:latest
gytupvt3g9ir  amp_kibana         replicated  1/1       appcelerator/kibana:5.3.0-3
hk44erxnvtvn  amp_proxy          global      1/1       dockercloud/haproxy
pf1i9bnq519a  amp_amplifier      global      1/1       appcelerator/amplifier:latest
pwv5jfb31qrq  amp_elasticsearch  replicated  1/1       appcelerator/elasticsearch-amp:5.3.0
v7r0m6qm26io  amp_nats           replicated  1/1       appcelerator/amp-nats-streaming:v0.3.8
w1yr2lhmrvec  amp_etcd           replicated  1/1       appcelerator/etcd:3.1.5-1
yb3yfggp72ix  amp_funchttp       replicated  1/1       appcelerator/funchttp:latest
```
Notice the `latest` tag for these five services

Then, make sure the development workflow is also working as expected (forcing tag)
For this checkout this branch and check the following use cases.
```
$ hack/dev
$ docker exec m1 docker service ls
ID            NAME               MODE        REPLICAS  IMAGE
8vkwk5am0ef5  amp_agent          global      3/3       appcelerator/agent:local
95ov13bxxsnw  amp_funcexec       replicated  1/1       appcelerator/funcexec:local
endc6e5r87v4  amp_ampbeat        replicated  1/1       appcelerator/ampbeat:local
[...]

PR #1069, opened 2017-04-25
Add option to specify the image tag for core amp services

### LINKED ISSUE #1067: Cluster create command should have option to specify image tag
We currently set `:local` for the core image tags so that during development we're using the most recently built image on our own machines. However, this approach won't work for CLI users who have not built any of the images themselves. Using compose [environment variable substitution](https://docs.docker.com/compose/compose-file/#variable-substitution), we should make the default be `:latest` so that the latest images will be pulled from the repo. We should add a CLI option (e.g., `amp cluster create --tag=TAG`) to allow the tag to be overridden to a specific image tag during development (normally `local`), which the CLI will provide to the bootstrap process. We will need to update the bootstrap image accordingly, as well as publish the appropriate images with tags, and an updated CLI.

subfuzion:
Closes #1067, depends on #1068 

This PR adds the `--tag|-t` option to the CLI to specify the tag to be used for the collection of cluster images.

Equivalent to the default:

    $ amp cluster create --tag=latest

For local development:

    $ amp cluster create --tag=local

Explicit version tag:

    $ amp cluster create --tag=0.9.0


ndegory (approved):
LGTM

PR #1072, opened 2017-04-25
Add amp -ui

freignat91:
add amp-ui, client Angular 4, server Go, first version

## test

- ampmake clean build
- amp cluster create --server localhost --tag local

Once the cluster is up, we have for now to launch manually the infrastructure ampui.stack.yml stack which now contains amp-ui:

    $amp stack deploy -c stacks/ampui.stack.yml ui

- then open Chrome and browse url:  https://ui.local.atomiq.io

image

image


amp-ui/client/src/app/services/stacks.service.ts
```diff
@@ -0,0 +1,50 @@
...
+
+@Injectable()
+export class StacksService {
+  stacks : Stack[] = []
+  stack : Stack = null
+  onStacksLoaded = new Subject();
+  onStacksError = new Subject();
+
+  constructor(private httpService : HttpService) {
+    this.stacks.push(new Stack("907345782", "haproxy", 1, "freignat", "USER"))
```
subfuzion, line 14: @freignat91 What's all this?

glide.lock
```diff
@@ -1,5 +1,5 @@
 hash: fff401d3955c816110993c628fa41dd5f983036587ee137479f478b3e967dccb
-updated: 2017-04-26T21:39:21.015444832Z
+updated: 2017-04-26T23:59:00.032458759Z
```
subfuzion, line 2: Glide update shouldn't be included in PR.

amp-ui/client/src/app/services/stacks.service.ts
freignat91: It's a hardcoded stack i used for test, to be removed

glide.lock
freignat91: ok, i should be able to remove this vendor update amp-ui doesn't need

subfuzion: Ok, but I thought that's what the vendor update from yesterday was for?

subfuzion (changes requested):
@freignat91 Great work overall! I'll be honest, I'm on the fence on one thing, however: one advantage of this particular gateway approach is we can point to any amp server (although I don't see this as a big use case right now); but a big disadvantage is that we now need to manage a separate server, and manually forward/transform what was automatically mapped for us using grpc REST gateway support. This approach creates more build artifacts, duplicates code and creates more code paths to be tested, and doesn't exploit available automated code generation that's generally preferable over manual effort. This is something we will want to discuss and resolve before we go too far down a particular path and the codebase for this blossoms. Let's take time for an offline discussion over this. The path I really prefer is to support auto-generation for multiple grpc clients, and then (and only if really necessary) build a lightweight JavaScript client library that simplifies any remaining logic for establishing a connection, setting any custom HTTP headers for authentication tokens, etc, that make it easy to call the backend.

subfuzion:
@freignat91 Updating the description `docker run --rm -v $PWD/stacks:/stacks:ro --network hostnet docker -H m1 stack deploy -c stacks/ampui.stack.yml amp` -> `amp stack deploy -c stacks/ampui.stack.yml ui`.

subfuzion (approved):

subfuzion:
Nice work, @freignat91. Approving as is and we'll continue the discussion about alternatively restoring the REST gateway support over the next day or so. 👍

PR #1073, opened 2017-04-25
Voting app example

bquenin:
This is the voting app example from docker (https://github.com/docker/example-voting-app) adapted to AMP.

## To test
```
amp cluster create
amp --server=localhost stack deploy -c examples/stacks/voting-app/voting.stack.yml voting
```

Then you should be able to reach:
 - the voting app:  https://vote.local.atomiq.io
 - the result app: https://result.local.atomiq.io


neha-viswanathan (approved):

examples/stacks/voting-app/LICENSE
```diff
@@ -1,191 +0,0 @@
-
```
subfuzion: We shouldn't remove Docker's license.

subfuzion (changes requested):
I don't we should remove the files that are associated with the project (we can append our own Copyright statement after Docker's in the LICENSE), although I think it's ok to rename the original README.md to Instavote.md and provide a link to it from our README.md -- which is missing!

examples/stacks/voting-app/LICENSE
bquenin: I'll add it back

subfuzion (approved):

subfuzion (approved):
Nice! 👍

PR #1094, opened 2017-04-28
Fix CLI color output

bquenin:


subfuzion (commented):
Sorry, closing this because I made those changes to only colorize the line prefixes based on my own preferences, but after I did discuss with the team.

Issue #1063, opened 2017-04-25
Core services HA


ndegory:
This issue is not about scaling the core services, only HA.

The core stack includes elasticsearch and etcd. These services are databases and are designed to work as a cluster.
Deploying it as a Swarm service is not trivial, the way each task should be started is dependant of the other tasks and is not always symetric.
So far I've not found any serious cluster implementation of etcd or elasticsearch with Docker Swarm.

Constraints to consider to have a production ready mechanism:
- container healthcheck should not be based on the whole cluster status (Elasticsearch healthcheck colors are about the full cluster, so yellow doesn't mean that container has any issue, it means the replication across the cluster is not healthy). No operation should be done when the cluster is yellow.
- rolling update should be cluster aware: a blind rolling update across the nodes may injure a cluster if it doesn't give enough time to each instance to stabilize, which can take a long time on a cluster with a lot of data.
- how are we suppose to deal with the loss of a server, will a new one just spawn a new container, integrate the cluster, and replace the missing container without any data loss?
- dnsrr endpoint mode stack file setting is now a network option with Docker 17.05, this may help for the unicast discovery of elasticsearch (and maybe etcd) using the service name (it will get the task IPs instead of the single VIP). We can start with a fixed number of nodes (3 for instance), that can help the discovery mechanism.
- The scaling strategy should be considered in parallel of the persistency strategy. Can we recover a node storage or do we start from scratch?

PR #1097, opened 2017-04-28
elasticsearch cluster

ndegory:
ref #1063 

- elasticsearch and kibana 5.3.2
- elasticsearch cluster

how to test:
```
$ hack/dev
...
# if you have jq, if not, parse the json output and ask yourself why you haven't install jq yet!
$ docker exec m1 docker run --rm --network ampnet appcelerator/alpine:3.5.2 curl elasticsearch:9200/_cluster/state | jq '.nodes | length'
3
```

### DIFF
--- a/images/elasticsearch/Dockerfile
+++ b/images/elasticsearch/Dockerfile
@@ -1,9 +1,9 @@
 FROM appcelerator/alpine:3.5.2
 
-RUN apk update && apk upgrade && apk --no-cache add openjdk8-jre
+RUN apk update && apk upgrade && apk --no-cache add openjdk8-jre bind-tools
 
 ENV PATH /bin:/opt/elasticsearch/bin:$PATH
-ENV ELASTIC_VERSION 5.3.0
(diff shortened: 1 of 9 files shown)


subfuzion:
Lol, well, you can also ask why you don't just containerize everything since you can pipe to jq in a container. Since we install it in our `gotools` image, which `amptools` is based on, you can also just do this:

$ alias jq="docker run -i --rm appcelerator/amptools jq"
$ docker exec m1 docker run --rm --network ampnet appcelerator/alpine:3.5.2 curl elasticsearch:9200/_cluster/state | jq '.nodes | length'

Why use a package manager when you have Docker? Containerize all the things! :p

subfuzion:
@ndegory ^^ although I should mention we might want to update `gotools` since the `jq` version is still `jq-1.4-1-e73951f` and I believe `1.5` is the latest...

subfuzion (approved):
Excellent! 👍

PR #1098, opened 2017-04-28
InfraKit refresh

### LINKED ISSUE #1065: Development workflow and core stack update
When testing updates on core stack files, hack/dev fails to update the stacks accordingly.
A Docker volume has the stacks in cache and is not refresh at each run.
As a workaround one can remove the volume before running hack/dev:
```
$ docker volume remove amp-stacks
$ hack/dev
```

ndegory:
- infrakit updated to the dev version (infrakit 0.5 is not yet released)
- configuration update accordingly
- trap exit signal for cleanup
- update AMI refs in cloudformation template
- fix #1065 

how to test
```
$ hack/dev
# the deployment should be successful
$ docker volume ls | grep amp-stacks
# the volume shouldn't exist
```


subfuzion (changes requested):
@ndegory Failing checks...

ndegory:
Build on Travis was failing because the `kill -s HUP` on the Docker daemon of the DinD containers was killing the containers, which is not the case on Docker for Mac and even on Docker 17.04 on Ubuntu 16.04. Disabling this signal which was intended to reload the Docker configuration (daemon.json) with new labels, creating issue #1123 for this.

subfuzion (approved):
LGTM

PR #1104, opened 2017-04-28
Default stack name

### LINKED ISSUE #1081: Stack deploy should infer the stackname from the stack file name if none is provided
For instance:
```
$ amp stack deploy -c awesome.yml
```
Should create a stack named `awesome`.

```
$ amp stack deploy -c awesome.stack.file.yml
```
Should create a stack named `awesome`.

neha-viswanathan:
Closes #1081 

If no stack name is provided during stack creation, a default stack name is assigned. 
For example: 
`amp -s localhost stack start -c examples/stacks/pinger/pinger.yml` creates a stack with a default stack name `pinger`

### To test
`$ amp -s localhost stack start -c examples/stacks/pinger/pinger.yml`
or
`$ amp -s localhost stack start -c stacks/visualizer.stack.yml`

### DIFF
--- a/cli/command/stack/deploy.go
+++ b/cli/command/stack/deploy.go
@@ -3,6 +3,8 @@ package stack
 import (
 	"errors"
 	"io/ioutil"
+	"path/filepath"
+	"strings"
 
 	"github.com/appcelerator/amp/api/rpc/stack"
 	"github.com/appcelerator/amp/cli"
@@ -12,7 +14,6 @@ import (
 )
 
 type deployOpts struct {
-	name string
 	file string
 }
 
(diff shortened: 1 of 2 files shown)


cli/command/stack/deploy.go
```diff
@@ -26,26 +28,37 @@ func NewDeployCommand(c cli.Interface) *cobra.Command {
...
 	cmd.Flags().StringVarP(&opts.file, "compose-file", "c", "", "Path to a Compose v3 file")
 	return cmd
 }
 
-func deploy(c cli.Interface) error {
-	c.Console().Printf("Deploying stack %s using %s\n", opts.name, opts.file)
+func deploy(c cli.Interface, args []string) error {
+	var name string
+	charArr := []string{".", "_"}
+	if len(args) == 0 {
```
subfuzion: Let's put validation  code like this in a pkg utility class. We shouldn't be doing any substitution, just checking that it's a valid domain name component and returning an error if not. Basically a valid name is a valid domain name component that can contain letters and digits, is 1 - 63 characters in length, starts and ends with a letter or digit, and can contain sequences of one or more hyphens and underscores in the string (hostnames can't contain underscores, but domain name components can -- a valid hostname is a valid domain name component, even if a valid domain name component isn't a valid hostname). For go, I think a regex like the following would work:

`^[a-zA-Z0-9]{1}$|^[a-zA-Z0-9][a-zA-Z0-9-_]{1,61}[a-zA-Z0-9]$`

I did (non-exhaustive) testing of the regex [here](https://regex-golang.appspot.com/assets/html/index.html) and it seemed good enough for now. You could dig through Docker's code and see what they're doing to validate. In fact, we might just be better off not validating at all and letting Docker reject the name if it's not valid... (this would definitely be the most expedient way for now).

subfuzion (changes requested):

cli/command/stack/deploy.go
neha-viswanathan: ok I'm removing the validation for now.

freignat91 (approved):
LGTM, after Tony updates done

cli/command/stack/deploy.go
```diff
@@ -26,26 +28,31 @@ func NewDeployCommand(c cli.Interface) *cobra.Command {
...
 	cmd.Flags().StringVarP(&opts.file, "compose-file", "c", "", "Path to a Compose v3 file")
 	return cmd
 }
 
-func deploy(c cli.Interface) error {
-	c.Console().Printf("Deploying stack %s using %s\n", opts.name, opts.file)
+func deploy(c cli.Interface, args []string) error {
+	var name string
+	if len(args) == 0 {
+		name = strings.Split(strings.TrimSuffix(filepath.Base(opts.file), ".yml"), ".")[0]
```
subfuzion: You're assuming the extension will always be `.yml`, but `.yaml` is also valid. I would just take the filename base regardless of extension. If it's not a valid yaml file, the the deploy will fail anyway.

subfuzion (changes requested):
See comment.

cli/command/stack/deploy.go
subfuzion: I think something like the following would work just fine:

    basename := filepath.Base(opts.file)
    name = strings.TrimSuffix(basename, filepath.Ext(basename))

subfuzion (approved):

Issue #1105, opened 2017-04-28
[UX enhancement] amp service ls


neha-viswanathan:
`amp service ls` should list all user services. AMP internal services are always excluded.
```
$ amp service ls
[localhost:50101]
ID        NAME          MODE        REPLICAS  STATUS            IMAGE                       TAG
h8kub1n   go_counter    replicated  3/3       RUNNING           htilford/go-redis-counter   latest
Yii77om   redis_counter replicated  2/3       STARTING          redis                       latest
r45fe3d   foo_counter   global      0/0       NO MATCHING NODE  foo                         latest
```

freignat91:
Yes, but only the services the current users can see, so only the services of the stacks the user is owner to or have permission to see

JosephGJ:
ref: #1216 

moving to 0.11.0 milestone

PR #1112, opened 2017-05-01
Add stack service api and cli

freignat91:
- Add api verb stack.Services for UI to get all services of a stack
- Add CLI command amp stack services 

## test

- ampmake clean build
- amp cluster create --server localhost
- amp login/signup if needed
- amp stack up -c examples/stacks/pinger/pinger.yml pinger --server localhost 
- amp stack services pinger --server pinger


bquenin:
Is it related to #1105 ? It seems really close, yet different, from what has been discussed.

FR: This one list only the services for one stack. 
A global 'service ls' showing all user allowed stacks services is not handleable right now. 
Could be added later when we could get list of allowed stacks for a user

freignat91:
moved to branch portal

PR #1125, opened 2017-05-02
[Refactor CLI] Code cleanup

neha-viswanathan:
Updated CLI code to update args/options and remove the ones not in use and also updated command usage.


subfuzion (approved):
Nice job, @neha-viswanathan 👍

PR #1128, opened 2017-05-02
Restoring amplifier-gateway

### LINKED ISSUE #1124: Restore gRPC gateway for amplifier
We need to restore the gRPC gateway so that the UI can invoke `amplifier` through REST APIs.

bquenin:
Fixes #1124 

Verification: `platform/tests/gateway/pr-1128`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,7 @@ cleanall: clean cleanall-deps
 # When running in the amptools container, set DOCKER_CMD="sudo docker"
 DOCKER_CMD ?= "docker"
 
-build-base: install-deps protoc build-server build-beat build-agent build-ui build-bootstrap
+build-base: install-deps protoc build-server build-gateway build-beat build-agent build-ui build-bootstrap
 build: build-base buildall-cli
 
 # =============================================================================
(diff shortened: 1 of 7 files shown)


bquenin:
@subfuzion I checked if it was possible or idiomatic to embed the gRPC gateway inside the amplifier process, however, it doesn't seem to be the case ...
I think we should stick to the canonical usage for now and circle back to it later

stacks/amp.stack.yml
```diff
@@ -31,6 +31,26 @@ services:
...
+      - "SERVICE_PORTS=80"
+      - "VIRTUAL_HOST=https://gw.*"
+    deploy:
+      mode: global
+      labels:
+        io.amp.role: "infrastructure"
+      restart_policy:
+        condition: on-failure
+      placement:
+        constraints: [node.role == manager]
```
subfuzion: I don't think we need this constraint (global, manager) since it doesn't need access to the docker socket on a swarm manager and I don't believe there's any kind of guarantee that it communicates with the amplifier instance co-located on the same node for any potential performance benefit. I believe it's fine to place the gateway on workers and let them scale independently from amplifier instances running on managers.

subfuzion (approved):

stacks/amp.stack.yml
bquenin: ok sure

subfuzion (changes requested):
I see pros and cons to doing it this way, but the more I think about it, the more I think it's better that we do consolidate this to a single process (amplifier) instead of a separate gateway process. The code that it adds to amplifier is minimal relative to all the changes we're adding here to support yet another process in our stack. See [#6 here](https://github.com/grpc-ecosystem/grpc-gateway#user-content-usage) and also this [example](https://github.com/philips/grpc-gateway-example/blob/master/cmd/serve.go).

subfuzion (approved):
Looks good @bquenin. Merging although I'd like to pursue having this consolidated into the amplifier process.

subfuzion:
`make clean build` will not succeed for anyone who hasn't installed the full toolchain.

```
$ make clean build
api/rpc/account/account.proto
/Users/tony/go/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis: warning: directory does not exist.
github.com/golang/protobuf/ptypes/empty/empty.proto: File not found.
github.com/appcelerator/amp/api/rpc/account/account.proto: Import "github.com/golang/protobuf/ptypes/empty/empty.proto" was not found or had errors.
github.com/appcelerator/amp/api/rpc/account/account.proto:11:39: "google.protobuf.Empty" is not defined.
```

`ampmake clean build` is the generic way that will take care of this ensuring the same versions of all tools are consistently used for building and also ensure the necessary binaries are created for the actual host platform.

The steps should be:

    $ ampmake build
    $ amp --server localhost cluster create --tag local
    $ ...

bquenin:
@subfuzion sorry, it's an habit typo :)

subfuzion (approved):
Nice! 👍

PR #1133, opened 2017-05-03
Pass-through registry for local deployments

ndegory:
The deployment scripts already include a registry where we can push images not available on an official repository. This registry does not cache images from Docker Hub.
For local deployments this PR adds a second registry dedicated for image cache, boosting the deployment cache.

- bump the VERSION file
- catch infrakit template errors
- add a pass-through registry for local deployments

test with ```make build-cli && amp -s localhost cluster create -t local```


subfuzion (approved):
LGTM. Please resolve conflicts so I can merge.

PR #1140, opened 2017-05-03
amp-bootstrap 1.2.0

ndegory:
- build appcelerator/amp-bootstrap 1.2.0 (already pushed on DockerHub)
- use it by default in the CLI

$ make build-cli
$ amp -s localhost cluster create


subfuzion (approved):
LGTM

PR #1145, opened 2017-05-04
ampmake shouldn't change file permissions in the repo

### LINKED ISSUE #1110: ampmake build breaks file permissions on Linux
the containerized build chain breaks permissions on Linux.

```
$ find . ! -user $USER
# should be empty
$ ampmake build
$ find . ! -user $USER
# many files are now owned by root, mostly glide cache, vendor and binaries
```

ndegory:
fix #1110 

ampmake builds a local image on Linux systems customized for the user running ampmake, the permissions of the files changed by the build process are then still owned by that user on the host.
No change on Docker for Mac.

To test:
```
# first make sure all files in your repo are owned by your user:
find . ! -user $USER
# if not the case, fix it:
sudo chown -R $USER:$(id -g) .
ampmake clean build
# the build should succeed and file permissions should not be changed
find . ! -user $USER
```


bquenin (approved):
LGTM :+1:

subfuzion (approved):
LGTM. 👍

PR #1148, opened 2017-05-04
check for potential errors before running CLI cluster commands

### LINKED ISSUE #1086: When amp cluster create fails...
...we should provide more helpful guidance (eg, `use: amp cluster rm`). We should also have an `amp cluster create --force|-f` option) so the `rm` step can be skipped.

```
$ amp cluster create --tag=local
[localhost:50101]
local deployment
preparing a Docker image with the configuration data
warning: force start, removing Docker volume amp-bootstrap
Error response from daemon: unable to remove volume: remove amp-bootstrap: volume is in use - [155b5d5fc5b218689ce7df01c9c9d105ec5d3d35f91d95507a275354e171bc72]
force start failed, unable to remove the Docker volume amp-bootstrap, it's probably still used by an infrakit container
bootstrap failed (exit 1)
clean up
infrakit has been destroyed
registry has been destroyed
configuration volume deleted
f573e897-7aa0-4516-a195-42ee91039e97
Error: exit status 1
```

ndegory:
fix #1086 

check for potential errors before running CLI cluster commands and print hints on how to fix it if it's the case
```
$ make build-cli
$ amp -s localhost cluster create
...
# in a second terminal while the first command is still processing
$ amp -s localhost cluster create
[localhost:50101]
Error: the amp-bootstrap container is already running. Another CLI is probably still running, wait for it to stop or kill it. If that's not the case, you should force remove the container with `docker rm -f amp-bootstrap` and try again.
```

### DIFF
--- a/cli/command/cluster/check.go
+++ b/cli/command/cluster/check.go
@@ -1,0 +1,26 @@
+package cluster
+
+import (
+	"errors"
+	"os/exec"
+)
+
+// check if a cluster command is not already in progress, and provides hints if it's the case
+func check(provider string) error {
+	cmd := "docker"
+	switch provider {
+	case "local":
+		// check that the amp-boostrap container does not exist
+		args := []string{
+			"container", "inspect", "amp-bootstrap",
+		}
+		proc := exec.Command(cmd, args...)
+		err := proc.Run()
+		if err == nil {
+			return errors.New("A cluster operation is still in progress. Please wait for it to finish unless you want to terminate it now (docker rm -f amp-bootstrap) and try again.")
+		}
+	default:
+		// no check
+	}
+	return nil
+}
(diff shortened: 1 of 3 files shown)


subfuzion (changes requested):
@ndegory  I think the check is a good idea, but I don't see using this command in my normal workflow (ie, run `amp cluster check` before other cluster commands). A better approach would be to run this check as part of each cluster command, displaying the error and exiting when appropriate.

Also, I would simplify the message =>

> A cluster operation is still in progress. Please wait for it to finish unless you want to terminate it now (`docker rm -f amp-bootstrap`) and try again.

ndegory:
@subfuzion, you're not supposed to run `amp cluster check`.
I'll update the message.

subfuzion (approved):
@ndegory My mistake for assuming you implemented a command (I wrote my comment around 2 AM last night); you already implemented it exactly as I was proposing in my comment. Thanks. 👍

PR #1158, opened 2017-05-06
system checks for local bootstrap

ndegory:
Check system prerequisites, in particular for Linux systems:
- vm.max_map_count should be at least 262144
- ip_vs module should be loaded

### DIFF
--- a/cli/command/cluster/bootstrap.go
+++ b/cli/command/cluster/bootstrap.go
@@ -10,7 +10,7 @@ import (
 
 const (
 	bootstrapImg = "appcelerator/amp-bootstrap:%s"
-	bootstrapTag = "1.3.0"
+	bootstrapTag = "1.3.1"
 )
 
 var (
--- a/platform/bootstrap/bootstrap
+++ b/platform/bootstrap/bootstrap
@@ -195,15 +195,29 @@ EOF
 }
 
 _system_check() {
-  docker info | grep -q "Swarm: active" && return 0
-  if [ $? -ne 0 ] && [ $(uname) = "Darwin" ]; then
-    echo "Warning: Swarm mode is not enabled on your Mac, the DinD cluster will probably not work correctly"
-    return 1
+  local mmcmin=262144
+  typeset -i mmc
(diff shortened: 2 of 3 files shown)


bootstrap/bootstrap
```diff
@@ -759,6 +773,9 @@ if [ -n "$swarm_join_ip" ]; then
   _init_swarm $swarm_join_ip || exit 1
   exit 0
 fi
+if [[ "x$target" = "xlocal" ]]; then
```
subfuzion: Since it's our image with a modern bash and this is a bash script, I think it's ok to simplify with the following:

    [[ $target = "local" ]] && ( _system_check || exit 1 )

bootstrap/bootstrap
```diff
@@ -195,15 +195,29 @@ EOF
...
-  lsmod | grep -q ip_vs
-  if [ $? -ne 0 ]; then
-    echo "Warning: the IPVS module is not loaded, consider modprobing ip_vs or alternatively initializing your host Docker Swarm"
-    return 1
+
+  # if swarm is initialized, the modules will already be loaded
+  docker info | grep -q "Swarm: active" && return 0
+  # on Linux system, we have to make sure that the modules are loaded
+  if [[ $(uname) = "Linux" ]]; then
+    lsmod | grep -q ip_vs
```
subfuzion: It's probably better to grep on `ip_vs_rr`, since then we know `ip_vs` is also loaded.

bootstrap/bootstrap
```diff
@@ -195,15 +195,29 @@ EOF
...
+    if [[ $? -ne 0 ]]; then
+      echo "Warning: the IPVS modules are not loaded, consider loading them ($ sudo modprobe ip_vs_rr) or alternatively initializing your host Docker Swarm"
```
subfuzion: Recommend: "Warning: required IPVS modules are not loaded. See: https://github.com/appcelerator/amp/tree/master/docs#linux

subfuzion (approved):

PR #1165, opened 2017-05-09
UI as default virtual host

### LINKED ISSUE #1036: Point cloud.atomiq.io at proper server
It doesn't appear that cloud.atomiq.io is pointed at a proper server. It should at least have a web page to describe it, or if the service isn't running, we should create it somewhere.

ndegory:
fix #1036

cloud.appcelerator.io (or local.appcelerator.io) will route on the UI service.


subfuzion (dismissed):
LGTM

subfuzion (approved):

PR #1178, opened 2017-05-10
Disable notifications

### LINKED ISSUE #1177: Disable amplifier notifications
We need a way to disable amplifier notifications.
The following environment variable should control the `amplifier` service:
 - `NOTIFICATIONS`: can be `true`(default) or `false`.

bquenin:
Resolves #1177 

## How to test
```
$ ampmake build
$ alias amp='amp -s localhost'
$ amp cluster create --tag=local --registration=none --notifications=false
$ amp user signup --name user --password password --email email@user.amp
$ amp login --name user --password password
$ amp org create --org org --email email@org.amp
$ amp logs -i amp_amplifier | grep Notifications
[localhost:50101]
Notifications: false
```
You should not receive an emails during that test.


cmd/amplifier/server/server.go
```diff
@@ -53,28 +65,32 @@ var serviceInitializers = []serviceInitializer{
...
-		log.Fatalf("Unable to listen on %s: %v\n", c.Port[1:], err)
+		log.Fatalf("Unable to listen on %s: %v\n", a.config.Port[1:], err)
 	}
-	log.Println("Listening on port:", c.Port[1:])
+	log.Println("Listening on port:", a.config.Port[1:])
 	log.Fatalln(s.Serve(lis))
 }
 
-func initClients(config *Configuration) {
+func initClients(amp *Amplifier) {
```
subfuzion: The changes look good. My only observation is that instead of passing around all these pointers, perhaps we should just make all of these Amplifier methods.

api/rpc/logs/logs.go
```diff
@@ -20,8 +20,8 @@ import (
 
 // Server is used to implement log.LogServer
 type Server struct {
-	Es            *elasticsearch.Elasticsearch
-	NatsStreaming *ns.NatsStreaming
+	Es *elasticsearch.Elasticsearch
+	Ns *ns.NatsStreaming
```
subfuzion: Bike-shedding a bit here, but consider `Ns` -> `NS` (and `Es` to `ES` while you're at it). See [Initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms), the slide [here](https://talks.golang.org/2014/names.slide#6), and [Golint tests](https://github.com/golang/lint/blob/master/lint_test.go#L226-L227).

bquenin: Sure

cmd/amplifier/server/server.go
bquenin: Why not, in that case, we probably should remove the `clientInitializer  func(*Amplifier) error` type since I don't think it's possible to define a function type with a receiver.

subfuzion (approved):
👍

PR #1181, opened 2017-05-10
CLI config file

### LINKED ISSUE #1175: CLI configuration file
The CLI configuration file should follow [FSH](http://www.pathname.com/fhs/pub/fhs-2.3.html#HOMEUSERHOMEDIRECTORIES) convention as mentioned in issue #1153 and should therefore be a dot directory called `.amp`.
This directory will be looked up using this sequence:
 - current folder(`./.amp`),
 - user home directory (`$HOME/.config/amp`).

The content of the `.amp` directory is:
 - `amp`: the configuration file for the CLI.
 - `$serverAddress.credentials`: The credentials used for the given serverAddress.

Note that credentials should only be stored in user's home directory.

neha-viswanathan:
Resolves #1175 

The smoke tests are located in `platform/tests/config` 

You should see a `localhost.credentials` file in $HOME/.config/amp directory
If no server address is specified on the command, the server settings specified in the CLI config file is read. 
Note that the CLI config file can be found in one of the following locations - 
- current working directory (`./.amp`)
- user's HOME directory  #(`$HOME/.config/amp`)

If no config file is found or no server address is specified on the command, the default server address is used.

### DIFF
--- a/platform/tests/config/config_test.sh
+++ b/platform/tests/config/config_test.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+# server passed on commandline overrides both local and home configs
+test_cli_local_home() {
+  # create local config
+  mkdir -p $PWD/.amp
+  echo "Server: LOCAL" > $PWD/.amp/amp.yml
+
+  # create home config
+  mkdir -p $HOME/.config/amp
+  echo "Server: HOME" > $HOME/.config/amp/amp.yml
+
+  # server address passed as command line argument
+  amp -s SERVER version | pcregrep -q "Server:[[:space:]]+SERVER"
+  local ec=$?
+
+  rm -R $PWD/.amp
+  rm -R $HOME/.config/amp
+  return $ec
+}
+
+# server passed in local config overrides home config
+test_local_home() {
+  # create local config
+  mkdir -p $PWD/.amp
+  echo "Server: LOCAL" > $PWD/.amp/amp.yml
+
+  # create home config
+  mkdir -p $HOME/.config/amp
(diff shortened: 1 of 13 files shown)


bquenin (dismissed):
LGTM

platform/tests/cli/configuration
```diff
@@ -0,0 +1,126 @@
...
+mkdir -p $PWD/.amp
+echo "Server: present" > $PWD/.amp/amp.yml
+
+# for config file in home directory
+mkdir -p $HOME/.config/amp
+echo "Server: home" > $HOME/.config/amp/amp.yml
+
+# no config file; server address passed as command line argument
+amp $addr version | grep -Eq "Server:[[:space:]]+command-line"
+if [[ $? -ne 0 ]]; then
```
ndegory: if you use `set -e` this will never be executed.

platform/tests/config/config_test.sh
```diff
@@ -0,0 +1,115 @@
...
+
+  # for config file in home directory
+  mkdir -p $HOME/.config/amp
+  echo "Server: home" > $HOME/.config/amp/amp.yml
+
+  # server address passed as command line argument
+  amp $addr version | grep -Eq "Server:[[:space:]]+command-line"
+
+  rm -R $PWD/.amp
+  rm -R $HOME/.config/amp
```
subfuzion: This won't work. The test will pass as long as the last command passes. If you want to clean up at the end of your test, just make sure to preserve the result of the previous grep (`local ec=$?`) and then return it at the end of the function (`return $ec`).

bquenin (approved):
LGTM

Issue #944, opened 2017-04-11
HA for deployment tool


ndegory:
Right now the InfraKit container runs only on one manager and uses the OS mode to manage it's plugins.
InfraKit is able to run on all 3 managers and use the Swarm annotations to share the deployment state across the nodes.
Once implemented in our deployment scripts, we should check that the loss of a manager node leads to the election of a new InfraKit leader (as well as a new Swarm manager leader), and that when the missing manager is reestablished it is correctly incorporated in the cluster (Swarm manager and Infrakit, both non leader).

JosephGJ:
This is possible to do using `infrakit-manager swarm` functionality which allows for the synchronization of the infrakit leaders with the docker swarm leaders. Along with this, the election process of the new leader (in the case that the leader goes down) involves freeing any groups managed by the current swarm leader infrakit node and committing them to the new leader. 

However, because we are using terraform to manage the health of the instances within the recommitted groups, the `.tfstate` files that are generate on the original m1 node are not moved over to the new leader in the instance that m1 is terminated, and so the health of the worker group can no longer managed by infrakit.

PR #1182, opened 2017-05-11
[multi-swarm] HA Swarm managers + infrakit on AWS

### LINKED ISSUE #1150: Manager nodes bootstrap
The first implementation of the cluster deployment (managed swarm) was done with the help of the  InfraKit Terraform instance plugin. It doesn't meet our expectations (in particular a generic tool for deployment on cloud providers) and is not compatible with a HA deployment (Terraform state files should also be distributed).
Moving to native instance plugins.

Part of it is a generic deployment of a 3 instances cell with swarm managers and InfraKit, with automated leader election. The InfraKit configuration template will be committed by another mean, from remote.

JosephGJ:
fixes #1150, #944 

Updates to the cloudformation stack .yml file for use with the current deployment. New cloudformation deploys 3 manager nodes with an Autoscaling group into 3 separate subnets in different Availability zones within the given region. Each manager determines the leader node for the swarm via the AWS cli and docker remote api. 

In addition, the bootstrap script is modified for better HA deployment as now Infrakit runs on each manager and the manager plugin runs in swarm mode, so that the management of the worker group can be reassigned to different managers.

Some limitations to be worked on:
 - Currently, we can only deploy into regions with 3 availability zones.
- Workers are only deployed onto a single subnet, for the moment, and instead they should be deployed using another autoscaling group.
- There are two different plugin files for both swarm and os modes for the infrakit-manager plugin. This should be modified for better code reuse.

### DIFF
--- a/platform/bootstrap/bootstrap
+++ b/platform/bootstrap/bootstrap
@@ -325,8 +325,10 @@ _instance_plugin_compute_subtype(){
 _run_ikt_container() {
   local _should_wait_for_plugins=0
   local _infrakit_image
-  docker run --rm -v $BOOTSTRAP_VOLUME:$INFRAKIT_HOME alpine:3.5 sh -c "echo 'group' >> $INFRAKIT_HOME/leader" || exit 1
 
+  if [ "$manager_plugin" = "os" ]; then
+    docker run --rm -v $BOOTSTRAP_VOLUME:$INFRAKIT_HOME alpine:3.5 sh -c "echo 'group' >> $INFRAKIT_HOME/leader" || exit 1
+  fi
   # remove old customized configuration just in case
   docker run --rm -v $BOOTSTRAP_VOLUME:$INFRAKIT_HOME alpine:3.5 sh -c "sed -i.bak 's/^{{ var \"\/script\/baseurl\".\*$//' $INFRAKIT_HOME/env.ikt" || exit 1
   local _extra_plugins=""
@@ -356,7 +358,7 @@ _run_ikt_container() {
       local _ip
       _ip=$(_get_ip)
       docker run --rm -v $BOOTSTRAP_VOLUME:$INFRAKIT_HOME alpine:3.5 sh -c "echo '{{ var \"/bootstrap/ip\" \"$_ip\" }}' >> $INFRAKIT_HOME/env.ikt" || exit 1
-      echo "bootstrap ip ($_ip) added in env.ikt" >&2 
+      echo "bootstrap ip ($_ip) added in env.ikt" >&2
     fi
     if [ "x$provider" = "xdocker" ]; then
         echo "making sure the network $BRIDGE_NETWORK is created" >&2
@@ -370,7 +372,7 @@ _run_ikt_container() {
     docker run -d --restart always --name infrakit \
          $INFRAKIT_OPTIONS $INFRAKIT_PLUGINS_OPTIONS -e PLUGIN_DIR=$INFRAKIT_HOME/${provider} -e INFRAKIT_LOG_LEVEL=$INFRAKIT_LOG_LEVEL $_infrakit_image \
          infrakit plugin start --wait --config-url $CONTAINER_PLUGINS_CFG --exec os --log $INFRAKIT_LOG_LEVEL \
-         manager group-stateless flavor-swarm flavor-vanilla flavor-combo $_extra_plugins >&2
+         manager-${manager_plugin} group-stateless flavor-swarm flavor-vanilla flavor-combo $_extra_plugins >&2
     _should_wait_for_plugins=1
   else
     echo "infraKit container is already started" >&2
@@ -598,6 +600,7 @@ MANAGED_PLUGINS="docker aws"
 target=local
 provider=docker
 default_provider=1
+manager_plugin=swarm
 envfile=""
 providerfile=""
 manager_count=1
@@ -700,7 +703,12 @@ trap _finish EXIT
 if [ "x$target" != "xlocal" ] && [ $default_provider -eq 1 ]; then
     provider="$target"
 fi
+
 subtype=$(_instance_plugin_compute_subtype $provider)
+
+if [ "x$target" = "xlocal" ]; then
+    manager_plugin=os
+fi
 
 if [ -n "$status_request" ]; then
   _status "$status_request"
@@ -722,8 +730,7 @@ if [ $init_swarm -eq 1 ]; then
   _init_swarm || exit 1
 fi
 if [ -n "$swarm_join_ip" ]; then
-  _init_swarm "$swarm_join_ip" || exit 1
-  exit 0
+  _init_swarm $swarm_join_ip || exit 1
 fi
 [[ $target = "local" ]] && ( _system_check || exit 1 )
 _set_source "$1"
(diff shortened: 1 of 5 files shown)


platform/bootstrap/bootstrap
```diff
@@ -325,8 +325,10 @@ _instance_plugin_compute_subtype(){
 _run_ikt_container() {
   local _should_wait_for_plugins=0
   local _infrakit_image
-  docker run --rm -v $BOOTSTRAP_VOLUME:$INFRAKIT_HOME alpine:3.5 sh -c "echo 'group' >> $INFRAKIT_HOME/leader" || exit 1
 
+  if [ "$manager_plugin" == "os" ]; then
```
ndegory: the comparison operator should be =, not ==

platform/bootstrap/bootstrap
```diff
@@ -700,8 +703,13 @@ trap _finish EXIT
 if [ "x$target" != "xlocal" ] && [ $default_provider -eq 1 ]; then
     provider="$target"
 fi
+
 subtype=$(_instance_plugin_compute_subtype $provider)
 
+if [ "x$target" == "xlocal" ]; then
```
ndegory: the comparison operator should be =, not ==

platform/bootstrap/bootstrap.yml
```diff
@@ -4,24 +4,97 @@ Description: Docker Swarm cluster deployed with InfraKit
...
+    subnet3:
+      cidr: 192.168.32.0/24
     vpc:
       cidr: 192.168.0.0/16
 
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+    - Label:
+        default: Swarm Size
```
ndegory: Swarm size should be merged in the Swarm properties

platform/bootstrap/bootstrap.yml
```diff
@@ -4,24 +4,97 @@ Description: Docker Swarm cluster deployed with InfraKit
...
+        default: Number of Swarm managers?
+      ManagerDiskSize:
+        default: Manager ephemeral storage volume size?
+      ManagerDiskType:
+        default: Manager ephemeral storage volume type
+      WorkerSize:
+        default: Number of Swarm workers?
+      KeyName:
+        default: Which SSH key to use?
+      EnableCloudWatchLogs:
```
ndegory: we should remove the cloudwatch settings for now, we should rely on a more generic monitoring system

platform/bootstrap/bootstrap.yml
```diff
@@ -31,40 +104,44 @@ Parameters:
...
-    ConstraintDescription: must be the name of an existing EC2 KeyPair
-    Description: Name of an existing EC2 KeyPair to enable SSH access to the bootstrap instances
-    MinLength: '1'
-  KeyName:
-    Type: AWS::EC2::KeyPair::KeyName
-    ConstraintDescription: must be the name of an existing EC2 KeyPair
-    Description: Name of an existing EC2 KeyPair to enable SSH access to the instances
-    MinLength: '1'
+    ConstraintDescription: Must be a valid EC2 HVM instance type.
+    Default: t2.micro
```
ndegory: I would upgrade the default to t2.medium

platform/bootstrap/bootstrap.yml
```diff
@@ -31,40 +104,44 @@ Parameters:
...
     Type: String
     AllowedValues:
+      - standard
+      - gp2
+    Default: standard
+    Description: Manager ephemeral storage volume type.
+  WorkerInstanceType:
+    Type: String
+    AllowedValues:
+    - t2.nano
```
ndegory: we should add more options: m4.xlarge, m4.2xlarge, c4.large, c4.xlarge, c4.2xlarge, c4.4xlarge, r4.large, r4.xlarge, r4.2xlarge, r4.4xlarge.

platform/bootstrap/bootstrap.yml
```diff
@@ -180,16 +231,114 @@ Resources:
...
+        ToPort: '22'
+      - CidrIp: 0.0.0.0/0
+        IpProtocol: tcp
+        FromPort: '80'
+        ToPort: '80'
+      - CidrIp: 0.0.0.0/0
+        IpProtocol: tcp
+        FromPort: '443'
+        ToPort: '443'
+      - CidrIp: 0.0.0.0/0
```
ndegory: this one should be limited to the VPC at least

platform/bootstrap/bootstrap.yml
```diff
@@ -201,6 +350,7 @@ Resources:
           Principal:
             Service:
             - ec2.amazonaws.com
+            - autoscaling.amazonaws.com
```
ndegory: do you really need it?

platform/bootstrap/bootstrap.yml
```diff
@@ -300,36 +453,69 @@ Resources:
...
+  AutoScalingGroup:
+    Type: AWS::AutoScaling::AutoScalingGroup
+    DependsOn:
+      - PublicSubnet1
+      - PublicSubnet2
+      - PublicSubnet3
+    UpdatePolicy:
+      AutoScalingRollingUpdate:
+        MaxBatchSize: 1
+        MinInstancesInService: !Ref ManagerSize
```
ndegory: the min instances in service shouldn't be the target size of the autoscaling group. It would be better to have that value minus one, but we don't want to calculate it here. I would leave it to 0, we have the batch size and the pause time already to give time to other instances to stabilize.

platform/bootstrap/bootstrap.yml
```diff
@@ -300,36 +453,69 @@ Resources:
...
+        PauseTime: PT20M
```
ndegory: 20 min is maybe too much. 5 or 10 min should be enough to allow an instance to stabilize.

platform/bootstrap/bootstrap.yml
```diff
@@ -300,36 +453,69 @@ Resources:
...
+           - !Ref PublicSubnet3
+  AsgLaunchConfig:
+    Type: AWS::AutoScaling::LaunchConfiguration
+    DependsOn:
+      - ProvisionerInstanceProfile
+      - SecurityGroup
+    Properties:
+      AssociatePublicIpAddress: true
+      BlockDeviceMappings:
+        - DeviceName: /dev/xvdb
```
ndegory: I'm confused. you can use ephemeral disk (vs EBS), which is what the /dev/xvd[a-e] range is reserved for, and you can't set the size and type for ephemerals. But be aware ephemeral disks are not supported anymore on recent instance types.

platform/bootstrap/plugins-swarm.json
```diff
@@ -0,0 +1,73 @@
...
+                "Cmd" : "infrakit-group-default --poll-interval 10s --name group-stateless --log 5 > {{env "INFRAKIT_HOME"}}/logs/group-stateless.log 2>&1"
+            }
+        }
+    }
+    ,
+    {
+        "Plugin" : "instance-docker",
+        "Launch" : {
+            "os" : {
+                "Cmd" : "infrakit-instance-docker --log 5 > {{env "INFRAKIT_HOME"}}/logs/instance-docker.log 2>&1"
```
ndegory: I thought we used a variable for the log level

ndegory (changes requested):

platform/bootstrap/config.aws-aws.tpl
```diff
@@ -117,9 +46,9 @@
                 "Plugin": "flavor-swarm/worker",
                 "Properties": {
                   "InitScriptTemplateURL": "{{ var "/script/baseurl" }}/worker-init.tpl",
-                  "SwarmJoinIP": "{{ var "/m1/ip" }}",
+                  "SwarmJoinIP": "{{ var "/bootstrap/ip" }}",
                   "Docker" : {
-                    "Host" : "tcp://{{ var "/m1/ip" }}:{{ var "/docker/remoteapi/port" }}"
+                    "Host" : "tcp://{{ var "/bootstrap/ip" }}:{{ var "/docker/remoteapi/port" }}"
```
ndegory: the Docker Host parameter should be localhost (it's for the InfraKit client, and it's always running on the same node as the swarm leader)

platform/bootstrap/plugins-swarm.json
```diff
@@ -0,0 +1,73 @@
+[
```
ndegory: a way to keep a single plugins.json file would be to declare two plugins for manager: manager-os and manager-swarm, these are just identifiers, the Cmd value effectly names the plugin.

ndegory (changes requested):

ndegory (approved):

platform/bootstrap/bootstrap.yml
```diff
@@ -300,36 +452,64 @@ Resources:
...
+    DependsOn:
+      - PublicSubnet1
+      - PublicSubnet2
+      - PublicSubnet3
+    UpdatePolicy:
+      AutoScalingRollingUpdate:
+        MaxBatchSize: 1
+        MinInstancesInService: 0
+        PauseTime: PT10M
+        WaitOnResourceSignals: true
```
ndegory: this parameter goes with a tool call cfn-signal, specific to AWS. Either we install and configure it for a good Autoscaling experience, or we just set this setting to false for the time being (and genericity).

PR #1187, opened 2017-05-12
Updated stack name during stack creation

### LINKED ISSUE #1136: Stack name validation
At stack creation, if no stack name is specified, a default stack name is assigned. But with the current stack name validation, the only special character allowed in the stack name is a `-` and in the example below, stack creation fails.
```
$ amp stack deploy -c stacks/visualizer.stack.yml 
[localhost:50101]
Deploying stack visualizer.stack using stacks/visualizer.stack.yml
Error: name is invalid
```

Is there a particular naming convention to be followed for stack files or should there be change in backend validations to include other special characters?

neha-viswanathan:
Closes #1136 

The smoke tests are located under `platform/tests/stacks/up`


platform/tests/cli/stacks
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+amp="amp -s localhost"
+$amp user signup --name user --password password --email email@user.amp
```
ndegory: you probably want to use "--registration=none"

platform/tests/cli/stacks
```diff
@@ -0,0 +1,11 @@
...
+$amp login --name user --password password
+$amp stack up -c platform/stacks/visualizer.stack.yml
+$amp stack ls | grep "visualizer" | wc -l
```
ndegory: `$amp stack ls 2>/dev/null | grep -Eq "\svisualizer\s"` would be better

platform/tests/cli/stacks
```diff
@@ -0,0 +1,11 @@
...
+if [[ $? -ne 1 ]]; then
```
ndegory: `if [[ $? -ne 0 ]]; then` if you use the grep -Eq suggested above

platform/tests/cli/stacks
```diff
@@ -0,0 +1,11 @@
...
+    echo "error"
+    exit 1
```
ndegory: we should remove the stack before exiting the script (if the stack has been created)

ndegory (changes requested):

platform/tests/stacks/up/up_test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+amp="amp -s localhost"
+
+$amp login --name user1 --password password
+$amp stack up -c platform/stacks/visualizer.stack.yml
```
ndegory: I'd prefer to see an absolute path, you're making an assumption that the test is run at the root of the project

ndegory (dismissed):

ndegory (approved):
LGTM

platform/tests/stacks/up/up_test.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+test_setup() {
```
subfuzion, line 3: @appcelerator/amp Just be aware that the test runner doesn't have formal support for test_setup/test_teardown; this works right now because the test runner doesn't currently run tests in parallel. Most developers will assume test_setup and test_teardown is something that runs before and after *each* test; for something to run just once for the file, we will need to add `setup()` and `teardown()` support (but those won't match the current pattern for `test_*()`).

PR #1195, opened 2017-05-16
Integration tests for accounts and storage

bquenin:
## How to test
Check travis status


subfuzion (dismissed):
The set up adds a TON of time to the CI build: https://travis-ci.org/appcelerator/amp/builds/233333848#L842-L976

.travis.yml
```diff
@@ -24,6 +24,7 @@ script:
   - ls -la bin/linux/amd64
   - amp --server=localhost cluster create --tag=local --registration=none --notifications=false
   - TESTINCLUDE=platform/testing testrunner platform/tests
+  - tests/run
```
subfuzion, line 27: We shouldn't need to build an image for the CI build. The entire environment is already up and running on travis, so you should be able to execute `go test -v github.com/appcelerator/amp/tests/integration/` directly (or a script that executes that directly. Building the image adds a *ton* of time to the CI build, although it makes sense for local testing (non-CI testing).

subfuzion (changes requested):

.travis.yml
bquenin: Some of the tests are accessing services that are not exposed outside of our swarm (`etcd` or `nats` for instance). These are the options I see:
 - Create an integration test container (that's what I did),
 - Expose these ports externally for tests purposes,
 - Mount the source code of `amp` in the `DinD` cluster and use this for tests.
 - Use a local swarm instead of a `DinD` swarm.

subfuzion: Ah, I see ... I didn't look at it closely enough, but I understand the problem you're trying to solve: you want tests that can run inside the swarm. In that case, we can look at optimizing the build time by using an existing image that already has all the tools needed (like `amptools`) (or we can create a lighter weight one specifically for these tests). This is an optimization we can worry about later. Thanks!

subfuzion (dismissed):

subfuzion (approved):
👍

PR #1197, opened 2017-05-16
Test enhancements

subfuzion:
Adds a new test runner for bash scripts to handle platform smoke tests. The test runner and test helper scripts are located under `platform/testing`. The test runner is called `testrunner`.

`testrunner` looks for test scripts in the test path, which by default is a subdirectory of the working directory called `tests`, but the location can be overridden by specifying the path as an argument (ex: `testrunner /path/to/tests`).

`testrunner` looks for test helper scripts to include (eg, `testing.sh`) in the working directory, but this can be overridden by setting the `TEST_INCLUDE` environment variable (ex: `TEST_INCLUDE=/path/to/include testrunner`).

`testrunner` supports 3 levels of verbosity. The default is `1`, but this can be overridden by specify the level as an environment variable (ex: `VERBOSE=0 testrunner`).
  `0` - only print the test result summary
  `1` - print the output of failing tests, plus the test result summary and list of failed tests
  `2` - print passing and failing test output, plus the test result summary and lists of both failed and passing tests

The test path points to a directory that is the top level test suite. A test suite consists of tests and any child test suites located in subdirectories.

The tests of a test suite are any files at a particular directory level that match one of the following:
 * `*setup.sh`
 * `*_test.sh`
 * `*teardown.sh`

Tests are run in subshells and processed in the following order:
 1. Source all the files in the directory that match `*setup.sh`. This evaluates each script and if any script fails at this level, the suite fails and no further tests are run at this level or below. These count toward passing and failing tests based on their exit (`return`) code.
 2. Run any functions in the setup script(s) that match the pattern `test_x()`. If any tests fail, the suite fails and no further tests are run at this level or below. These count toward passing and failing tests based on their exit (`return`) code.
 3. Source all the files in the directory that match `*.test.sh`. This evaluates each script and these count toward passing and failing tests based on the exit code.
 4. Run any functions in the test script(s) that match the pattern `test_x()`. These count toward passing and failing tests based on their exit (`return`) code.
 5. After running the tests from step 4:
  5.1 If there are no subdirectories, source all the files in the directory that match `*teardown.sh`. This evaluates each script and these count toward passing and failing tests based on the exit code. Run any functions in the teardown script(s) that match the pattern `test_x()`. These count toward passing and failing tests based on their exit (`return`) code.
  5.1 If there are subdirectories, run steps 1-4 on each subdirectory before step 5.1


bquenin (approved):
LGTM :+1:

ndegory (approved):
LGTM

PR #1198, opened 2017-05-17
Add windows binaries

bquenin:
This modification provide `ampmake.exe` for windows and add windows to the `buildall-cli` make target.


hack/xbuild
```diff
@@ -8,20 +8,23 @@ LDFLAGS=$4
...
-    alpine amd64
+    windows amd64
 )
 
 for ((i=0; i<"${#TARGETS[@]}"; $((i=$i+2)) )); do
   GOOS="${TARGETS[i]}"
   GOARCH="${TARGETS[i+1]}"
-  echo $DEST/$GOOS/$GOARCH/$NAME
+  OUTPUT=$DEST/$GOOS/$GOARCH/$NAME
+  if [[ $GOOS = "windows" ]]; then OUTPUT=$OUTPUT".exe"; fi
```
ndegory: this is OK, but it's generally safer to write it as follow:
OUTPUT="$DEST/$GOOS/$GOARCH/$NAME"
[[ "$GOOS" = "windows" ]] && OUTPUT="${OUTPUT}.exe"

cmd/ampmake/build
```diff
@@ -7,11 +7,14 @@ DEST=$GOPATH/src/$REPO/bin
...
+    windows amd64
 )
 
 for ((i=0; i<"${#TARGETS[@]}"; $((i=$i+2)) )); do
   GOOS="${TARGETS[i]}"
   GOARCH="${TARGETS[i+1]}"
-  echo $DEST/$GOOS/$GOARCH/ampmake
-  GOOS=$GOOS GOARCH=$GOARCH go build -o $DEST/$GOOS/$GOARCH/ampmake $PKG || exit $?
+  OUTPUT=$DEST/$GOOS/$GOARCH/ampmake
+  if [[ $GOOS = "windows" ]]; then OUTPUT=$OUTPUT".exe"; fi
```
subfuzion: I recommend doing what I've done in other places and working with strings like this: `OUTPUT="${OUTPUT}.exe"`. You should do it in the other location as well (`xbuild`). `make` doesn't care about quotes per se, but preserves and passes them literally to the shell for any commands that are run, so it's really about ensuring the shell gets properly quoted strings where it matters. It's even more important that we're diligent now about quoting path strings since we're adding Windows support now...

ndegory (dismissed):
LGTM

subfuzion (approved):
Thanks @bquenin

PR #1200, opened 2017-05-17
Fix test runner

bquenin:


subfuzion (approved):
LGTM. I updated it a bit as well.

PR #1205, opened 2017-05-18
Modifications to support single-manager local bootstrap

### LINKED ISSUE #1190: Local deployment on local Swarm manager node
A default deployment mode that relies on the host being a swarm manager and doesn't use Docker-in-Docker to build a swarm cluster, and doesn't require pushing images to the local swarm cluster. The current local cluster mode that relies on InfraKit and creates a full isolated cluster using DinD can remain an option.
This is a prerequisite for the Windows CLI, so it may be possible to provide a Window CLI with this release or otherwise the next one.

subfuzion:
Fixes #1190 

    $ platform/bin/deploy


subfuzion:
@ndegory @bquenin Thanks for the review and updates.

Issue #1220, opened 2017-05-19
Browser experience on UI for local deployment should be easier


ndegory:
The steps to make the UI work on a local deployment are too complex (involves the clipboard and the KeyChain Access tool).
At most, it should be a single click on the browser to validate a local certificate.

bquenin:
We should maybe stick to `http` instead of `https` for local experience. What do you think?

ndegory:
that would be the easy solution

ndegory:
the change in #1222 doesn't seem to improve that much the experience: it still should be added in a trusted store, just accepting the certificate in Chrome is not enough.

ndegory:
One thing to consider: TLS is forced on the service depending on the value of the FORCE_SSL environment variable. If set to 1, TLS is forced. We should consider variabilizing it, setting it to 0 for local deployments.

PR #1222, opened 2017-05-19
wildcard local certificate

ndegory:
ref #1220

wildcard certificate for local deployments

The generated certificate has a Subnet Alt Name for *.local.appcelerator.io and *.local.atomiq.io.
It also includes the (local) CA certificate. After import in a trust store, a user only has to set the X509 operations as trusted on the CA.

### DIFF
--- a/platform/bin/mkcert
+++ b/platform/bin/mkcert
@@ -2,6 +2,36 @@
 
 set -e
 which openssl >/dev/null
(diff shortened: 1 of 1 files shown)


freignat91 (approved):
The certificate with wilcard doesn't allow to use the UI one shot after the first validation, but anyway the wildcard stays better than 'dashboard' for our usage

PR #1234, opened 2017-05-20
Portal update for certificate

### LINKED ISSUE #1221: Portal cleanup (followup of #1209)
gtw.yml
```
+      io.amp.role: "infrastructure"
+    environment:
+      - "SERVICE_PORTS=80"
+      - "VIRTUAL_HOST=https://gw.*"
```
you don't need redirection from http to https?

gtw.yml
```
+      - "SERVICE_PORTS=80"
+      - "VIRTUAL_HOST=https://gw.*"
+    ports:
+      - "8080:80"
```
do you need to expose on port 8080?

gtw.yml
```
+      - "VIRTUAL_HOST=https://gw.*"
+    ports:
+      - "8080:80"
+    deploy:
```
we should add a placement constraint to restrict it on the worker nodes (see other stack files)

api/rpc/stats/stats.proto
```
-  MetricsIOEntry io = 14;
-  MetricsMemEntry mem = 15;
-  MetricsNetEntry net = 16;
+  string sgroup = 13;
```
shouldn't we add new fields at the end, instead of breaking the ids?

pkg/docker/docker.go
```
@@ -155,6 +158,7 @@ func (d *Docker) StackList(ctx context.Context) (output string, err error) {
 	return string(output), nil
 }
 
+// StackRemove remoce a stack
```
s/remoce/remove/

platform/stacks/portal.stack.yml
```
+services:
+
+  portal:
+    image: appcelerator/amp-ui:1.0.0
```
we started aligning the versions of the artifacts. Please use the "${TAG:-0.10.1}" version as for other core services.

platform/stacks/portal.stack.yml
```
+    image: appcelerator/amp-ui:1.0.0
+    networks:
+      - default
+    deploy:
```
should contain a placement constraint too

Please also remove portal/.DS_Store, it contains a binary

freignat91:
closes #1221

Address most of Nicolas comment (1221)
Add clear error message if certificate not imported i browser
change build name from amp-ui to portal


ndegory (changes requested):
Looks good, but still missing a placement constraint on the portal service.

FR: Which placement constraint,  portal service can be anywhere, you mean a constraint on workers?
NDE: yes, please add this:
```
      placement:
        constraints:
        - node.role == ${NODE_ROLE:-worker}
```

portal/client/src/app/auth/signup/signup.component.ts
```diff
@@ -29,16 +30,20 @@ export class SignupComponent implements OnInit {
       return
     }
     if (event.form.value.password != event.form.value.passwordVerif) {
-        this.message = "your password is not the same than your password confirmation"
+        this.messageError = "your password is not the same than your password confirmation"
```
subfuzion: => "passwords must match"

portal/client/src/app/auth/signup/signup.component.ts
```diff
@@ -29,16 +30,20 @@ export class SignupComponent implements OnInit {
       return
     }
     if (event.form.value.password != event.form.value.passwordVerif) {
```
subfuzion: `passwordVerif` => `passwordConfirm`

portal/client/src/app/auth/signup/signup.component.ts
```diff
@@ -29,16 +30,20 @@ export class SignupComponent implements OnInit {
...
-        this.message = "your password is not the same than your password confirmation"
+        this.messageError = "your password is not the same than your password confirmation"
         return
     }
     this.httpService.signup(event.form.value.username, event.form.value.password, event.form.value.email).subscribe(
       data => {
-        this.message = "Your account is created, you are going to receive an email. Confirmed you received it, before sign in"
+        this.message = "Your account is created, you are going to receive an email to validate your account"
```
subfuzion: Message isn't accurate if registration is disabled

freignat91: yes, it needs to add a way to know using api if registration is disabled or not

portal/client/src/app/auth/signup/signup.component.ts
freignat91: That's a variable name not seeing from outside

portal/client/src/app/auth/signup/signup.component.ts
freignat91: ok

portal/client/src/app/auth/signup/signup.component.ts
subfuzion: Yes, I'm just suggesting a better name.

portal/client/src/app/auth/signup/signup.component.ts
subfuzion: Yes, and if IRC, we set an environment variable for this when we run the cluster create command. I'm commenting from my phone right now and can check later, but you can also check `cli\command\cluster\create.go`

portal/client/src/app/auth/signup/signup.component.ts
freignat91: You are not asking me to change my variable names?

subfuzion:
@ndegory I agree that our compose files should be careful to identify constrains for where replicas should be placed, but we should also proactively ensure that workers don't accidentally run on our managers due to an inadvertent omission. See [here](https://docs.docker.com/engine/swarm/admin_guide/#run-manager-only-nodes) for draining manager nodes and ensure the scheduler doesn't assign any more workers to them as well.

portal/client/src/app/auth/signup/signup.component.ts
subfuzion: @freignat91 We do set the environment variable [here](https://github.com/appcelerator/amp/blob/master/cli/command/cluster/create.go#L52) and that's how amplifier also gets it. I don't want this to hold up the PR, so I'll open an issue so that we make sure to the feedback is correct for the current configuration.

platform/stacks/portal.stack.yml
```diff
@@ -9,17 +9,20 @@ networks:
...
       replicas: 1
       labels:
         io.amp.role: "infrastructure"
+      placement:
+        constraints:
+          - node.role == ${NODE_ROLE:-worker}
     environment:
       SERVICE_PORTS: "3333"
-      VIRTUAL_HOST: "http://ui.*,https://ui.*,http://cloud.*,http://local.*,https://cloud.*,https://local.*"
+      VIRTUAL_HOST: "http://portal.*, https://portal.*,http://ui.*,https://ui.*,http://cloud.*,http://local.*,https://cloud.*,https://local.*"
```
subfuzion: Can you remove the references to `portal` and `ui` ... it's our portal, but we want to access it at `cloud.*`. Thanks.

freignat91: ok

ndegory (approved):

subfuzion (approved):
LGTM

Issue #1075, opened 2017-04-25
BYOS feature


ndegory:
Bring Your Own Swarm
allow to deploy AMP on an existing cluster, already managed by another mean.

ndegory:
should be able to use the AMP CLI on a manager node and deploy the core services on the cluster, as well as the needed secrets (certificates, amplifier configuration).

Issue #773, opened 2017-03-04
InfraKit self healing robustness


ndegory:
Test and configure properly InfraKit so that the loss of a Swarm node (especially a manager, especially the leader) is recoverable.
One area to cover is how to manage the node registration: when a node disappears, it still is registered in the Swarm (as shown by `docker node ls`), the self healing process will just add new nodes to the list so we need to clean up the unreachable/inactive nodes.

ndegory:
when Infrakit is replacing a node (because configuration has been changed), it cleans the swarm status, the old node is removed from the list.
That's good enough for now, node disappearing because of a node issue should anyway trigger alerts, cleaning the node from the manager should be part of the procedure.

PR #1235, opened 2017-05-20
Bootstrap refactoring

### LINKED ISSUE #944: HA for deployment tool

### LINKED ISSUE #945: InfraKit container high resources usage
InfraKit is using a lot of resources of the manager nodes:
```
$ docker stats
CONTAINER           CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O           PIDS
[infrakit]        30.19%              880.5MiB / 3.674GiB   23.40%              8.48GB / 3.78GB     57.3kB / 2.01GB     60
```

In the container:
```
  PID  PPID USER     STAT   VSZ %VSZ CPU %CPU COMMAND
15274 15264 root     S     126m   3%   0  25% /usr/bin/terraform internal-plugin provider aws
15894 15890 root     S     124m   3%   0   0% /usr/bin/terraform apply
15890    49 root     S     121m   3%   0   0% terraform apply
   49    44 root     S    30248   1%   0   0% infrakit-instance-terraform --dir=/infrakit/terraform --log 5 
```

### LINKED ISSUE #973: Add datacenter HA to Swarm manager deployment
Instead of deploying in a single subnet (same datacenter), we should use 3 subnets and deploy a manager on each of them. The security group should be shared by all instances.
Terraform can help on this with the `count` property and an array of availability zone for each region.

### LINKED ISSUE #1074: Switch back to the InfraKit native AWS instance plugin
The promise of a generic instance plugin with Terraform is not meeting our expectations.
The extra layer with the generation of tf files, the application of the files with the Terraform binary and the discovery through the state file makes the process less stable. The different way to manage resource control permission on the different cloud providers is anyway a hindrance regarding the way we want to deploy clusters.
native instance plugins already exist for AWS and GCP, and are in progress for Azure and DigitalOcean, and seem now to get better traction than Terraform.
Moving away from Terraform means that the bootstrap will have to be done with a native tool instead of a bootstrap terraform file.

### LINKED ISSUE #1123: kill -HUP on dind Docker daemon sometimes kills the container
Happens on Travis builds and on an EC2 Ubuntu instance, but not on Mac and not on a Linux laptop.
To reproduce: add a kill -HUP $(cat /var/run/docker.pid) in the bootstrap/manager-init.tpl template.

### LINKED ISSUE #1150: Manager nodes bootstrap

### LINKED ISSUE #1183: System checks before bootstrap deployments are ignored
The system checks in the bootstrap script should exit the script in error when the checks are not successful.
Besides, the travis builds do not meet the script expectations, the IP VS modules are not all loaded.

ndegory:
Fix #1183 (System checks are ignored)
Fix #1123 (HUP signal on dind)
Fix #1150 (Manager nodes bootstrap)
Fix #1074 (Native AWS plugin)
Fix #973 (HA manager nodes)
Fix #945 (Infrakit high resource usage)
Fix #944 (HA deployment tools)
Ref #1075 (BYOS)
Ref #773 (Self healing)

Deployment for local swarm, local DinD cluster and AWS cluster can all be done from same script (deploy).
AWS deployment with the CLI not ready yet for prime time, TBC.


subfuzion (approved):
LGTM (although slightly smaller, more granular PRs are appreciated!) 👍

PR #1242, opened 2017-05-22
Post 0.11.0

ndegory:
- Updates from release/0.11
- VERSION bumped to 0.12.0-dev


bquenin (changes requested):
Looks good, we should probably also update the version on the project homepage.

ndegory:
good catch @bquenin , I forgot it was hardcoded.

bquenin (dismissed):
:+1:

bquenin (approved):
LGTM

PR #1257, opened 2017-05-25
Fix storage update

### LINKED ISSUE #1238: 'amp team member add' issue
we got an error, when adding the owner of an organization in a team of his organization

to reproduce it:
- log with an user: `amp login` (for instance freignat)
- create an organization: myorg (`amp org create`)
- create a team in this organization: myteam (`amp team create`)
- add the user (owner of the organization myorg) to the team myteam:

```
amp team member add -s localhost
[localhost:50101]
organization name: myorg
team name: myteam
member name: freignat
Error: proto: Marshal called with nil
```

By the way, the user seems well added in the team

bquenin:
#Fixes #1238 

Integration tests have been updated accordingly.

### DIFF
--- a/data/storage/etcd/store.go
+++ b/data/storage/etcd/store.go
@@ -173,6 +173,9 @@ func (s *etcd) Update(ctx context.Context, key string, uf storage.UpdateFunc, te
 		if err != nil {
 			return fmt.Errorf("Error updating object: %s", err)
 		}
+		if update == nil { // update function returned no error and nil data, no update to perform
+			return nil
+		}
 
 		data, err := proto.Marshal(update)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


JosephGJ (dismissed):
This does fix the issue of the error returning. However, the user who creates the org / team is added to it by default, so there should be an error returned if the user already exists in the org.

bquenin:
@JosephGJ This is actually debatable. When the request you make, in this case adding a user to the organization is fulfilled (since the user is already a member of this organization), you can consider the request has been performed successfully. However, this is mainly for UI reasons, when you add a member already part of the organization, so that it doesn't report an error.
We can discuss this over an issue ;)

JosephGJ (approved):

PR #1259, opened 2017-05-25
Change team name

### LINKED ISSUE #1120: Be able to update team name
UI/CLI needs an api verb to update an team name

bquenin:
Resolves #1120

Added a `ChangeTeamName` method to accounts API.

Updated integration tests in `tests/integration/accounts/accounts_test.go`


data/accounts/store.go
```diff
@@ -820,6 +820,40 @@ func (s *Store) ChangeTeamResourcePermissionLevel(ctx context.Context, organizat
...
+		if !ok {
+			return nil, fmt.Errorf("value is not the right type (expected Organization): %T", organization)
+		}
+
+		// Get team
+		team := organization.getTeam(teamName)
+		if team == nil {
+			return nil, TeamNotFound
+		}
+
```
JosephGJ: Need to check if `newName` is the name of an existing team in that org.

data/accounts/store.go
```diff
@@ -820,6 +820,40 @@ func (s *Store) ChangeTeamResourcePermissionLevel(ctx context.Context, organizat
...
+		// Update team name
+		team.Name = newName
```
JosephGJ, line 850: Need to check if `newName` is the name of an existing team in that org.

bquenin: Good catch!

JosephGJ (approved):

PR #1262, opened 2017-05-25
Adding a resource API

### LINKED ISSUE #1141: Needs a new api verb for team resources management for UI
Get all resources of an organization with their id, type and if exists a name

bquenin:
Resolves #1141 

Added a resources API in order to list available resources.
The method `ListResources` will only return resources belonging to the requesting user, based on credentials.
Added integrations tests here `tests/integration/resources/resources_test.go`.


freignat91 (approved):
LGTM, tested ok

PR #1265, opened 2017-05-26
Specialized swarm nodes

### LINKED ISSUE #949: Set Swarm node labels at node deployment
Labels can be set at the Docker engine level (in /etc/docker/daemon.json) or at the Swarm node level (docker node update).
Both can be used as constraint for service deployment, but it makes more sense to doit at the Swarm node level.

Should or could be used for:
- pet servers (for database clusters like elasticsearch)
- environments (dev, test, prod)
- organizations

The nodes are deployed by InfraKit, so the labels should be set:
- by InfraKit when building the Swarm cluster (amp cluster start)
- by InfraKit when adding or updating nodes (not yet implemented in the CLI)

### LINKED ISSUE #1121: local bootstrap deployment should scale down databases services
Deploying 3 nodes of etcd and elasticsearch can be an issue on systems with low resources.
The bootstrap should be able to deploy a single container of each service when deploying a local swarm.

### LINKED ISSUE #1218: Dedicated nodes for core services
we should be able to allocate nodes for services like elasticsearch, etcd or nats
That means:
- adding a placement constraint on each core service definition, based on node labels
- adding all the labels to the cluster
  - that means that for local deployments (dnd or single node) we should make sure that all labels are set

A list of predefined labels can be:
- amp.type.mq (nats)
- amp.type.kv (etcd)
- amp.type.search (elasticsearch)
- amp.type.api (amplifier)
- amp.type.metrics
- amp.type.route (haproxy)
- amp.type.core (all other core services)
- amp.type.user (for non core services)

This way we can get rid of the node.role == worker, now we got a better mechanism to orchestrate the services.

ndegory:
Fix #1218 (dedicated nodes)
Fix #1121 (scale down db services on local deployments)
Fix #949 (labels on swarm nodes)

The InfraKit configuration has been spread in several groups (optional): general purpose, cpu optimized, memory optimized, storage optimized and monitoring.
Based on the requested size of the cluster these groups are activated, and core services are distributed on them (service oriented labels are set on these groups to define which service should go on which group).
When the size of the cluster does not allow enough nodes, DB services (etcd and elasticsearch) are reduced to a single replica. To do that the stack file has been duplicated (ampmon-single|cluster.1.stack.yml).

Local deployments are setting all labels on the local swarm node.
DinD deployments are setting api and route labels on the manager, monitoring on the proxy node and all the other labels on the main worker group.

For convenience on AWS deployment, a tag is added at stack level to identify who has deployed the stack (Unix user name).


subfuzion (approved):
LGTM

Issue #1192, opened 2017-05-15
Self-signed certificate issue with UI using the gateway


freignat91:
When starting the UI, chrome allows to accept the self-signed certificate even not secure, but when the ui itself requests the amplifier-gateway chrome doesn't ask for validation and immediately refuses the request without even sending the first CORS http request.
see: 
https://www.sencha.com/forum/archive/index.php/t-185039.html?s=626c4e09a57d34151bad48ae5d81f8da

A possible workaround is to get the refused gateway url (in the console) and to put it in the new browser tab, then chrom ask for validation and the user can continue to use the UI.

however, it's not an acceptable solution, so we need the trusted certificate

PR #1280, opened 2017-05-29
Portal custom dashboard

### LINKED ISSUE #1270: Add customizable dashboard
As example:

image

### LINKED ISSUE #1271: Manage team resources assignment

### LINKED ISSUE #1272: Manage users/organizations authorizations in UI globally

freignat91:
Closes #1270 
Closes #1271 
Closes #1272 

PR #1304 have to be merged first 

Add portal custom dashboard with team sharing capabilities.

image


ndegory:
@freignat91 
I'm trying to access the UI through local.appcelerator.io, when I try to register the Submit button has no effect. Probably a certificate / gateway issue, but I'm not sure what is the recommended procedure to go on, can we add it to the documentation?

api/rpc/dashboard/dashboard.go
```diff
@@ -60,6 +60,9 @@ func (s *Server) List(ctx context.Context, in *ListRequest) (*ListReply, error)
 	if err != nil {
 		return nil, convertError(err)
 	}
+	for _, dashboard := range dashboards {
+		dashboard.Data = "" //remove data, data are loaded one by one using get
```
bquenin, line 64: @freignat91 This line is causing the tests to fail. Tests need to be updated to reflect this modification.

bquenin (changes requested):

ndegory (changes requested):
I tried to access the UI through local.appcelerator.io, when I try to register the Submit button has no effect. Probably a certificate / gateway issue, but I'm not sure what is the recommended procedure to go on, can we add it to the documentation?
It would be great also to catch this issue and notify about it directly in the UI, most user won't look in the documentation on how to use a UI...
Ref #1220 and #1192 

{FR}-> for the certificate error message, it should be the case, its a bug. The error is trapped and a specific message is sent with saying that the certificat is not there and indicating to read doc.
However for local usage no certificate should be better to simplify the deployment.

For the record, adding the certificate in my system keychain didn't help, Chrome still refused the certificate. I had to accept the certificate for local.appcelerator.io, open another tab and type https://gw.local.appcelerator.io, accept the certificate, and go back to the UI to signup.

When deploying a stack with the CLI, there's a default name. The UI should do the same, when loading a file, it should use the basename.

{FR} that's a server issue, the default should be on server side, not on UI , nor on CLI, in order that UI and CLI benefit of them for free

In the dashboard UI, the rounded box checkbox doesn't have any effect (it's not really mandatory, but if we keep the checkbox, we should make sure it works). The default text for the legends should be changed (stack legend instead of legend stack)

{FR}: yes, rounded checkbox is not yet implemented. i'll remove them for now

Histo period should be changed to period (or range). The text for the 5 min range is 10 min. The text for 2 years is 1 years.

{FR}; yes, Ok

api/rpc/dashboard/dashboard.go
freignat91: yes, to be fixed

bquenin (dismissed):

bquenin:
> When deploying a stack with the CLI, there's a default name. The UI should do the same, when loading a file, it should use the basename.
> 
> {FR} that's a server issue, the default should be on server side, not on UI , nor on CLI, in order that UI and CLI benefit of them for free

It's not a server feature, the server doesn't infer names (and it cannot anyway). It's up to the client to pick his stack name.

{FR}: I don't agree. It's better to have only one place to compute defaults, otherwise  CLI and UI have to synchronize all the defaults of all verbs, coded separately . It's source of issues like this one and here it's simple, we have more complicated defaults in our API.
The server should have all the information to compute defaults, in this case, if the default is the stack file name, then, it should have the stack file name.
This way even direct REST API calls (without UI, nor CLI) will benefit of the defaults computation.

ndegory (dismissed):

ndegory (approved):

ndegory (commented):
@freignat91 please add .DS_Store in your global gitignore.

Issue #1283, opened 2017-05-30
Gateway replicas on cloud cluster


freignat91:
The amplifier-gateway use in global mde with several replicas on cloud cluster failed to use amplifier.
Reduce it to only one replicas makes it work correctly, but concidering that locally it works well with several replicas, there is no reasons that it doesn't work this way on cloud.
to be dig...

PR #1284, opened 2017-05-30
update gtw service setting

freignat91:
see: #1283 

Several gtw replicas on cluster make gtw unable to use amplifier
Back to one only, for now, time to dig more on this issue


freignat91:
Found better solution

PR #1295, opened 2017-06-01
Add button on verification email

### LINKED ISSUE #1289: Restore HTML buttons in verification emails

freignat91:
Closes #1289

Add button in the verification email which allows users to just click on it to validate their email instead to execute a CLI command.
The CLI command stays possible.

# test

- amp -s localhost cluster create -t local --registration=email --notifications=true
- amp -s localhost user signup
got the email and clic on the button and got a page saying your emai has been successfullycerified.
Then yo can clic on another button to login

or use the cli to login copying the command in the mail


api/rpc/account/account.go
```diff
@@ -121,12 +120,13 @@ func (s *Server) SignUp(ctx context.Context, in *SignUpRequest) (*empty.Empty, e
...
 func (s *Server) Verify(ctx context.Context, in *VerificationRequest) (*empty.Empty, error) {
 	// Validate the token
+	log.Printf("verify token=%s\n", in.Token)
 	claims, err := s.Tokens.ValidateToken(in.Token, auth.TokenTypeVerification)
 	if err != nil {
 		return nil, accounts.InvalidToken
 	}
-	if err := s.Accounts.VerifyUser(ctx, claims.AccountName); err != nil {
-		return nil, convertError(err)
+	if errv := s.Accounts.VerifyUser(ctx, claims.AccountName); err != nil {
```
bquenin: You don't need to rename this error, it's scoped. Plus the test has not been changed.

bquenin (changes requested):

api/rpc/account/account.go
freignat91: it's to avoid warning on my ide

bquenin: Still it's a bug, the test is wrong

freignat91: fixed

bquenin (approved):

PR #1297, opened 2017-06-01
Fix SignUp issue

### LINKED ISSUE #1291: amp signup: a not validated user is still created on mailer error
At signup with option validation by email, if the SendGrid key doesn't exist for instance, the user is still created with a status "Not verified", but then there is not way to remove or verify him (except super-user, not yet merged)

freignat91:
Closes #1291 

Fix the SignUp error which didn't delete the user when internal error happened (as email not working) 

# test

- Do not install the SendGrid key 
- amp -s localhost cluster create -t local --registration=email --notifications=true
- amp -s localhost user signUp -> error SendGrid API key is empty
- amp -s localhost user signUp -> error SendGrid API key is empty

needed to signup two times, in order to verify the user as well been deleted at the first signUp error

### DIFF
--- a/data/accounts/store.go
+++ b/data/accounts/store.go
@@ -344,8 +344,25 @@ func (s *Store) ListUsers(ctx context.Context) ([]*User, error) {
 	return users, nil
 }
 
+// DeleteNotVerifiedUser deletes the user by name only if it's not verified
+func (s *Store) DeleteNotVerifiedUser(ctx context.Context, name string) error {
+
+	//Get user to verify it is well not verified
+	user, err := s.GetUser(ctx, name)
+	if err != nil {
+		return err
+	}
+	if user != nil && !user.IsVerified {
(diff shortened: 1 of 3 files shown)


data/accounts/store.go
```diff
@@ -23,6 +23,8 @@ const organizationsRootKey = "organizations"
 const superUser = "su"
 const superOrganization = "so"
 
+const SignUpErrorContextValue = "SignUpError"
```
bquenin: Dead code

bquenin (changes requested):

data/accounts/store.go
freignat91: oui, effectivement

freignat91: done

bquenin (dismissed):

bquenin (approved):
LGTM

PR #1298, opened 2017-06-01
Using synchronous connections to amplifier

### LINKED ISSUE #1283: Gateway replicas on cloud cluster

bquenin:
Fixes #1283


freignat91 (approved):
LGTM, tested ok

PR #1318, opened 2017-06-05
Portal fixes and UX improvments

freignat91:
Fix portal issues and improve UX :
- better information displayed for list of stacks, list of services, list of containers, add auto refresh
- add separate list of users for organization with capability to change the user role (member/owner)
- improve navigation, fixe navigation error, make broswer page back work anycase
- most of the UI pages are accessible in separate way using their own url
- fixe list refresh missing error after creation (user, organizations)
- add tooltip for side bar in condensed mode
- improve default dashboard
- add error messages in some place where they missing
- and a lot of other little changes


bquenin (approved):
LGTM

PR #1326, opened 2017-06-05
Default tag value for amp cluster create

neha-viswanathan:
default tag value for `amp cluster create` is read from the configuration.

```
    $ amp cluster create --help
    
    Usage:  amp cluster create [OPTIONS]

    Create an amp cluster

    Options:
      -h, --help                  Print usage
      -m, --managers int          Intial number of manager nodes (default 3)
          --name string           Cluster Label
      -n, --notifications         Enable/disable server notifications (default is 'false')
          --provider string       Cluster provider (default "local")
      -r, --registration string   Specify the registration policy (possible values are 'none' or 'email') (default "none")
      -s, --server string         Specify server (host:port)
      -t, --tag string            Specify tag for cluster images (use 'local' for development) (default "v0.12.0-dev")
      -w, --workers int           Initial number of worker nodes (default 2)
```

### DIFF
--- a/cli/command/cluster/create.go
+++ b/cli/command/cluster/create.go
@@ -24,8 +24,8 @@ func NewCreateCommand(c cli.Interface) *cobra.Command {
 	flags.IntVarP(&opts.managers, "managers", "m", 3, "Intial number of manager nodes")
 	flags.StringVar(&opts.provider, "provider", "local", "Cluster provider")
 	flags.StringVar(&opts.name, "name", "", "Cluster Label")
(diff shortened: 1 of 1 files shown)


subfuzion (approved):
Thanks @neha-viswanathan !

neha-viswanathan:
👍

ndegory:
@neha-viswanathan @subfuzion 
that means that during a development cycle `amp cluster create`, without the -t option, fails, because there's no tag 0.12.0-dev on Docker hub or in the local Docker store.
Is that what you wanted?

PR #1335, opened 2017-06-07
Portal display server version

### LINKED ISSUE #1330: Portal UI version should be synced with the project
Portal UI should use the VERSION file instead of hardcoding in sidebar.component.html

freignat91:
Closes #1330 

## test

- display UI and see server version on the bottom side bar
This become red with additional portal version if the portal version doesn't match


subfuzion (approved):
LGTM

PR #1337, opened 2017-06-08
Remove-portal

freignat91:
Portal has been moved to repository: https://github.com/appcelerator/portal

Remove portal for amp repository

These vendor was used by portal to check if they can be removed:
- github.com/gorilla/mux
- github.com/gorilla/context
- github.com/phyber/negroni-gzip
- github.com/urfave/negroni


platform/stacks/portal.stack.yml
```diff
@@ -9,7 +9,7 @@ networks:
 services:
 
   portal:
-    image: appcelerator/portal:${TAG:-0.11.0}
+    image: appcelerator/portal:latest
```
bquenin: Shouldn't we move the stack file in the portal project as well?

bquenin (commented):
LGTM, just one comment

platform/stacks/portal.stack.yml
freignat91: I don't think so, we need a way to launch the portal service for amp cluster command

freignat91: but we can replace 'latest' by the current amp version

bquenin: ok, then yes, maybe keep the amp tagged version

freignat91: ok, i gonna change it

bquenin (approved):
LGTM

PR #1358, opened 2017-06-12
Using service for integration tests

### LINKED ISSUE #1359: Travis builds are unstable
The reason is that sometimes docker container attached to an overlay network is unable to reach other services whereas a service task seems to be able to. I've searched for an open issue in moby but I wasn't able to find one.
The workaround is to use a docker service instead of a docker container to run integration tests.

bquenin:
Resolves #1359 

Using a docker service instead of a docker container to run integration tests.

## How to test
Make sure the travis build passes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ cleanall: clean cleanall-deps
 DOCKER_CMD ?= "docker"
 
 build-base: install-deps protoc build-server build-gateway build-beat build-agent build-bootstrap build-monit
-build: build-base buildall-cli
+build: build-base build-cli
 
 # =============================================================================
 # BUILD CLI (`amp`)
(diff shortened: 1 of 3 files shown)


neha-viswanathan (approved):

Makefile
```diff
@@ -108,7 +108,7 @@ cleanall: clean cleanall-deps
 DOCKER_CMD ?= "docker"
 
 build-base: install-deps protoc build-server build-gateway build-beat build-agent build-bootstrap build-monit
-build: build-base buildall-cli
+build: build-base build-cli
```
subfuzion, line 111: This isn't a good idea right now -- and anyway this PR is not the right place to change how we use the rules -- but you can optimize the CI build by calling the specific rules you want so time isn't being wasted cross-compiling for all OS target.s

bquenin: @subfuzion you're right, it was to speedup the build a little. Let me rework that in another PR.

PR #1365, opened 2017-06-13
Set default portal tag to latest

freignat91:
Considering we moved portal in a separated repository, we can't anymore use the tag 
${TAG:-0.11.0}  for portal, because the portal image is not build in amp build and so the tag local does't exist (when local build the TAG=local)
Move to tag 'latest'.
Versionning of portal to be managed in future pr


bquenin (approved):
LGTM

PR #1378, opened 2017-06-14
Initial commit of aws plugin

subfuzion:
## Description

Initial commit of work in progress. See [here](https://github.com/appcelerator/amp/wiki/AMP-Clusters) for design documentation.

### Options

The plugin allows you to provide all the parameters that are supported by the [Docker for AWS CloudFormation template](https://docs.docker.com/docker-for-aws/#configuration-options), including:

 * KeyName
 * EnableCloudWatchLogs
 * ManagerSize
 * ManagerInstanceType
 * ManagerDiskSize
 * ManagerDiskType
 * ClusterSize
 * InstanceType
 * WorkerDiskSize
 * WorkerDiskType

The format for adding a parameter (`-p | --parameter`) is more compact than the format used by the AWS CLI. For example, instead of:

    --parameters ParameterKey=ManagerInstanceType,ParameterValue=t2.medium ...

use this format instead:

    --parameter ManagerInstanceType=t2.medium ...

And for updates, instead of:

    --parameters ParameterKey=ManagerInstanceType,UsePreviousValue=true ...

just use this instead:

    --parameter ManagerInstanceType   # no assigned value assumes UsePreviousValue=true ...

## Prerequisites

* Ensure you have a keypair for the region you want to use to create a stack using the plugin. I recommend you upload your own public key and give it the same name for each region you want to test (see [keypair docs](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html#how-to-generate-your-own-key-and-import-it-to-aws)).

* Ensure your AWS credentials are configured (see [credentials docs](http://docs.aws.amazon.com/cli/latest/userguide/cli-config-files.html)).

### Sync Operations

You can block until a command completes using the `-s | --sync` option. For example:

    $ docker run -it --rm -v ~/.aws:/root/.aws appcelerator/amp-aws-plugin update --region us-west-2 --stackname tony-amp-10 -p KeyName=tony-amp-dev -p ClusterSize=4 --sync

## Verification

**NOTE:** I've provided a lot of detail here for the benefit of everyone on the team who might want to look while this is under development; most of you won't care about ssh tunnels, etc, and will interact with the swarm via the AMP CLI when this is finished.

Please ensure that any stacks you create for testing have your name in it using the following pattern:

    name-test-foo

Ex:

    tony-test-amp-awsplugin

This makes it a bit easier to identify test resources that can be deleted without guilt in case you forget to remove them. Please do try to remember to remove on your own after testing.

From the `cluster/plugin/aws` directory, run the following:

    $ make compiler
    $ make
    $ export REGION=us-west-2 # or any region where you have a valid keypair
    $ export KEYNAME=tony-amp-dev
    $ export STACKNAME=tony-amp-1
    $ export ONFAILURE=DO_NOTHING
    $ docker run -it --rm -v ~/.aws:/root/.aws appcelerator/amp-aws-plugin init \
        --region $REGION \
        --stackname $STACKNAME \
        --onfailure $ONFAILURE \
        --parameter KeyName=$KEYNAME \
        --parameter ...

Verify output similar to the following:

```
$ docker run -it --rm -v ~/.aws:/root/.aws appcelerator/amp-aws-plugin init --region us-west-2 --name tony-amp-7
2017/06/14 20:31:23 {
  StackId: "arn:aws:cloudformation:us-west-2:654814900965:stack/tony-amp-7/6d88d410-5140-11e7-8af9-503acbd4dc29"
}
2017/06/14 20:31:23 2010-05-15
```

Also verify the stack has been created in the AWS console. The URL looks like this (updated the region value as appropriate):

https://us-west-2.console.aws.amazon.com/cloudformation/home?region=us-west-2#/stacks?filter=active&tab=resources

### SSH to an instance

On the page opened by the previous link, you can click on your stack, then click on "Outputs", then click on the "Managers" link to the manager instances that were created. From there, you can obtain the public IP and DNS names created for each manager. Using the ssh key associated with your AWS keypair, you can ssh into any of these instances.

For example, if your private key is called `id_rsa` and the IP address of a manager node is `34.210.103.37`, then you can ssh into it like this:

    $ ssh -i ~/.ssh/id_rsa docker@ 34.210.103.37

You can then run docker swarm commands, like `docker node ls`.

Once you are connected to a manager node you can then ssh to a worker node using the node name as reported by `docker node ls`. Make sure you have ssh agent forwarding set up from your host system to ensure your keys propagate so that your private key doesn't need to be installed on each node you ssh from (see [docs](https://docs.docker.com/docker-for-aws/deploy/#using-ssh-agent-forwarding)).

    # can only ssh to a worker from a manager node
    $ ssh <worker>

### SSH Tunnel

If you don't need to ssh to a work and just want to connect to a manager for running docker commands, it can be more convenient to open an ssh tunnel. You can do this as follows:

    $ ssh -i ~/.ssh/id_rsa -NL 127.0.0.1:2374:/var/run/docker.sock docker@34.210.103.37

Or run the command and put it in the background -- just don't forget to kill it when done to release the port.

    $ ssh -i ~/.ssh/id_rsa -NL 127.0.0.1:2374:/var/run/docker.sock docker@34.210.103.37 &

Then you can specify the docker host to connect to using that port for the tunnel to run docker commands. For example:

    $ docker -H localhost:2374 node ls

Finally, you can set `DOCKER_HOST` so you don't have to use the `-H` option each time:

    $ export DOCKER_HOST=localhost:2374
    $ docker node ls

### Update Stack

Update an existing stack with the `update` command:

    $ docker run -it --rm -v ~/.aws:/root/.aws appcelerator/amp-aws-plugin update \
        --region $REGION \
        --stackname $STACKNAME \
        -p KeyName=$KEYNAME \
        -p ClusterSize=4 # new value! \
        --sync # block until done!

### Destroy Stack

Execute the `destroy` command with the stackname and region used for the `init` command:

    $ docker run -it --rm -v ~/.aws:/root/.aws appcelerator/amp-aws-plugin destroy \
        --region $REGION \
        --stackname $STACKNAME

## Tests

    $ KEYNAME=<aws-keypair-name> REGION=<aws-region> make test

Tests now automatically destroy the stacks they create.

### DIFF
--- a/cluster/plugin/aws/Dockerfile
+++ b/cluster/plugin/aws/Dockerfile
@@ -1,0 +1,4 @@
+FROM alpine
+RUN apk add --no-cache ca-certificates
+COPY aws.alpine /usr/local/bin/aws
+ENTRYPOINT [ "aws" ]
--- a/cluster/plugin/aws/Dockerfile.compiler
+++ b/cluster/plugin/aws/Dockerfile.compiler
@@ -1,0 +1,6 @@
+FROM golang:1.8-alpine
+RUN apk add --no-cache git
+RUN go get -u github.com/aws/aws-sdk-go/...
+RUN go get -u github.com/spf13/cobra/cobra
+
+
--- a/cluster/plugin/aws/Dockerfile.compiler
+++ b/cluster/plugin/aws/Dockerfile.compiler
@@ -1,6 +1,7 @@
 FROM golang:1.8-alpine
+
 RUN apk add --no-cache git
 RUN go get -u github.com/aws/aws-sdk-go/...
 RUN go get -u github.com/spf13/cobra/cobra
 
-
+ENTRYPOINT [ "go" ]
--- a/cluster/plugin/aws/Dockerfile.compiler
+++ b/cluster/plugin/aws/Dockerfile.compiler
@@ -3,5 +3,6 @@ FROM golang:1.8-alpine
 RUN apk add --no-cache git
 RUN go get -u github.com/aws/aws-sdk-go/...
 RUN go get -u github.com/spf13/cobra/cobra
+RUN go get -u github.com/satori/go.uuid
 
 ENTRYPOINT [ "go" ]
--- a/cluster/plugin/aws/Makefile
+++ b/cluster/plugin/aws/Makefile
@@ -1,0 +1,17 @@
+
+build: aws.alpine
+	docker build -t appcelerator/amp-aws-plugin .
+
+aws.alpine: main.go
+	docker run -it --rm -v $${PWD}:/go/src/github.com/appcelerator/amp/cluster/plugin/aws \
+		-w /go/src/github.com/appcelerator/amp/cluster/plugin/aws \
+		awssdk go build -o aws.alpine .
+
+.PHONY: compiler
+compiler:
+	docker build -f Dockerfile.compiler -t awssdk .
+
+.PHONY: clean
+clean:
+	rm -f aws.alpine
+
--- a/cluster/plugin/aws/Makefile
+++ b/cluster/plugin/aws/Makefile
@@ -6,13 +6,23 @@ build: aws.alpine
 aws.alpine: $(SRC)
 	docker run -it --rm -v $${PWD}:/go/src/github.com/appcelerator/amp/cluster/plugin/aws \
 		-w /go/src/github.com/appcelerator/amp/cluster/plugin/aws \
-		awssdk go build -o aws.alpine cmd/main.go
+		go:aws build -o aws.alpine cmd/main.go
 
 .PHONY: compiler
 compiler:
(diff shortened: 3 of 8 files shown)


bquenin (commented):
Look good so far. Tested on linux.

freignat91 (dismissed):
LGTM,
tested ok on OS/x
I just would used t2-large instead of t12=micro as EC2 instances

subfuzion:
t2.micro is just the default for instances in the Docker for AWS CloudFormation template. You have to override using the InstanceType (for workers) and ManagerInstanceType parameters. However, I didn't have support for passing through custom parameters until my last commit. Try it now ... you can tune all the template parameters (see the updated description above).

bquenin (approved):
LGTM. Tested ok.

PR #1386, opened 2017-06-15
Fix node labels retrieval

freignat91:
Api verb node.GetNodes didn't retrieve the labels. Fix it.
Testable only on UI


bquenin (approved):
LGTM

PR #1388, opened 2017-06-15
Using client keepalive between the gateway and amplifier

bquenin:
Enabled a keep-alive between the `amplifier-gateway` and `amplifier`. The keep-alive is client driven, i.e. the client sends ping requests to the server, which in turn responds.
The keep-alive frequency is set to every 5 minutes, with a 20 seconds timeout for the response.

This is supposed to improve reliability of the connection between `amplifier-gateway` and `amplifier` based on the conclusions of this issue: https://stackoverflow.com/questions/43311004/docker-services-stops-communicating-after-some-time/43528819


freignat91 (approved):
LGTM

ndegory:
nice. we'll probably need it also on etcd connection in amplifier, and nats connection in the agent.

PR #1390, opened 2017-06-15
Use nats-streaming durable queue groups

bquenin:
Spread the load among agents by using queue groups and restart from last message in case of a crash using durable.


freignat91 (approved):
LGTM

PR #1395, opened 2017-06-16
Update io-net agent computation

freignat91:
Improve the precision of the IO and NET agent Metrics agent computation

## test

- ampmake rebuild-agent
- amp -s localhost cluster create -t local
- amp -s localhost stats -i  -> net and io are more accurate


bquenin (approved):
LGTM. Tested ok.

PR #1403, opened 2017-06-20
Ensure required env vars set for test`

subfuzion:
## Verification

    make test-amp-aws
    # should exit with error due to missing KEYNAME or REGION env var

    KEYNAME=keyname REGION=region make test-amp-aws
    # should succeed with valid keyname and region values


bquenin (approved):
LGTM. Tested ok

PR #1410, opened 2017-06-21
Update docker package name for Travis

bquenin:
Updates the docker package name for Travis builds.

## How to test
Make sure Travis successfully updates docker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ language: bash
 install:
   - sudo sysctl -w vm.max_map_count=262144
   - sudo apt-get update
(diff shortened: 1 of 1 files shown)


ndegory:
what was the issue?

bquenin:
https://travis-ci.org/appcelerator/amp/builds/245511986#L274-L284

ndegory:
from what I see in the build logs, now by default docker is already installed with version 17.03.1 (latest stable, for a few days at least).
we're supposed to be compatible with this version, why should we upgrade to the edge version?

bquenin:
because we used to build against 17.05 until now , so it's just a matter of continuity, I don't mind using 17.03 though

ndegory (approved):
LGTM
we'll switch to stable channel once 17.06 is released

PR #1413, opened 2017-06-22
Allow --template parameter for aws plugin

subfuzion:
Override the default Docker for AWS cloudformation template by specifying a url to a template with the `-t | --template` option.

### DIFF
--- a/cluster/plugin/aws/plugin.go
+++ b/cluster/plugin/aws/plugin.go
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	templateURL = "https://editions-us-east-1.s3.amazonaws.com/aws/edge/Docker.tmpl"
+	DefaultTemplateURL = "https://editions-us-east-1.s3.amazonaws.com/aws/edge/Docker.tmpl"
 )
 
 // RequestOptions stores raw request input options before transformation into a AWS SDK specific
@@ -27,6 +27,8 @@ type RequestOptions struct {
 	StackName string
 
 	Sync bool
+
+	TemplateURL string
 }
 
 func parseParam(s string) *cf.Parameter {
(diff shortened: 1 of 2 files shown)


cluster/plugin/aws/plugin.go
```diff
@@ -61,7 +63,7 @@ func CreateStack(ctx context.Context, svc *cf.CloudFormation, opts *RequestOptio
 		},
 		OnFailure:        aws.String(opts.OnFailure),
 		Parameters:       toParameters(opts.Params),
-		TemplateURL:      aws.String(templateURL),
+		TemplateURL:      aws.String(DefaultTemplateURL),
```
ndegory: shouldn't it be opts.TemplateURL instead of DefaultTemplateURL?

cluster/plugin/aws/plugin.go
```diff
@@ -75,7 +77,7 @@ func UpdateStack(ctx context.Context, svc *cf.CloudFormation, opts *RequestOptio
 			aws.String("CAPABILITY_IAM"),
 		},
 		Parameters:       toParameters(opts.Params),
-		TemplateURL:      aws.String(templateURL),
+		TemplateURL:      aws.String(DefaultTemplateURL),
```
ndegory: ditto

ndegory (changes requested):

cluster/plugin/aws/plugin.go
subfuzion: Yup to both ... I did a search/replace on templateURL to rename, but didn't switch these over to use the option now. Thanks for catching.

ndegory (dismissed):

ndegory (approved):

PR #1414, opened 2017-06-22
Cloudformation template customized for AMP

ndegory:
AWS Cloudformation templates samples for BYOS
- option for Docker channel
- manager group, core services group, user services group, monitoring group (1 node)
- label the swarm nodes for service scheduling
- options for block storage

Whenever an instance is created (this will happen every time an autoscaling group doesn't reach its desired state) the instance userdata will:
- install Docker
- configure Docker (remote API, metrics endpoint, registries)
- if it's a manager, will look for a leader, or establish who's the leader candidate
- initialize or join the swarm
- label the nodes
- install plugins
- create overlay networks
- basic smoke tests (swarm membership)

see examples/cluster/README.md for instructions
The template is also available on https://s3.amazonaws.com/io-amp-binaries/templates/latest/aws-swarm-asg.yml so it can be used even if you don't checkout this branch.

## Verification
Use the aws cli or the console to deploy a swarm cluster.
It is compatible with regions us-east-1, us-east-2, us-west-2, eu-west-1 and ap-southeast-2.

Alternatively it can be used through the aws plugin of AMP:
```docker run -it --rm -v ~/.aws:/root/.aws appcelerator/amp-aws:latest init --region us-west-2 --stackname STACKNAME --parameter KeyName=KEYNAME --parameter ConfigurationURL=https://raw.githubusercontent.com/appcelerator/amp/cloudformation-for-amp/examples/clusters --parameter OverlayNetworks=ampnet --parameter DockerChannel=edge --template https://s3.amazonaws.com/io-amp-binaries/templates/latest/aws-swarm-asg.yml```

For reference, see the list of parameters below.

| Parameter | Description | Default Value | Example |
| --------- | ----------- | ------------- | ------- |
| KeyName   | Name of an existing EC2 KeyPair to enable SSH access to the instances | - | YOURNAME-REGION |
| ManagerSize | Number of manager nodes, should be 1, 3 or 5 | 3 | |
| CoreWorkerSize | Number of worker nodes for core services | 3 | |
| UserWorkerSize | Number of worker nodes for user services | 3 | |
| MetricsWorkerSize | Number of worker nodes for metrics services | 1 | |
| LinuxDistribution | AMI OS, Debian or Ubuntu | Ubuntu | Debian |
| ManagerInstanceType | Instance type for the manager nodes. Must be a valid EC2 HVM instance type | t2.small | m4.large |
| CoreInstanceType | Instance type for the core worker nodes. Must be a valid EC2 HVM instance type | m4.large | c4.large |
| UserInstanceType | Instance type for the user worker nodes. Must be a valid EC2 HVM instance type | t2.medium | m4.large |
| MetricsInstanceType | Instance type for the metrics worker nodes. Must be a valid EC2 HVM instance type | t2.large | m4.large |
| DrainManager | Should we drain services from the manager nodes? | false | true |
| ConfigurationURL | URL for manager and worker userdata configuration | https://raw.githubusercontent.com/appcelerator/amp/master/examples/clusters | |
| AufsVolumeSize | Size in GB of the EBS for the /var/lib/docker FS | 26 | 100 |
| OverlayNetworks | name of overlay networks that should be created once swarm is initialized | ampnet | public storage search mq |
| DockerChannel | channel for Docker installation | stable | edge |
| DockerPlugins | space separated list of plugins to install | | rexray/ebs |

### DIFF
--- a/examples/clusters/userdata-aws-worker
+++ b/examples/clusters/userdata-aws-worker
@@ -1,0 +1,196 @@
+#!/bin/bash
+LABELS=${LABELS:-amp.type.user=true amp.type.core=true amp.type.search=true amp.type.mq=true amp.type.kv=true amp.type.metrics=true}
+CHANNEL=${CHANNEL:-stable}
+#PLUGINS=
+REGION=${REGION:-us-west-2}
+STACK_NAME=${STACK_NAME:-unset}
+VPC_ID=${VPC_ID:-unset}
+#DOCKER_DEVICE=/dev/sdl
+SYSTEMD_DOCKER_OVERRIDE=/etc/systemd/system/docker.service.d/docker.conf
+SYSV_DOCKER_DEFAULT=/etc/default/docker
+#LEADER=
+
+_init_system(){
+  systemctl --version >/dev/null 2>&1 && echo systemd && return
+  [[ `/sbin/init --version` =~ upstart ]] && echo upstart && return
+  echo sysv
+}
+
+_install_docker(){
+  local _release=$(lsb_release -is)
+  local _host
+  case $CHANNEL in
+  stable) _host="get.docker.com" ;;
+  edge|beta|test) _host="test.docker.com" ;;
+  experimental) _host="experimental.docker.com" ;;
+  *) return 1 ;;
+  esac
+  echo "installing Docker from $_host" >&2
+  wget -qO- "https://$_host/" | sh || return 1
+  [[ "x$_release" = "xUbuntu" ]] && usermod -G docker ubuntu
+  [[ "x$_release" = "xDebian" ]] && usermod -G docker admin 
+  if [[ $(_init_system) = "systemd" ]]; then
+    systemctl enable docker.service
+    systemctl start docker.service
+  else
+    chkconfig docker on
+    service docker start
+  fi
+  docker version >&2
+}
+
+_install_plugins(){
+  local plugin
+  local alias
+  for plugin in $PLUGINS; do
+    alias=${plugin#store/}
+    alias=${alias%:*}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by subfuzion — COMMENTED** [review]
@ndegory This looks very good so far. I'm going to test with the updated CLI plugin support for the aws plugin to verify. Aside from the brief `sanity_check` functions you have, I'd like to see some kind of post swarm setup smoke test (or am I missing that?). Thanks.

**Comment by ndegory** [comment]
there's no post swarm smoke check, and the cloudformation stack gets the CREATE_COMPLETE status before the instances are even booting, so it's not conditioned on the status of the swarm.
I can add a wait condition to make sure the stack status reflects the real status of the swarm.
I've been reluctant to do that, since it adds more adherence to AWS (we have to install the cfn tools to send signals once the userdata (and the smoke tests) are done).
Or, I can add a simple check: if the smoke tests are failing, the instance terminates.
To avoid deadlocks, I think the smoke tests should be limited to a simple swarm membership check.

File: examples/clusters/userdata-aws-worker [hunk 1]
```diff
@@ -0,0 +1,208 @@
...
+    chkconfig docker on
+    service docker start
+  fi
+  docker version >&2
+}
+
+_install_plugins(){
+  local plugin
+  local alias
+  for plugin in $PLUGINS; do
```
subfuzion, line 45: Just a placeholder or are there any specific plugin requirements?

subfuzion: oh nvm ... rexray, or potentially other storage plugins, etc.

subfuzion (approved):
Very nice 👍

PR #1418, opened 2017-06-22
don't validate the etcd connection when it fails

### LINKED ISSUE #1417: amplifier says it's connected to etcd even when the connection fails
To reproduce: place etcd on a different overlay network than amplifier. When the connection is attempted, we have two messages: connection successful, and connection error.

ndegory:
Fix #1417


bquenin (changes requested):
Good catch! However, I think you can just insert a:
```
if err != nil {
    return err
}
```
right after the call to `clientv3.New`

bquenin (approved):
LGTM

PR #1421, opened 2017-06-23
amp cli cluster plugin

### LINKED ISSUE #1303: Cluster deployment on AWS with CLI
A user should be able to deploy a cluster on AWS, only prerequisites are Docker, the amp CLI and AWS credentials.

subfuzion:
Closes #1303

See documentation [here](https://github.com/appcelerator/amp/wiki/AMP-Clusters) for more details.

Using `amp cluster create --provider aws` supports creating an AWS cluster using the AMP user's AWS credentials. The AWS credentials are automatically read from the user's home directory (`$HOME/.aws`) (a future update will support command line options as well, like the AWS CLI).

### Using the Docker for AWS template

```
$ REGION=us-west-2                  # or any desired region where your keys are deployed
$ KEYNAME=tony-amp-dev              # use your own keypair name
$ STACKNAME=tony-tmp-stack1         # customize for yourself
$ CLUSTERSIZE=2                     # number of workers
$ MANAGERSIZE=1                     # number of managers

$ amp cluster create --provider aws \
   --aws-region $REGION \
   --aws-stackname $STACKNAME \
   --aws-parameter KeyName=$KEYNAME \
   --aws-parameter ClusterSize=$CLUSTERSIZE \
   --aws-parameter ManagerSize=$MANAGERSIZE \
   --aws-sync

# as one line (for copy/paste):
$ amp cluster create --provider aws --aws-region $REGION --aws-stackname $STACKNAME --aws-parameter KeyName=$KEYNAME --aws-parameter ClusterSize=$CLUSTERSIZE --aws-parameter ManagerSize=$MANAGERSIZE --aws-sync
```

### Using the custom AMP CloudFormation template

```
$ REGION=us-west-2                  # or any desired region where your keys are deployed
$ KEYNAME=tony-amp-dev              # use your own keypair name
$ STACKNAME=tony-tmp-stack1         # customize for yourself
$ TEMPLATE=https://s3.amazonaws.com/io-amp-binaries/templates/latest/aws-swarm-asg.yml
$ CONFIGURL=https://raw.githubusercontent.com/appcelerator/amp/master/examples/clusters

$ amp cluster create --provider aws \
   --aws-region $REGION \
   --aws-stackname $STACKNAME \
   --aws-template $TEMPLATE \
   --aws-parameter KeyName=$KEYNAME \
   --aws-parameter OverlayNetworks=ampnet \
   --aws-parameter DockerChannel=edge \
   --aws-parameter ConfigurationURL=$CONFIGURL \
   --aws-sync

# as one line (for copy/paste):
$ amp cluster create --provider aws --aws-region $REGION --aws-stackname $STACKNAME --aws-template $TEMPLATE --aws-parameter KeyName=$KEYNAME --aws-parameter OverlayNetworks=ampnet --aws-parameter DockerChannel=edge --aws-parameter ConfigurationURL=$CONFIGURL --aws-sync

```

### DIFF
--- a/cli/command/cluster/create.go
+++ b/cli/command/cluster/create.go
@@ -23,20 +23,21 @@ func NewCreateCommand(c cli.Interface) *cobra.Command {
 	}
 
 	flags := cmd.Flags()
-	flags.IntVarP(&opts.workers, "workers", "w", 2, "Initial number of worker nodes")
 	flags.IntVarP(&opts.managers, "managers", "m", 3, "Initial number of manager nodes")
-	flags.StringVar(&opts.provider, "provider", "local", "Cluster provider")
 	flags.StringVar(&opts.name, "name", "", "Cluster Label")
-	flags.StringVarP(&opts.tag, "tag", "t", c.Version(), "Specify tag for cluster images (use 'local' for development)")
-	flags.StringVarP(&opts.registration, "registration", "r", configuration.RegistrationNone, "Specify the registration policy (possible values are 'none' or 'email')")
 	flags.BoolVarP(&opts.notifications, "notifications", "n", false, "Enable/disable server notifications (default is 'false')")
+	flags.StringVar(&opts.provider, "provider", "local", "Cluster provider")
+	flags.StringVarP(&opts.registration, "registration", "r", configuration.RegistrationNone, "Specify the registration policy (possible values are 'none' or 'email')")
+	flags.StringVarP(&opts.tag, "tag", "t", c.Version(), "Specify tag for cluster images (use 'local' for development)")
+	flags.IntVarP(&opts.workers, "workers", "w", 2, "Initial number of worker nodes")
 
 	// aws options
-	flags.String("aws-region", "", "The region to use when launching the instance")
-	flags.String("aws-stackname", "", "The name of the AWS stack that will be created")
 	flags.String("aws-onfailure", "", "'DO_NOTHING', 'ROLLBACK' (default), or 'DELETE")
 	flags.String("aws-parameter", "", "A key-value pair to supply to the AWS template")
+	flags.String("aws-region", "", "The region to use when launching the instance")
+	flags.String("aws-stackname", "", "The name of the AWS stack that will be created")
 	flags.Bool("aws-sync", false, "If true, block until the command finishes (default: false)")
+	flags.String("aws-template", "", "UNSUPPORTED: cloud formation template url")
 
 	return cmd
 }
--- a/cli/command/cluster/create.go
+++ b/cli/command/cluster/create.go
@@ -87,6 +87,7 @@ func create(c cli.Interface, cmd *cobra.Command) error {
 	config := PluginConfig{
 		Provider: opts.provider,
 		Options: opts.options,
+		DockerOpts: opts.docker,
 	}
 	p, err := NewPlugin(config)
 	if err != nil {
--- a/cli/command/cluster/create.go
+++ b/cli/command/cluster/create.go
@@ -1,18 +1,21 @@
 package cluster
 
 import (
+	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/appcelerator/amp/cli"
 	"github.com/appcelerator/amp/cmd/amplifier/server/configuration"
 	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
 )
 
 // NewCreateCommand returns a new instance of the create command for bootstrapping an cluster.
 func NewCreateCommand(c cli.Interface) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "create [OPTIONS]",
-		Short:   "Create an amp cluster",
+		Short:   "Set up a cluster in swarm mode",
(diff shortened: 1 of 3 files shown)


cli/command/cluster/create.go
```diff
@@ -21,33 +24,76 @@ func NewCreateCommand(c cli.Interface) *cobra.Command {
...
 	flags.StringVarP(&opts.tag, "tag", "t", c.Version(), "Specify tag for cluster images (use 'local' for development)")
 	flags.StringVarP(&opts.registration, "registration", "r", configuration.RegistrationNone, "Specify the registration policy (possible values are 'none' or 'email')")
 	flags.BoolVarP(&opts.notifications, "notifications", "n", false, "Enable/disable server notifications (default is 'false')")
+
+	// aws options
+	flags.String("aws-region", "", "The region to use when launching the instance")
+	flags.String("aws-stackname", "", "The name of the AWS stack that will be created")
+	flags.String("aws-onfailure", "", "'DO_NOTHING', 'ROLLBACK' (default), or 'DELETE")
+	flags.String("aws-parameter", "", "A key-value pair to supply to the AWS template")
+	flags.Bool("aws-sync", false, "If true, block until the command finishes (default: false)")
```
ndegory: as discussed, we also need the `--aws-template` flag to be able to use a non default template. Because it's a URL, it's probably better to use something similar to what is done in the aws CLI: `--template-url` (`--aws-template-url`). That would allow us to later add a --aws-template-body.

cli/command/cluster/create.go
```diff
@@ -21,33 +24,76 @@ func NewCreateCommand(c cli.Interface) *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.IntVarP(&opts.workers, "workers", "w", 2, "Initial number of worker nodes")
-	flags.IntVarP(&opts.managers, "managers", "m", 3, "Intial number of manager nodes")
+	flags.IntVarP(&opts.managers, "managers", "m", 3, "Initial number of manager nodes")
```
ndegory: the --managers and --workers are global options of the CLI, but the AWS plugin has its own options for node counts (depending on the template).
We should guess a default parameter name (the one from the Docker for AWS template is probably the best we can choose), and when apply the -m and -w values to these --aws-PARAMETER.

ndegory (dismissed):

cli/command/cluster/create.go
subfuzion: Yes, I already mentioned I intended to do this. Keep in mind this is only temporary to make it easy for us to switch while we are testing. We don't want to expose this here at the amp cli level and have users swapping to an unsupported template). I'm exposing it for now as `--aws-template` since the plugin expects `--template`.

cli/command/cluster/create.go
subfuzion: Yes, don't worry about the local options just yet -- please focus on the aws stuff for right now. I mentioned when I spoke to you all this is getting refactored (it's a wip) and asked you to just let me know if you have any issue with aws. Thanks!

subfuzion: The options are generic options for the `cluster create` command. The choices I'm considering for changing this are either that CLI users only use options that are specific for each provider (so for example, aws: `--parameter ClusterSize=5`; local: `--parameter workers=5` or else we automatically map these generic options to provider specific options internally (for example for aws we map `--workers=5` to `--parameter ClusterSize=5`). The second choice is nice, but a bit magical; the first choice is more explicit. I'm leaning toward implementing the first one for right now, but the second choice is easy enough to do. We would need to make sure to document and make clear to users that there are standard options that we handle with the command, and then more provider-specific options that are handled in a provider-specific way.

cli/command/cluster/create.go
subfuzion: @ndegory Try it when you have a chance.

cli/command/cluster/create.go
ndegory: fine by me, moving the generic options to provider specific options. At least that wouldn't lead to side effects due to different default values and to not always perfect mappings between the cli options and the plugin options.
I couldn't test the PR because of the missing template option, but now that you added it I will give it a try.

ndegory:
I'm confused about the use of the _parameter_ options (or _aws-parameter_ more exactly), if I specify multiple parameters, will they all be passed to the plugin?
The output from the CLI let me think that it's not the case (I pass multiple --aws-parameter, the debug output only prints one --parameter, the last --aws-parameter that was provided).

subfuzion:
@ndegory Yes, it will work as expected, like the plugin itself does. To make it work, it needs to be a `StringSlice` (like I did for the plugin itself [here](https://github.com/appcelerator/amp/blob/master/cluster/plugin/aws/cmd/main.go#L100)), not a `String` (as I inadvertently did [here](https://github.com/appcelerator/amp/blob/0355930a3108da1d33f83e833406b64fc1e0a34c/cli/command/cluster/create.go#L36)). I'm still working on that right now actually and haven't tested yet, but it will work and I will be committing again shortly.

ndegory (approved):
LGTM

PR #1426, opened 2017-06-23
rename amplifier-gateway to gateway

### LINKED ISSUE #1422: Rename `amplifier-gateway` to `gateway`
The current gRPC gateway service name is `amplifier-gateway` which is long and might be also inappropriate (if we ever proxy other services behind). It should be renamed to `gateway` everywhere:
- folder name in the code,
- executable name,
- Dockerfile,
- code occurence,
- stack file.
- service name,
- Makefile,
- etc.

neha-viswanathan:
closes #1422

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,11 +185,11 @@ clean-server:
 
 
 # =============================================================================
-# BUILD GATEWAY (`amplifier-gateway`)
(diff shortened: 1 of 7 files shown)


bquenin (approved):
LGTM

PR #1429, opened 2017-06-23
Set CLI logger to non verbose

### LINKED ISSUE #1427: Set the CLI logger back to non verbose
The CLI is currently outputting some error messages twice on error. It seems that it's related to the gRPC logger set to `verbose` mode in the CLI: `cli/logger.go`.

neha-viswanathan:
Resolves #1427


bquenin (approved):
LGTM

PR #1438, opened 2017-06-27
Implements TLS for amp, amplifier and gateway

### LINKED ISSUE #1058: CLI - server TLS connection
> The connection between the amp CLI and amplifier is not secured by default.
> We already can add TLS termination on the amp stack, but the CLI should be modified to allow TLS.
> It can be TLS by default and fallback to non TLS if it's unavailable, and it can also use an option to force or disable mutual auth.

What needs to be done on `amplifier`:
- TLS should be enabled by default.
- `amplifier` will use the certificate located in `certificate_atomiq` docker secret.

What needs to be done on `amp` CLI:
- Enable TLS connection by default to amplifier.
- Add an option to skip certificate verification, for instance `-k` as in curl.

What needs to be done on `amplifier-gateway`:
- Update the gateway to use a TLS connection to amplifier if the server is using it.

The UI is not impacted as it's talking to the `amplifier-gateway` only.

bquenin:
Resolves #1058 

This PR enables TLS for the AMP platform:
- `amplifier` reads the server certificate chain and key from the docker secret `certificate_atomiq`, using the same certificate as the reverse proxy. There is currently no option to listen without TLS.
- `amp` CLI has been updated accordingly to negotiate TLS by default. There is currently no option to establish a clear connection. However, a `-k` (or `--insecure`) option has been added to enable skipping the certification chain verification (typically when using a self-signed certificates in local deployment).
- the gRPC gateway has been updated accordingly to negotiate TLS by default. There is currently no option to establish a clear connection.

The smokes tests and integration tests have been updated accordingly.

## How to test
```
$ ampmake build
$ docker secret rm certificate_atomiq
$ amp cluster create -t local
$ amp -k version
```

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"strings"
 
+	"github.com/appcelerator/amp/cmd/amplifier/server/configuration"
 	"github.com/spf13/cobra"
 	"google.golang.org/grpc"
 )
@@ -22,6 +23,7 @@ type Interface interface {
 
 	Server() string
 	SetServer(server string)
+	SetSkipVerify(skipVerify bool)
 	Connect() (*grpc.ClientConn, error)
 	ClientConn() *grpc.ClientConn
 
@@ -38,6 +40,7 @@ type cli struct {
 	out        *OutStream
 	err        io.Writer
 	clientConn *grpc.ClientConn
+	skipVerify bool
 }
 
 const (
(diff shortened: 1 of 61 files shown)


subfuzion (approved):
LGTM. @ndegory would like your review on this as well.

ndegory (dismissed):
LGTM

subfuzion (changes requested):
@bquenin Please resolve conflicts.

bquenin:
@subfuzion @ndegory done

subfuzion (approved):

PR #1444, opened 2017-06-27
amp agent boilerplate

subfuzion:
This is the boilerplate for the image that will be used to run a container in a target cluster environment. It will be used for running tasks, such as environment smoke tests, initialization, clean up, etc. There is nothing really to verify since this is meant to be started code for adding commands as we go.

Note: We will probably want to choose a better name ultimately, so as not to be confused with Lyft Envoy.

From the README:

# Envoy

Runs a container in a target cluster environment to perform various tasks, which may include
set up, smoke tests, etc.

`Dockerfile.compiler` builds an image for compiling the Go source files for
the final target image (`appcelerator/amp-envoy`).

`Dockerfile` builds the image for running a container in the target cluster.

# Build

`make compiler` builds an Alpine image (`appcelerator/amp-envoy-compiler`) with
a Go compiler, the AWS SDK package, and other necessary packages to to build 
the `appcelerator/amp-envoy` image.

An automated build for the repo also creates the `appcelerator/amp-envoy-compiler`
image on [Docker Hub](https://hub.docker.com/r/appcelerator/amp-envoy-compiler/).

`make` (or `make build`) builds `appcelerator/amp-envoy`.

`make clean` removes the target binary (`envoy.alpine`) that is created by the
compiler to be copied into the `appcelerator/amp-envoy` image when building it.

`make test` uses `appcelerator/amp-envoy` to create, update, and remove test stacks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -360,6 +360,23 @@ test-amp-aws: build-amp-aws
 	@cd $(CPAWSDIR) && $(MAKE) test
 
 # =============================================================================
+# BUILD ENVOY (`amp-envoy`)
+# =============================================================================
+ENVOYDIR := cluster/envoy
+
+.PHONY: build-amp-envoy-compiler
+build-amp-envoy-compiler:
+	@cd $(ENVOYDIR) && $(MAKE) compiler
+
+.PHONY: build-amp-envoy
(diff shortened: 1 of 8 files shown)


ndegory (approved):
LGTM

PR #1447, opened 2017-06-27
reduce fail rate at aws cluster creation

ndegory:
the worker groups now are created only after the manager group is fully up.
This reduces the risk to see a transition phase with worker instances getting killed because the manager nodes are not fully available.


subfuzion (approved):
LGTM

PR #1456, opened 2017-06-28
Dashboard updates

ndegory:
- use tagged versions of monitoring related images for better stability
- local deployment swarm init used to break the metrics gathering process
- the monitoring image (prometheus) doesn't need access to the manager nodes anymore. It discovers the nodes based on the ingress network on the local node.
- updated dashboards with swarm information and various fixes in existing graphs.


bquenin (commented):
Not related to your PR but it might be a good idea to reference the `monitoring` folder in the `fmt` rule of the Makefile.

ndegory:
@bquenin that's 2 open PRs with the same line to modify, as it's not related to this PR, let's do it next time. The good thing is that this folder is already well formatted, so no issue today.

subfuzion (approved):
LGTM

PR #1452, opened 2017-06-27
Updating vendor

bquenin:
Related to #1449 and #1450 

Adding the following packages to vendor:
- `github.com/grpc-ecosystem/go-grpc-middleware`
- `github.com/grpc-ecosystem/go-grpc-prometheus`
- `github.com/prometheus/client_golang`
- `github.com/beorn7/perks/quantile`

## How to test
Check Travis build

PR #1457, opened 2017-06-28
Instrument amplifier with gRPC prometheus metrics

### LINKED ISSUE #1450: Using gRPC middleware for prometheus metrics
Study the usage of https://github.com/grpc-ecosystem/go-grpc-prometheus to instrument `amplifier`.

bquenin:
:warning: needs to be merged after #1452
 
Fix #1450 

This PR enables the https://github.com/grpc-ecosystem/go-grpc-prometheus middleware for amplifier.

:warning: This PR introduces an HTTP/1 server inside `amplifier` in a separate go routine, as well as a new listening port (5100).
I didn't figure out how to serve both protocols (HTTP/1 and HTTP/2) from the same port. It might be possible somehow ... Also note that this new port is not publicly exposed.

A smoke test has been added to `platform/tests/metrics/amplifier-prometheus_test.sh`.

## How to test
```
$ ampmake build
$ amp cluster create -t local
$ docker run --rm --network ampnet appcelerator/amptools curl -sf http://amplifier:5100/metrics
```

### DIFF
--- a/platform/tests/metrics/amplifier-prometheus_test.sh
+++ b/platform/tests/metrics/amplifier-prometheus_test.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+docker run --rm --network ampnet appcelerator/alpine:3.6.0 curl -sf http://amplifier:5100/metrics
--- a/cmd/amplifier/main.go
+++ b/cmd/amplifier/main.go
@@ -46,6 +46,7 @@ func main() {
 		Version:          Version,
 		Build:            Build,
 		Port:             configuration.DefaultPort,
+		H1Port:           configuration.DefaultH1Port,
 		EmailSender:      mail.DefaultSender,
 		SmsSender:        sms.DefaultSender,
 		EtcdEndpoints:    []string{etcd.DefaultEndpoint},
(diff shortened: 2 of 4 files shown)


platform/tests/metrics/amplifier-prometheus_test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+docker run --rm --network ampnet appcelerator/amptools curl -sf http://amplifier:5100/metrics
```
ndegory: amptools is quite big, unless it's already present on the VM, it would be faster to use appcelerator/alpine:3.6.0, it also has curl.

ndegory (dismissed):
LGTM (the test can be optimized though)

platform/tests/metrics/amplifier-prometheus_test.sh
subfuzion: I agree with the principle, although in all probability, `amptools` has already been pulled to the system since it's used for building. My question is do we really need to use docker for this? Isn't curl going to be present?

subfuzion (dismissed):
LGTM

subfuzion:
@ndegory I'll leave it up to you and @bquenin about changing the image for the test. I'm fine with it since amptools is used pervasively for building/testing, so it will already be present on in the build environment. But yes, it's a huge beast getting close to 1.4GB and would definitely be massive overkill for a one use just to curl (and isn't curl already present anyway? Do we even need to launch a container for this?). Anyway, you can merge the PR when you're fine with it.

ndegory (approved):

PR #1463, opened 2017-06-28
envoy use multi-stage build, add vendoring

subfuzion:
This PR adds vendoring support using [vndr](https://github.com/LK4D4/vndr) and multi-stage support for building the target image.

The multi-stage Dockerfile will pull the specified vendor packages and then run tests, ensuring that the build environment is valid and tests passes before building the final stage for a lightweight image based only on alpine and the executable.

You can run the following make rules:

* `make` (or `make build`) - builds target `bin/envoy`
* `make clean` - removes `bin/envoy`
* `make test` - runs go tests
* `make build-image` - builds `appcelerator/envoy` using the multi-stage `Dockerfile`

## Verification

All the make rules should succeed.


bquenin (approved):
LGTM

ndegory (approved):
LGTM

PR #1482, opened 2017-07-01
Custom AMI and ampadmin execution

ndegory:
- custom built AMI for AMP
  - the system package installation (as well as Docker) is now done once at AMI build
  - upstream repo issues won't affect the creation of a stack
  - quicker boot
  - it is used when the `LinuxDistribution` parameter is set to `Default` (this is now the default)
  - to build it, read the README.md file
- new stack parameter: `InstallApplication` (defaults to `true`)
  - when true, will wait for the full cluster to be up and will run a appcelerator/ampadmin container
  - a new WaitCondition is linked to the end of this phase, the stack creation will wait for it

Limitations:
the custom built image is only available from the appc account, it should be public so that other accounts can use it, but it requires more work to get there.

### DIFF
--- a/examples/clusters/README.md
+++ b/examples/clusters/README.md
@@ -7,7 +7,7 @@ It is compatible with regions us-east-1, us-east-2, us-west-2, eu-west-1 and ap-
 
 Alternatively it can be used through the aws plugin of AMP:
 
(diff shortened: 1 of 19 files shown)


examples/clusters/README.md
```diff
@@ -54,3 +48,16 @@ The output of the stack lists the DNS name of the ELB in front of the manager no
 | VpcId | VPC ID |
 | DNSTarget | public facing endpoint for the cluster, It can be used for ssh access, https access to swarm services and configuration of the remote server in the CLI |
 | MetricsURL | URL for cluster health dashboard |
+
+## Custom AMI
+
+the default option for the AMI (Default) is a pre package AMI based on Ubuntu Xenial, with prerequisite packages already installed (in particular Docker).
```
subfuzion, line 54: 👍 (regarding the build documentation -- thanks!)

subfuzion (approved):
Nice! 👍

PR #1483, opened 2017-07-01
public AMI

ndegory:
This PR fixes the limitation of PR #1482. The custom AMI is now public and can be used in any other AWS account.


subfuzion (approved):
LGTM

PR #1501, opened 2017-07-06
move from the local tag to the dev version

### LINKED ISSUE #1331: amp cluster create error (missing amp-bootstrap)
The default command usage should not fail, so a failure is considered a bug.

Currently it uses the version that is compiled into the amp binary. So, for example, in the following case it uses `appcelerator/amp-bootstrap:v0.12.0-dev`.  Since there is no version of the image on the hub at the moment and since the locally built image is `appcelerator/amp-bootstrap:local`, the command fails:

```
$ amp --server=localhost cluster create
[localhost:50101]
Unable to find image 'appcelerator/amp-bootstrap:v0.12.0-dev' locally
docker: Error response from daemon: manifest for appcelerator/amp-bootstrap:v0.12.0-dev not found.
See 'docker run --help'.
f573e897-7aa0-4516-a195-42ee91039e97
```

At the very least we should augment the message to say that `X-dev` versions aren't published, did the user want to specify the `--tag=local` or `--tag=latest` option?

ndegory:
Fix #1331 (amp cluster create error on dev version)

When using the AMP CLI compiled from the repo (dev version, as opposed to the downloadable released versions), it used to reference the `local` tag of core services images, as well as the amp-bootstrap image.
From now on, it will use the dev version (0.13.0-dev as of today) instead.
As an exception, the portal image is now deployed from the `latest` tag, as it's not part of the repo anymore.

### Verification
```
# remove all of the previous local images, for instance with:
docker image ls | grep -w local | awk '{print $1 ":local"}' | xargs docker image rm
# build the 0.13.0-dev images:
ampmake build
# create the cluster
amp cluster create
# check the version of the service
docker ps --format '{{.Image}}'
# there should be no local image
```

### DIFF
--- a/cli/command/cluster/create.go
+++ b/cli/command/cluster/create.go
@@ -29,7 +29,7 @@ func NewCreateCommand(c cli.Interface) *cobra.Command {
 	flags.BoolVarP(&opts.notifications, "notifications", "n", false, "Enable/disable server notifications (default is 'false')")
 	//flags.StringVar(&opts.provider, "provider", "local", "Cluster provider (\"local\" (default) or \"aws\")")
 	flags.StringVarP(&opts.registration, "registration", "r", configuration.RegistrationNone, "Specify the registration policy (possible values are 'none' or 'email')")
-	flags.StringVarP(&opts.tag, "tag", "t", c.Version(), "Specify tag for cluster images (use '0.13.0-dev' for development)")
+	flags.StringVarP(&opts.tag, "tag", "t", c.Version(), "Specify tag for cluster images")
 
 	// local options
 	flags.String("local-managers", "1", "Initial number of local manager nodes")
--- a/cli/command/cluster/create.go
+++ b/cli/command/cluster/create.go
@@ -29,7 +29,7 @@ func NewCreateCommand(c cli.Interface) *cobra.Command {
 	flags.BoolVarP(&opts.notifications, "notifications", "n", false, "Enable/disable server notifications (default is 'false')")
 	//flags.StringVar(&opts.provider, "provider", "local", "Cluster provider (\"local\" (default) or \"aws\")")
 	flags.StringVarP(&opts.registration, "registration", "r", configuration.RegistrationNone, "Specify the registration policy (possible values are 'none' or 'email')")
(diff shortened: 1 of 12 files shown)


bquenin (dismissed):
LGTM

cli/command/cluster/status.go
```diff
@@ -17,7 +17,7 @@ func NewStatusCommand(c cli.Interface) *cobra.Command {
 	}
 	flags := cmd.Flags()
 	//flags.StringVar(&opts.provider, "provider", "local", "Cluster provider")
-	flags.StringVarP(&opts.tag, "tag", "t", c.Version(), "Specify tag for bootstrap images (default is '0.12.0', use 'local' for development)")
+	flags.StringVarP(&opts.tag, "tag", "t", c.Version(), "Specify tag for bootstrap images (use '0.13.0-dev' for development)")
```
bquenin: idem

cli/command/cluster/create.go
```diff
@@ -29,7 +29,7 @@ func NewCreateCommand(c cli.Interface) *cobra.Command {
 	flags.BoolVarP(&opts.notifications, "notifications", "n", false, "Enable/disable server notifications (default is 'false')")
 	//flags.StringVar(&opts.provider, "provider", "local", "Cluster provider (\"local\" (default) or \"aws\")")
 	flags.StringVarP(&opts.registration, "registration", "r", configuration.RegistrationNone, "Specify the registration policy (possible values are 'none' or 'email')")
-	flags.StringVarP(&opts.tag, "tag", "t", c.Version(), "Specify tag for cluster images (use 'local' for development)")
+	flags.StringVarP(&opts.tag, "tag", "t", c.Version(), "Specify tag for cluster images (use '0.13.0-dev' for development)")
```
bquenin: `(use '0.13.0-dev' for development)` I'm not sure we need to document that inside the CLI. This should be part of the dev documentation instead.

bquenin (changes requested):
LGTM

cli/command/cluster/create.go
ndegory: correct, I'll remove it

bquenin (approved):

PR #1506, opened 2017-07-07
Handle case insensitive values for CLI command args/options

### LINKED ISSUE #1505: Some args and options for CLI commands should accept case-insensitive values
The following arguments and options should accept case-insensitive values on the CLI
- Organization member role
- Team resource permission level

neha-viswanathan:
resolves #1505 

Organization member role and Team resource permission level now accept case-insensitive values.

### Test
Smoke tests have been updated

### DIFF
--- a/cli/command/org/member/role.go
+++ b/cli/command/org/member/role.go
@@ -2,8 +2,8 @@ package member
 
 import (
 	"errors"
-
 	"fmt"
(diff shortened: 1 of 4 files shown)


JosephGJ (approved):
LGTM

bquenin (approved):
LGTM

Issue #1510, opened 2017-07-07
amp cluster create command starts too many services


subfuzion:
These services shouldn't be started (and no reason for 3 replicas of pinger).

```
$ amp cluster create
[user tony @ localhost:50101]
. . .
ID                  NAME                MODE                REPLICAS            IMAGE                                    PORTS
ss4cxe8q7i31        vis_portainer       replicated          1/1                 portainer/portainer:latest
wi41ujisapav        vis_visualizer      replicated          1/1                 manomarks/visualizer:latest
zszm5dji5ywc        pinger_pinger       replicated          3/3                 subfuzion/pinger:latest
```

subfuzion:
Note also that this fails in any case because visualizer has moved:

    wi41ujisapav        vis_visualizer      replicated          0/1                 manomarks/visualizer:latest

The new repo is:  dockersamples/visualizer

subfuzion:
@ndegory I was under the impression you handle last week for the release. Are you still going to take care of it or do you want me to? You moved the stack itself from `platform/stacks` to `examples`, but this isn't reflected yet with the bootstrap image and right now `cluster create` fails (due to visualizer).

ndegory:
@subfuzion the `amp cluster create` doesn't deploy the examples.
Here's the list I got after a fresh deployment:
```
ID                  NAME                   MODE                REPLICAS            IMAGE                                          PORTS
1hdxne9me94f        amp_nats               replicated          1/1                 appcelerator/amp-nats-streaming:v0.5.0
5ghiteb9m6zv        amp_etcd               replicated          1/1                 appcelerator/etcd:3.1.9
661qyut2ztat        amp_gateway            global              1/1                 appcelerator/gateway:0.13.0-dev
9o6avb6nset6        amp_node_exporter      global              1/1                 prom/node-exporter:v0.14.0                     *:9100->9100/tcp
e3ezjo7zkx9q        amp_amplifier          global              1/1                 appcelerator/amplifier:0.13.0-dev              *:50101->50101/tcp
iyjd9lukm9nn        amp_elasticsearch      replicated          1/1                 appcelerator/elasticsearch-amp:5.4.2
lmq4wbz0h06f        amp_kibana             replicated          1/1                 appcelerator/kibana:5.4.2
lnjqsz3sc24v        amp_nats_exporter      replicated          1/1                 appcelerator/prometheus-nats-exporter:latest
[...]

PR #1512, opened 2017-07-07
deploy script won't fail if there's a non core broken service

### LINKED ISSUE #1299: deploy never starts if it exists a failing service before it starts
If, before launching 'deploy' it exists a failed service, relaunching its containers and so not having the right replicas count, deploy will integrate it in its own count and then alway having a bad count after launching its first own services and so exit in timeout after a while

Deploy should take in account only its own services in its replicas calculation

ndegory:
Fix #1299 
Ref #1510 
Ref #1513

If there's an existing service with failing tasks on the local node where the `amp cluster create` is executed, the cluster creation fails. This PR fixes it by ignoring the non core services at the service check step.

## Verification

create a broken service:
```
docker service create --name broken_service alpine:3.6 /bin/false
```
This would prevent the creation of the cluster with AMP 0.12.0.

```
make build-cli
amp cluster create
```
The deployment should not fail anymore.
Don't forget to remove the test service:
```
docker service rm broken_service
```

### DIFF
--- a/platform/bin/servicescheck
+++ b/platform/bin/servicescheck
@@ -5,7 +5,7 @@ INTERVAL="${2:-2}"
 [[ "$DOCKERHOST" = "m1" ]] && amps="docker run -it --rm --network=hostnet docker --host=m1" || amps="docker --host=$DOCKERHOST"
 
 check() {
-  list=$($amps service ls)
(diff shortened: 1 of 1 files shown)


JosephGJ (approved):
LGTM

PR #1517, opened 2017-07-09
Disabling organization management 

bquenin:
Fix #1509 
Fix #1511 

This PR disables organization management.

All users are automatically added as members of the default organization. There are no personal accounts anymore. The first created user is the owner of the default organization.
`amp org` commands have been removed in the CLI.
Tests have been updated accordingly.

## How to test
Check Travis build

Issue #1509, opened 2017-07-07
amp cli should not show org


subfuzion:
`org` Is not supposed to be v0.12/v0.13 releases.

    $ amp
    ...
    Management Commands:
      org         Organization management operations
      ...

bquenin:
@subfuzion `team`s are associated to `org`s which is why they're still there. What I can do, is create a default org and remove the org creation/deletion commands in the CLI. What do you think?

subfuzion:
@bquenin  The command should not be exposed. We discussed this already. Yes, there is supposed to be a default org and no ability to create or manage orgs for this release.

bquenin:
@subfuzion The only issue is that teams are attached to organizations, so removing organizations removes also the teams. And teams are the only way to share resources.
I can remove organizations and leave the teams there, but there is a significant refactoring (design change).

subfuzion:
@bquenin Yes, I understand that teams are what correlate permissions to resources (I designed the [original conceptual schema](https://app.quickdatabasediagrams.com/#/schema/fPxxEg7bME6rrChvh691qw)). Again, this detail shouldn't be exposed by the CLI. Teams are supposed to be attached to a default org and the details of that org aren't supposed to be exposed. The `amp org` command shouldn't be there. We've already had the conversation about the refactoring.

bquenin:
@subfuzion This is not a detail, you are either working using your personal account or using an organization account. Even if we use a default organization and hide the organization commands, the user should switch between his personal account or organization account depending on what he wants to do.

subfuzion:
@bquenin Yes, again -- we discussed this. There is the only use case we support for now -- the organization account that supports teams. The user is automatically a member of it (member of owners team specifically) and automatically switched to it, and there is no switching to the personal account for now. The use case for right now is to only support a user being in a mode that can create teams to provide access to resources.

PR #1518, opened 2017-07-10
Remove `org` commands from /docs

JosephGJ:
ref: #1509 
ref: #1517 

Removing `org.md` and organizations from the main `README.md` in `/docs`. Added a section on creating teams and adding resources, stating users are added to a default organization


bquenin (approved):
LGTM

Issue #1434, opened 2017-06-26
General Documentation Updates


JosephGJ:
General documentation updates across the entire project. Going through by directory. Feel free to add anything if there are any changes you'd like to see.

### .github
- [ ] `ISSUE_TEMPLATE.md` 
    - Update the general support section. 
- [ ] `PULL_REQUEST_TEMPLATE.md`
    - Signing commits is suggested, but not implemented by anyone.
    - add "Resolves #xxxx"

### api
- [ ] `rpc/readme.md`
    - feels unnecessary, can remove this.

### cmd
- [ ] `README.md`
    - feels unnecessary, can remove this
    - same with READMEs in subdirectories

### data
- [ ] `README.md`
    - feels unnecessary, can remove this
- [ ] `storage/README.md`
    - feels unnecessary, can remove this

### docs
- [x] List all possible values for command options.
- [x] `README.md`
    - [x] Prerequisites 
        - Need to be more explicit. E.g. we need to state we are not supporting Docker Toolbox. 
        - Linux node
            - Aren't supporting DinD so don't need to load IPVS modules.
    - [x] Signup and Login on AMP
        - add `-s cloud.appcelerator.io` as localhost is now default target
    - [x] Create your own local AMP cluster
        - don't need `-s localhost` as it is default target
    - [x] Organization and Teams
        - More detail necessary
- [x] Most `.md` files need to have a consistent structure
- [x] Some of them have minor spelling and grammar mistakes

### examples
- [x] `clusters/README.md`
    - Use cli commands for aws plugin in "how to use" section
    - don't need `-s localhost` as it is default target

### images
- [ ] Readmes need have a consistent structure and more detail
- [ ] `nats/README.md`
    - Readme points to the `amp-nats-streaming` docker hub page for more information, but there is none on there
- [ ] `prometheus-nats-exporter`
    - Add readme detailing information on prometheus-nats-exporter

### platform
- [ ] `bin + bootstrap`
    - both directories should have detailed documentation explaining our infrakit deployment methods along with why we use infrakit and its benefits.
- [ ] `stacks`
    - small documentation explaining the purpose of each of our `.yml` stack files and how they fit into the core amp services.
- [ ] `testing and tests` <- should have better names
    - small documentation detailing our cli smoke tests from a developer stand point, include points on what these tests do for our project and how to create new tests.

### project
- [ ] `README.md`
    - A lot of the links are broken as the files have been removed
[...]

PR #1521, opened 2017-07-11
Adding documentation for `completion` cmd

JosephGJ:
ref: #1434 

Adding `completion.md` in the `/docs` folder. 

Minor updates to the existing docs due to recent changes on #1501


docs/README.md
```diff
@@ -185,11 +185,11 @@ Server:
...
 ```
-$ amp cluster create
+$ amp cluster create --tag [AMP version]
 ```
->*If you are using the unstable version of the CLI, you will need to build the `amp-bootstrap`
-image using the command `make build-bootstrap` or `ampmake build-bootstrap` if your don't have the build tools,
-and then specify the `--tag=local` option to the CLI*
+>*If you are using the unstable version of the CLI, you don't need to add the `--tag`
+option as it will use the development version as the image tag, which has already been
+built with `build-cli`*
```
ndegory: I would use `development` instead of `unstable`, that would not make any negative assumption on the stability of AMP :)
Also, instead of `make build-cli`, let's talk about `ampmake build`, this way we're sure all the images are available.

docs/cluster.md
```diff
@@ -37,20 +37,19 @@ We recommend you stick to the outlined examples for deploying cluster environmen
...
 ```
-$ amp cluster create --provider local
+$ amp cluster create --tag [AMP version]
 ...
 Cluster status: healthy
 ...
 ```
-This will deploy the `AMP` stack on your local docker engine. This AMP deployment only
+This will deploy the `AMP` stack on your local docker engine using the image version specified with the `--tag` flag.
+If you are using the unstable, development version of `AMP`, you will not need to provide this option. This AMP deployment only
```
ndegory: s/unstable, //
and the tag is not mandatory either with the released version of AMP, it's only useful when one wants to deploy the cluster in a different version than the CLI.

docs/completion.md
```diff
@@ -0,0 +1,101 @@
...
+cluster     completion  config      login       logout      logs        password    service     stack       stats       team        user        version     whoami
+```
+
+```
+$ amp stack
+deploy    logs      ls        rm        services
+```
+
+```
+$ amp stack deploy -c platform/stacks
```
ndegory: I'd prefer to use the examples/ folder instead of the platform/stacks/ for the example.

ndegory (changes requested):

docs/README.md
JosephGJ: done

docs/cluster.md
JosephGJ: done

docs/completion.md
JosephGJ: done

ndegory (approved):

Issue #1524, opened 2017-07-12
Remove references to `org`


neha-viswanathan:
Remove all references to `org` in other AMP commands.

PR #1532, opened 2017-07-14
Remove default org from CLI information

neha-viswanathan:
Ref #1524 

The default organization is no longer displayed on the CLI information. 
#### Before
```
$ amp whoami 
[user neha1 in organization default @ localhost:50101]
Logged in as user: neha1
```
#### After
```
$ amp whoami 
[user neha1 @ localhost:50101]
Logged in as user: neha1
```


bquenin (approved):
LGTM

PR #1535, opened 2017-07-14
Updated vendor

subfuzion:
This adds a package that I began prototyping last weekend to make the stack functionality of the CLI importable as a package. The vendored package is still a work in progress, but the agent can use it now to start a stack without having to shell out to the Docker CLI.

Currently `ampctl install` will deploy stacks in sorted (ascending) order under the `./stacks` directory.


bquenin (approved):
LGTM

PR #1537, opened 2017-07-16
Minor fixes in log messages

ndegory:
minor fixes in log messages, as we may use them to check service availability.


bquenin (approved):
LGTM

PR #1538, opened 2017-07-16
fix network options

ndegory:
- A few services were missing the network option in their definition
- Remove the network service aliases, they're duplicating the default name


bquenin (approved):
LGTM

PR #1539, opened 2017-07-17
Stack files refactoring for amp-agent

ndegory:
- AMP services are now deployed with one stack per service (*.amp.yml)
- a test stack file for each service that can be tested from another container (*.test.yml)
- labels defined at the service level for easier orchestration:
  - amp.service.stabilize.delay
  - amp.service.stabilize.timeout


bquenin (approved):
LGTM

PR #1548, opened 2017-07-19
Remove references to "atomiq"

### LINKED ISSUE #1547: Remove references to Atomiq
In a few places we're still using labels or names with `atomiq` instead of `amp`, it should be updated.

ndegory:
Fix #1547


bquenin (approved):
LGTM

PR #1550, opened 2017-07-20
add the node label to the atsea example

ndegory:
node label (for service scheduling) was missing.

## verification

won't make any difference on local deployment, but can be tested on a real swarm cluster. Deploy the stack (after making sure the expected secrets exist), and check that the tasks are all deployed on the `amp.type.user` nodes.


bquenin (approved):
LGTM

PR #1556, opened 2017-07-24
Alerting configuration

ndegory:
- configure Prometheus to send alerts to Alertmanager
- move the Prometheus rules configuration from a docker secret to a docker config


bquenin (approved):
LGTM

PR #1559, opened 2017-07-25
Cluster mode detection for AMP deployment

### LINKED ISSUE #1545: single instance or cluster mode for Elasticsearch and etcd
Ref #1544 (Core AMP Stack deployment)

The deployment of ES and etcd can be done in two ways: HA or not.
Based on the characteristics of the swarm cluster, one mode should be chosen.

If there's at least 3 swarm nodes with the label amp.type.search == true, Elasticsearch should be deployed with 3 replicas, else with a single replica.
Same with etcd and the label amp.type.kv == true.
Environment variables are also tuned for each use case.
This is already prepared in the stack files (01-elasticsearch|etcd-single|cluster.amp.yml).

ndegory:
- add ampagent in the main Makefile
- cluster mode check before core services deployment (Fix #1545)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ cleanall: clean cleanall-deps
 # When running in the amptools container, set DOCKER_CMD="sudo docker"
 DOCKER_CMD ?= "docker"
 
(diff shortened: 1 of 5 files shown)


bquenin (approved):
LGTM

PR #1565, opened 2017-07-25
Adding initial secret creation for amp

bquenin:


ndegory (approved):
LGTM. We'll also need the config for Prometheus, but let's do that in another PR.

PR #1568, opened 2017-07-25
system check for local deployment on Linux

### LINKED ISSUE #1446: Check system prerequisites for local deployment on linux
Related to #1401

We need to make sure of the following before deploying a local luster on linux:
- `ip_vs` module needs to be loaded (only for DinD deployments).
- `vm.max_map_count` must be greater than 262144.

If these requirements are not met, the local deployment should gracefully abort.

ndegory:
Fix #1446


bquenin (approved):
LGTM

PR #1569, opened 2017-07-25
Using environment variable to determine version

### LINKED ISSUE #1546: Environment variables in core stack files
Ref #1544 (Core AMP Stack deployment)

The core stack files in cluster/agent/stacks/ are using environment variables that should be passed while deploying AMP on a Swarm cluster.

These are mainly:
- TAG (version of the AMP core services)
- REGISTRATION (user registration mode)
- NOTIFICATIONS (enables notifications)

bquenin:
Fix #1546


ndegory (approved):
Awesome!

PR #1581, opened 2017-07-25
raise the application installation timeout

ndegory:
The timeout for the AMP stack deployment was too close to the real deployment time, resulting in deployment failure from time to time.
Doubling this timeout (backported in 0.13.0).


bquenin (approved):
LGTM

PR #1590, opened 2017-07-26
Fix ES alerts on multi nodes clusters

ndegory:
On clusters where ES has more than one replicas, alerts are falsely triggered.
Fixes the cluster status (es_cluster_status = 1 means YELLOW, es_cluster_status = 0 means GREEN).


bquenin (approved):
LGTM

bquenin (approved):
LGTM

PR #1570, opened 2017-07-25
Agent default command

ndegory:
- default behavior of ampagent image is to check and install
- cloudformation template runs the ampagent at the end of the install

PR #1592, opened 2017-07-28
Minor agent cli refactor to streamline default and to funnel errors to top level

subfuzion:
Related: https://github.com/appcelerator/amp/pull/1570

Simplifies code for default handling; also ensures all commands return an error which gets logged by the root handler.


ndegory (approved):
LGTM

Issue #1591, opened 2017-07-27
Local deployment should use ampagent


ndegory:
the AMP stack deployment on AWS is based on the ampagent (check and install commands).
The local deployment is still based on the older platform/bin/deploy script.
The local deployment should now use the same deployment method as the AWS plugin.

ndegory:
we should then make sure all steps are implemented in the ampagent:
- check the system prerequisites (sysctl)
- initialization of the swarm if it's not already initialized (local mode only)
  - identification of the network interface for the swarm initialization (local mode only)
- add amp.type.* labels if they're missing (local mode only)

the previous method also had the testrunner execution as an option, but I don't think it's a good idea to implement it in the ampagent, it can be ran on Travis as a single command.

subfuzion:
@bquenin Here are the moving parts from the AWS Plugin implementation to match up for local:
* AWS plugin: https://github.com/appcelerator/amp/tree/master/cluster/plugin/aws
* CLI support for running the plugins: https://github.com/appcelerator/amp/blob/master/cli/command/cluster/plugin.go#L162 (currently doesn't run any plugin but calls queryCluster; this should be updated to resemble awsPlugin.Run), https://github.com/appcelerator/amp/blob/master/cli/command/cluster/plugin.go#L162
Let me know if you would like me to walk through anything with you.

@ndegory Thanks for opening the issue for me. However, perhaps I misunderstood the last comment you made, but I do ultimately want the agent to be able to perform all the system checks as part of smoke testing on demand, independent of other environments (eg, travis) and scripts (eg, cloudformation). The goal is that it is responsible for being able to start up the amp backend *and* also able to verify the environment.

However, the other things you mentioned in your previous comment really should be part of the local plugin, not agent plugin. The agent is what we run inside the target cluster for setting up amp after it has been already initially provisioned as we do now with the AWS plugin. This is the extent of the install command's responsibility at the moment: https://github.com/appcelerator/amp/blob/master/cluster/agent/cmd/install.go#L42. There is overlap since the agent should verify the system meets all prerequisites. I prefer to keep the agent as simple as possible and conditional responsibilities to a minimum. Let's chat a bit over a whiteboard session perhaps before anyone starts work on this.

ndegory:
@subfuzion you're right, that first comment content is supposed to be in the local plugin, not in the ampagent, which should focus on the amp stack deployment.

subfuzion:
[...]

PR #1595, opened 2017-07-29
Adding uninstall command

bquenin:
Related to #1591 

Adding an `uninstall` command to `ampctl` that:
 - Remove the `amp` (or given namespace) stack,
 - Remove the `ampnet` network.

## How to test
```
$ cd cluster/agent
$ make
$ bin/ampctl install
$ bin/ampctl uninstall
$ docker stack ls
$ docker network ls
```

### DIFF
--- a/cluster/agent/cmd/uninstall.go
+++ b/cluster/agent/cmd/uninstall.go
@@ -1,0 +1,41 @@
+package main
+
+import (
+	"github.com/appcelerator/amp/cluster/agent/pkg/docker"
+	"github.com/appcelerator/amp/cluster/agent/pkg/docker/stack"
+	"github.com/docker/docker/pkg/term"
+	"github.com/spf13/cobra"
+)
+
+func NewUninstallCommand() *cobra.Command {
+	uninstallCmd := &cobra.Command{
+		Use:   "uninstall",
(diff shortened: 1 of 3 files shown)


cluster/agent/cmd/uninstall.go
```diff
@@ -0,0 +1,37 @@
...
+	namespace := "amp"
+	if len(args) > 0 {
+		namespace = args[0]
+	}
+
+	opts := stack.RemoveOptions{
+		Namespaces: []string{namespace},
+	}
+
+	if err := stack.Remove(dockerCli, opts); err != nil {
```
ndegory, line 32: is this async? In this case we may have an issue with the network removal just after.

ndegory (commented):
another thing we used to remove: the named volumes (amp_*).

cluster/agent/cmd/uninstall.go
bquenin: @ndegory Actually, the network removal is queued by the docker daemon and the network gets removed right after the stack is removed.

ndegory: cool

ndegory (approved):
LGTM

PR #1596, opened 2017-07-29
Local deployment plugin

ndegory:
Ref #1591

To be used by the CLI

### Verification

```
make build-amp-local
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v /var/run/docker:/var/run/docker appcelerator/amp-local:latest init
```


subfuzion (changes requested):
@ndegory Looks good overall (need a bit more time for review), but it would be better if the AWS updates were in their own PR instead of this one.

ndegory:
@subfuzion AWS/agent changes moved in #1597

subfuzion (approved):
LGTM 👍

PR #1598, opened 2017-07-31
Options for local deployment

### LINKED ISSUE #1586: add a --no-monitoring flag to the amp deployment
Local deployment can be lighter by skipping the monitoring block.

### LINKED ISSUE #1588: add a --fast option to the ampagent
The default deployment mode of the core stack includes tests, and it takes a significant time.
Adding a --fast option would allow to skip the tests and just run the amp core services.

ndegory:
Fix #1586 (option to skip monitoring services) 
Fix #1588 (option to skip tests)
Fix haproxy_exporter deployment test
Fix amplifier secret name

### Verification

```
make build-amp-local build-ampagent
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v /var/run/docker:/var/run/docker appcelerator/amp-local:0.14.0-dev init --fast
# the deployment is faster, the test stacks are not run
# remove the cluster, with docker stack rm amp for instance
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v /var/run/docker:/var/run/docker appcelerator/amp-local:0.14.0-dev init --no-monitoring
# the monitoring services are not deployed
# remove the amp services
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v /var/run/docker:/var/run/docker appcelerator/amp-local:0.14.0-dev init --fast --no-monitoring
# both expectations are met
```


cluster/plugin/local/plugin.go
```diff
@@ -99,7 +101,7 @@ func removeAgent(ctx context.Context, c *client.Client, cid string) error {
 	return c.ContainerRemove(ctx, cid, types.ContainerRemoveOptions{Force: true})
 }
 
-// RunAgent runs the ampagent image to init (action ="") or destroy (action="uninstall")
+// RunAgent runs the ampagent image to init (action ="install") or destroy (action="uninstall")
```
subfuzion, line 104: You've done a good job here making do with the current agent implementation, but it's a flaw in the agent that the cmd part isn't more separated from the package and it's something I want to fix. You should be able to just import the package to use here, not have to run another process just to pass the commands.

cluster/agent/stacks/09-amplifier.ampcore.yml
```diff
@@ -38,10 +38,10 @@ services:
         constraints:
         - node.labels.amp.type.api == true
     secrets:
-      - source: default.amplifier.yml
+      - source: amplifier_yml
```
subfuzion, line 41: I'd actually rather fix the name in the places where we're creating the secret (travis, etc) since it makes the intention clearer, but I guess we can do that another time.

subfuzion (approved):
LGTM

PR #1600, opened 2017-08-01
Adding support for amp secret rm

bquenin:
#1612 
Related: #1583

Adding support for `amp secret rm`

## How To Test
```
ampmake build
amp user signup --name user --email user@email.amp --password password
echo test1 | amp secret create test1 -
echo test2 | amp secret create test2 -
amp secret ls
amp secret rm test1 test2
amp secret ls
```


subfuzion (approved):
LGTM

PR #1603, opened 2017-08-01
Local plugin pulls agent image before running it

### LINKED ISSUE #1602: local plugin fails because of missing ampagent image
`RunAgent` calls ContainerCreate, but not ImagePull. The ideal fix, however, is not to pull and create an ampagent container, but instead to refactor ampagent so that necessary logic isn't embedded in the install command implementation, but available to be imported as a package that the local plugin can leverage.

For a temporary fix, I have the amp cluster create command ensuring that ampagent is pulled before running the local plugin container.

ndegory:
Fix #1602 

For released version: the plugin will pull the version image (e.g. 0.14.0)
For dev: the plugin will try to pull, but will fall back to the locally built image (e.g. 0.14.0-dev)
If using a `latest` local plugin, it will download the latest ampagent (that can be useful when deploying a remote cluster with a dev version).

In all cases, core services images used for the deployment will be the same as the tag version of the ampagent.

### Verification

pull the appcelerator/amp-local:latest image
remove the appcelerator/ampagent:latest image
run amp-local
it should pull ampagent (if the image on DockerHub has been built with version latest, not retagged from a dev version)

run the appclelerator/amp-local:0.14.0-dev
it should use the locally built amp-agent image

### DIFF
--- a/cluster/plugin/local/Makefile
+++ b/cluster/plugin/local/Makefile
@@ -8,7 +8,7 @@ IMAGE := appcelerator/amp-local
 export LDFLAGS := "-s -X=main.Version=$(VERSION) -X=main.Build=$(BUILD)"
 
 build:
-	go build -o $(TARGET) $(PKG)
+	go build -ldflags $(LDFLAGS) -o $(TARGET) $(PKG)
 
 vendor: vendor.conf
 	vndr
--- a/cluster/plugin/local/cmd/main.go
+++ b/cluster/plugin/local/cmd/main.go
@@ -66,6 +66,10 @@ func delete(cmd *cobra.Command, args []string) {
 	log.Println("cluster deleted")
 }
 
+func version(cmd *cobra.Command, args []string) {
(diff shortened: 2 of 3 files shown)


subfuzion (commented):
I already have a temporary fix [here](https://github.com/appcelerator/amp/blob/b633298fb0731d576a08217f013ada5acd6d55bd/cli/command/cluster/plugin.go#L163-L170), but this is better if it is going to be a permanent solution. However, I'm considering an alternative solution, as described in #1602. I want to refactor ampagent a bit so that it can be imported as a package by the local plugin -- I don't see a need to spin up another container just to run a command that will perform the installation -- and I was considering various strategies, like using a tool like [go-bindata](https://github.com/jteeuwen/go-bindata) to package up and expose the compose file assets. Let's chat at the office a bit before moving forward.

ndegory:
@subfuzion sure, we should move forward a better solution for long term. This PR also add a version command and the version variable to the binary build, which can be useful when testing changes locally.

PR #1607, opened 2017-08-01
Agent options

ndegory:
options from the CLI for tag, registration and notifications were ignored by the local deployment plugin, because expected as command line options, but passed as env variables instead.

Changed the local plugin to expect env variables.

Also removed a 10 sec sleep between all services, which should speed up the deployment.

### DIFF
--- a/cluster/plugin/local/cmd/main.go
+++ b/cluster/plugin/local/cmd/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 
 	plugin "github.com/appcelerator/amp/cluster/plugin/local"
 	"github.com/docker/docker/api/types/swarm"
@@ -24,6 +25,7 @@ var (
 	dockerClient  *client.Client
 	defaultLabels = map[string]string{"amp.type.api": "true", "amp.type.route": "true", "amp.type.search": "true", "amp.type.kv": "true", "amp.type.mq": "true", "amp.type.metrics": "true", "amp.type.core": "true", "amp.type.user": "true"}
 	opts          = &plugin.RequestOptions{
+		Tag:         Version,
 		InitRequest: swarm.InitRequest{},
 		Labels:      defaultLabels,
 		// sane defaults for the local plugin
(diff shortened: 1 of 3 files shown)


bquenin (approved):
LGTM

cluster/plugin/local/cmd/main.go
```diff
@@ -139,6 +138,20 @@ func main() {
 	}
 	destroyCmd.PersistentFlags().BoolVarP(&opts.ForceLeave, "force-leave", "", false, "force leave the swarm")
 
+	// override default value if env vars are set
```
subfuzion, line 141: We should be using command options instead of environment variables. It would be better to align the CLI to supply the correct args instead.

cluster/plugin/local/plugin.go
```diff
@@ -150,7 +150,7 @@ func RunAgent(ctx context.Context, c *client.Client, action string, opts *Reques
 	reader, err := c.ImagePull(ctx, image, types.ImagePullOptions{})
 	if err != nil {
 		// don't exit, if not in the registry we may still want to run the container with a local image
-		fmt.Printf("ImagePull failed, which is fine if using a development version: %f", err)
+		fmt.Println("ampagent image pull failed, which is expected on a development version")
```
subfuzion, line 153: Recommend:
`fmt.Sprintf("%s not found in registry; will attempt to use local image if available\n", image)`

cluster/plugin/local/plugin.go
```diff
@@ -150,7 +150,7 @@ func RunAgent(ctx context.Context, c *client.Client, action string, opts *Reques
...
 	} else {
 		// wait for the image to be pulled
```
subfuzion, line 155: You need `defer reader.Close()` here.

The make slice statement can be made more succinct since capacity defaults to length (`make([]byte, 1000)`). Related to this, In several other areas where you make channels, it's a bit more idiomatic to just use `make(chan <type>)` without explicitly specifying 1 for unbuffered channels.

But instead of doing this anyway, maybe import `github.com/docker/docker/pkg/jsonmessage` and then try using the following to print the response:

```
jsonmessage.DisplayJSONMessagesStream(
		reader,
		os.Stdout,
		os.Stdout.Fd(),
		false,
		nil)
```

Finally, even if we're going to refactor this or potentially eliminate it later, you can improve the code density / readability a bit by factoring out this image pull logic into a separate func.

PR #1618, opened 2017-08-02
CLI and SIGINT

### LINKED ISSUE #1166: CLI should trap interrupt signal
When Ctrl-C-ing the CLI while deploying a cluster, the signal is not sent to the bootstrap container, this leads to situation where the CLI can't be run anymore because of an existing amp-bootstrap container.

ndegory:
Fix #1166 

The CLI traps the interruption signal and stops the deployment plugin container, allowing it to stop without remaining container that would break a subsequent CLI command.

## Before the fix
a Ctrl-C on a `amp cluster create` would leave a amp-cluster-plugin-xxx container as well as ampagent. These two containers have to be manually killed and removed.

## After the fix
a Ctrl-C on a `amp cluster create` will stop (and remove) these two containers.


bquenin (approved):
LGTM

subfuzion (approved):
LGTM

PR #1620, opened 2017-08-03
Fix deployment with test but no monitoring

### LINKED ISSUE #1616: ampagent with option no-monitoring runs extra tests
when ampagent is ran without the fast option but with the no-monitoring option, the tests for the monitoring services are still run, leading to a deployment failure.

ndegory:
Fix #1616 

## Verification

```
$ amp cluster create --local-no-monitoring
# and for no regression tests, amp cluster create --local-fast and amp cluster create
```


subfuzion (dismissed):
getStackFiles is becoming a tad bit unwieldy at this point, and is a candidate for refactoring. Also, by changing the filename convention a little, we can simplify the logic and simply compute the companion test file:
Currently: 01-elasticsearch-cluster.ampmon.yml => 01-elasticsearch-cluster.ampmon.test.yml
Simpler:
For any ampmon.yml, compute the suffix:
// ampmon.yml => ampmon.test.yml

subfuzion (approved):
Good enough for now. 👍

PR #1634, opened 2017-08-10
Delete AMP stack on cluster creation failure

### LINKED ISSUE #1632: On cluster creation a failure should leave a clean state
The deployment of the AMP cluster with the ampagent shouldn't fail and leave whatever has been deployed. A rollback should be done.
In case of a creation (AMP was not running), it should delete the amp stack.
In case of an update (the amp stack already exists), it shouldn't delete the stack, _instead it should also revert to the previous state_. That last part can be done in a second step, but the rest of the issue should be addressed asap.

neha-viswanathan:
fixes #1632 

Handles Interrupt signals in the amp agent during cluster creation and ensures the AMP stack is deleted whenever it fails.

#### To Test

1. Hit `Ctrl+C` during `amp cluster create`. In another terminal window, run `docker service ls` command. You will see that no AMP services are running.
2. When using the dev version of the CLI, remove one image from the local cache (appcelerator/ampbeat:0.15.0-dev for instance) and run `amp cluster create`. It should fail and rollback the deployment.

### DIFF
--- a/cluster/agent/cmd/install.go
+++ b/cluster/agent/cmd/install.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"os/signal"
 	"path"
 	"path/filepath"
 	"regexp"
@@ -29,6 +30,7 @@ import (
 const (
 	TARGET_SINGLE  = "single"
 	TARGET_CLUSTER = "cluster"
+	TEST_NAMESPACE = "test"
 )
 
 type InstallOptions struct {
@@ -71,6 +73,17 @@ func install(cmd *cobra.Command, args []string) error {
 		namespace = args[0]
 	}
 
+	// Handle interrupt signal
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		for range c {
+			log.Println("\nReceived an interrupt signal - removing AMP services")
+			stack.Remove(dockerCli, stack.RemoveOptions{Namespaces: []string{namespace, TEST_NAMESPACE}})
+			os.Exit(1)
+		}
+	}()
+
 	etcdClusterMode, err := serviceDeploymentMode(dockerCli.Client(), "amp.type.kv", "true")
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


cli/command/cluster/plugin.go
```diff
@@ -95,7 +95,7 @@ func killContainer(c cli.Interface, name string, sig string) error {
 func RunContainer(c cli.Interface, img string, dockerOpts docker, args []string, env map[string]string, f func(r io.Reader)) error {
 	containerName := fmt.Sprintf("amp-cluster-plugin-%s", stringid.GenerateNonCryptoID())
 	dockerArgs := []string{
-		"run", "-t", "--rm", "--name", containerName,
```
ndegory: why not deleting implicitly the container? Unless you make sure that you remove the container in all cases it's successfully exiting or is killed, you don't want to leave the container in a stopped state.

cluster/agent/cmd/install.go
```diff
@@ -96,14 +109,15 @@ func install(cmd *cobra.Command, args []string) error {
 			log.Println(f)
       
 			err := deployTest(dockerCli, f, "test", 60 /* timeout in seconds */)
-			stack.Remove(dockerCli, stack.RemoveOptions{Namespaces: []string{"test"}})
+			stack.Remove(dockerCli, stack.RemoveOptions{Namespaces: []string{namespace, TEST_NAMESPACE}})
```
ndegory: I don't think you want to remove the amp stack every time you remove the test stack.

cluster/plugin/local/plugin.go
```diff
@@ -179,8 +179,8 @@ func RunAgent(ctx context.Context, c *client.Client, action string, opts *Reques
 		if err != nil {
 			fmt.Printf("%v\n", err)
 		}
-		_ = removeAgent(ctx, c, r.ID)
-		done <- true
+		//_ = removeAgent(ctx, c, r.ID)
```
ndegory: ditto, why not removing the agent container?

ndegory (changes requested):

ndegory (approved):

PR #1635, opened 2017-08-10
Use pcregrep in smoke tests

neha-viswanathan:
Use `pcregrep` instead of `grep` for AMP CLI smoke tests

This PR also contains updated tests for `amp secret` 

#### To Test
Check Travis Build


ndegory (changes requested):
a few comments:
- there's no need to use pcregrep on simple greps (non regex ones), the simpler the better
- the issue didn't start in this PR, but a few tests here don't do anything of the tests, it just goes on with the next one, without checking the result, so they're useless as is

PR #1636, opened 2017-08-11
Deployment of core services starts delay and timer once the image is pulled

### LINKED ISSUE #1630: Depending on network reliability and speed, the AMP deployment may fail
The timeout definition in the AMP stack files (cluster/agent/stacks/) is based on the time spent between the service creation/update request and all tasks in the expected state. One significant part of this time is spent pulling the Docker image (when it's not already there), which brings a high variability depending on the host where it's running, depending on the performance of Docker Hub and depending on what has already been pulled.
Raising the timeout to excessively high values to avoid a `Error: context deadline exceeded` is not a satisfying solution, there should be a check on the status of the tasks to identify when the image has been pulled, to start waiting for the timeout.

ndegory:
Fix #1630 

- Start the stabilize delay and timeout once the image has been pulled
- Fails early if the image does not exist
- When a test stack was failing, ampagent waited for the timeout, which is a loss of time. Skipping this now
- Set stabilization time for test stacks to 0 second (a stopped task can't be more stable)

The deployment is not dependent anymore on the speed of the network, and won't fail because the pull of the image took longer than usual (up to a limit). This allow to reduce the stabilize delay and the stabilize timeout, which leads to a faster deployment on systems where the image is already pulled.

## Verification

Tested successfully on an AWS deployment (deploying a dev version on AWS requires a few steps that should be explained elsewhere).

To test locally:
```
$ make build-ampagent
$ amp cluster create
# remove a big image
$ docker image rm appcelerator/kibana:5.5.0
$ amp cluster create --local-fast
# remove a local image
$ docker image rm appcelerator/ampbeat:0.15.0-dev
$ amp cluster create --local-fast
# this last deployment should fail because the image can't be pulled
```

### DIFF
--- a/cluster/agent/pkg/docker/stack/deploy_composefile.go
+++ b/cluster/agent/pkg/docker/stack/deploy_composefile.go
@@ -315,6 +315,7 @@ func deployServices(
 		// service stabilization defaults
 		stabilizeDelay := time.Duration(5) * time.Second
 		stabilizeTimeout := time.Duration(1) * time.Minute
+		pullTimeout := time.Duration(1) * time.Minute
 
 		// override service stabilization default settings based on spec labels
 		labels := serviceSpec.TaskTemplate.ContainerSpec.Labels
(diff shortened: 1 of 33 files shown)


bquenin (approved):
LGTM

PR #1638, opened 2017-08-11
Manager self healing sometimes fail #1633

### LINKED ISSUE #1633: Manager self healing sometimes fail
On a long running AWS cluster, instances may fail. There's already a mechanism to recover from it (autoscaling group of 2N+1 instances, userdata with deterministic leader election), but there's some occurrences where it fails.

In this particular case, the `docker info` on the supposedly leader shows that the swarm is broken, and that an extra manager has joined.
```
docker info
...
Swarm: active
 NodeID: 21ij3hopy02tkac7ieh5bux1f
 Error: rpc error: code = 2 desc = The swarm does not have a leader. It's possible that too few managers are online. Make sure more than half of the managers are online.
 Is Manager: true
 Node Address: 192.168.0.89
 Manager Addresses:
  192.168.0.89:2377
  192.168.16.164:2377
  192.168.16.25:2377
  192.168.32.39:2377
...
```

The autoscaling group keeps launching a new instance to reach the target (3 nodes here), but it fails because the node can't join the swarm.

```
[   21.198981] cloud-init[1755]: waiting for quorum (3) (will timeout after 300 s)
[   24.515421] cloud-init[1755]: quorum reached in 3 s
[   24.524711] cloud-init[1755]: leader election (will timeout after 900 s)
[   27.564122] cloud-init[1755]: all manager nodes have an available Docker engine API (3 s)
[   27.685229] cloud-init[1755]: no established leader
[   27.698021] cloud-init[1755]: found a new leader: 192.168.0.89
[   27.713683] cloud-init[1755]: retrieving the swarm manager token (will timeout after 300 s)
[   27.729344] cloud-init[1755]: Error response from daemon: rpc error: code = 2 desc = The swarm does not have a leader. It's possible that too few managers are online. Make sure more than half of the managers are online.
...
[  325.988360] cloud-init[1755]: Error response from daemon: rpc error: code = 2 desc = The swarm does not have a leader. It's possible that too few managers are online. Make sure more than half of the managers are online.
[  328.008192] cloud-init[1755]: timeout
```

Hint: even when the instances are stable (no server recycling), the leader may change, so it's not clear here that `192.168.0.89` was still the leader.

nikolai-derzhak-distillery:
fixes #1633 

Added swarm down nodes cleanup to manager user-data init script.
That will ensure that if we lost one of managers new one (spun up by ASG) will be able to join swarm.
So we always will have quorum.

### How to verify it
Terminate of managers in AWS.  It will change state to Down/Unreachable in node ls output.
Auto scaling group will spin up new manager instance. 
During init process it will demote and remove down node from managers pool.
You should see it in cloud-init output in instance system console log.
```
[   30.016959] cloud-init[1749]: removing dead nodes from the Swarm
[   30.058367] cloud-init[1749]: removing ip-192-168-32-43 from the Swarm
[   30.108114] cloud-init[1749]: Manager ip-192-168-32-43 demoted in the swarm.
[   33.420135] cloud-init[1749]: ip-192-168-32-43
```

### DIFF
--- a/examples/clusters/userdata-aws-manager
+++ b/examples/clusters/userdata-aws-manager
@@ -319,6 +319,21 @@ _create_networks(){
   done
 }
 
+_swarm_cleanup(){
+  local _manager=$1
+  echo "removing dead nodes from the Swarm" >&2
+	# possible node statuses: https://github.com/moby/moby/blob/master/api/types/swarm/node.go#L106-L115
+  _down_nodes=$(docker -H "$_manager:2375" node ls --filter "role=manager" --format "{{.Hostname}} {{.Status}}"| grep -i down$ | cut -d' ' -f1)
+	# should we also check if corresponding instance is not running/pending in AWS ?
+  # like resolve hostname to IP and lookup instance state
+  # but assuming if node is down then it is not temporary state and we do not need extra checks
+	for _node in $_down_nodes; do
+	  echo "removing $_node from the Swarm" >&2
+		docker -H "$_manager:2375" node demote $_node
+		docker -H "$_manager:2375" node rm $_node
+	done 
+}
+
 _swarm_join(){
   local _manager=$1
   local _token
@@ -456,6 +471,7 @@ if [[ "x$nodeip" = "x$leader" ]]; then
   _swarm_init "$nodeip" || exit 1
   _create_networks $OVERLAY_NETWORKS || exit 1
 else
+	_swarm_cleanup "$leader" 
   _swarm_join "$leader" || exit 1
 fi
 _label_node || exit 1


ndegory:
best way to test it: run the standard Cloudformation template, and use `https://raw.githubusercontent.com/nderzhak/amp/1633/examples/cluster` as ConfigurationURL parameter.

ndegory:
when a manager node is terminated, it's seen as unreachable by the other manager nodes in less than a second, and it takes something like 15 sec to be marked as down, which is good enough since the new instances is available after several minutes (this includes the time it takes for the AWS to detect a missing instance, launch a new one, and the boot time for this new instance).

Another scenario, that is closer to what happens IRL, is to stop docker service on one manager node, to simulate an issue on the server. It takes several minutes for the internal ELB to identify it as out of service, which leads to the ASG marking it as unhealthy, terminating it and at the same time launching a new instance.
The new instance is booting while the other instance is not yet terminated (still shutting down).
Still, the timing is good, the new node is able to clean up the downed node before registering to the swarm.

ndegory (approved):
LGTM

PR #1640, opened 2017-08-14
amp version (local) shouldn't wait 10 sec to time out

ndegory:
`amp version` will give back the control in 1 sec when AMP is not deploy (and display the client version). Commit f974d094b introduced the 10 second timeout useful for remote connections, but had this side effect on local cluster commands.


bquenin (approved):
LGTM

PR #1645, opened 2017-08-18
Monitoring service discovery and host labels

### LINKED ISSUE #1641: Alerting and node identification
The alerting based on the metrics is a bit rough and sometimes gives information about a container hostname, which isn't really helpful. The hostname or IP of the node on which the container is running, and / or the task id of the container would be more useful.

In this example, a core worker node went down, which led to an alert on the node and an alert for the services running on it:
image

### LINKED ISSUE #1647: Monitoring service discovery
Instead of hardcoding the services that should be monitored by Prometheus, they should be discovered in the Swarm annotations.
This will allow to define Prometheus targets based on what is really deployed and configured for monitoring.

ndegory:
Fix #1641 (Alerting and node identification)

The prometheus configuration now includes more information, in particular labels with the name of the host and the task id of the container. This allows to present useful information in dashboards and alerts, instead of the IP:port of the container.

Fix #1647 (Monitoring service discovery)

Labels set in the stack files are watched by Prometheus to discover the services and build its configuration file.
Moving Prometheus to a manager node, since we need access to the Swarm annotations, this is a compromise. This means that the cloud deployment doesn't need an instance dedicated for Prometheus, one less instance group (of a single instance) is deployed.

New service relaying the docker engine metrics through a global container. The code is now more generic and is not specific to Docker for Mac.

Added tests on Prometheus configuration.

This also brings Docker 17.06.1-ce instead of Docker 17.06.0-ce.

## Verification

build the prometheus image ```make build-monit build-ampagent```
deploy amp ```amp cluster create```
Check the dashboard on dashboard.local.appcelerator.io, the instance name in the legends is now an IP instead of IP:port or container IP.

### DIFF
--- a/monitoring/promctl/main.go
+++ b/monitoring/promctl/main.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"net"
 	"os"
 	"os/exec"
 	"os/signal"
@@ -24,19 +23,19 @@ import (
 )
 
 const (
-	defaultTemplate         = "/etc/prometheus/prometheus.tpl"
-	defaultConfiguration    = "/etc/prometheus/prometheus.yml"
-	defaultHost             = docker.DefaultURL
-	defaultPeriod           = 1
-	dockerForMacIP          = "192.168.65.1"
-	dockerEngineMetricsPort = 9323
-	systemMetricsPort       = 9100 // node-exporter
-	prometheusCmd           = "/bin/prometheus"
-	monitoringNetwork       = "ampnet"
-	stackName               = "amp"
-	metricsPortLabel        = "io.amp.metrics.port"
-	metricsPathLabel        = "io.amp.metrics.path"
-	metricsModeLabel        = "io.amp.metrics.mode"
+	defaultTemplate      = "/etc/prometheus/prometheus.tpl"
+	defaultConfiguration = "/etc/prometheus/prometheus.yml"
+	defaultHost          = docker.DefaultURL
+	defaultPeriod        = 1
+	dockerForMacIP       = "192.168.65.1"
+	prometheusCmd        = "/bin/prometheus"
+	monitoringNetwork    = "ampnet"
+	stackName            = "amp"
+	metricsPortLabel     = "io.amp.metrics.port"
+	metricsPathLabel     = "io.amp.metrics.path"
+	metricsModeLabel     = "io.amp.metrics.mode"
+	metricsModeTasks     = "tasks"
+	metricsModeExporter  = "exporter"
 )
 
 var prometheusArgs = []string{
@@ -48,10 +47,7 @@ var prometheusArgs = []string{
 }
 
 type Inventory struct {
-	Jobs                    []Job
-	Hostnames               []string
-	DockerEngineMetricsPort int
-	SystemMetricsPort       int
+	Jobs []Job
 }
 
 type Job struct {
@@ -100,7 +96,8 @@ func prepareJobs(client *docker.Docker, networkResource types.NetworkResource) (
 		name := service.Spec.Annotations.Name
 		strMetricsPort, ok := service.Spec.Annotations.Labels[metricsPortLabel]
 		if !ok {
-			return nil, fmt.Errorf("unable to get metrics port label for service %s", name)
+			fmt.Printf("Warning: unable to get metrics port label for service %s, ignoring it\n", name)
+			continue
 		}
 		metricsPort, err := strconv.Atoi(strMetricsPort)
 		if err != nil {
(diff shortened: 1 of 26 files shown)


monitoring/promctl/main.go
```diff
@@ -41,31 +48,122 @@ var prometheusArgs = []string{
...
+			}
+			// all "tasks" jobs have the same relabel config
+			job.RelabelConfigs = append(job.RelabelConfigs,
+				RelabelConfig{SourceLabels: []string{"hostip"}, Separator: "@", TargetLabel: "instance"})
+			jobs = append(jobs, job)
+		case "exporter":
+			shortName := strings.TrimSuffix(strings.TrimPrefix(name, fmt.Sprintf("%s_", stackName)), "_exporter")
+			job := Job{Name: shortName, Mode: "exporter", MetricsPath: metricsPath}
+			job.StaticConfigs = []StaticConfig{
+				StaticConfig{
```
bquenin: Remove unnecessary type `StaticConfig` (from `make fmt`)

monitoring/promctl/main.go
```diff
@@ -41,31 +48,122 @@ var prometheusArgs = []string{
...
 	}
-	networkId := networkResources[0].ID
-	// when the vendors are updated to docker 17.06:
-	//networkResource, err := client.GetClient().NetworkInspect(context.Background(), networkId, types.NetworkInspectOptions{})
-	networkResource, err := client.GetClient().NetworkInspect(context.Background(), networkId, false)
+	for _, service := range services {
+		name := service.Spec.Annotations.Name
+		strMetricsPort, ok := service.Spec.Annotations.Labels[metricsPortLabel]
+		if !ok {
+			return nil, fmt.Errorf("unable to get metrics port label for service %s", name)
```
bquenin: Shouldn't you `continue` to the next service instead of` return`ing immediately?

monitoring/promctl/main.go
```diff
@@ -41,31 +48,122 @@ var prometheusArgs = []string{
...
+		if !ok {
+			metricsMode = "tasks"
+		} else if metricsMode != "exporter" && metricsMode != "tasks" {
+			fmt.Printf("Warning: wrong metrics mode (%s) for service %s, force it to tasks\n", metricsMode, name)
+			metricsMode = "tasks"
+		}
+		fmt.Printf("discovered service %s on port %d and path %s, mode %s\n", name, metricsPort, metricsPath, metricsMode)
+		s, ok := networkResource.Services[name]
+		if !ok {
+			return nil, fmt.Errorf("Warning: service %s not found in network %s\n", name, monitoringNetwork)
```
bquenin: Shouldn't you `continue` to the next service instead of` return`ing immediately?

monitoring/promctl/main.go
```diff
@@ -41,31 +48,122 @@ var prometheusArgs = []string{
...
+		if err != nil {
+			log.Printf("Warning: non numerical port for service %s: %s\n", name, strMetricsPort)
+			continue
+		}
+		metricsPath, ok := service.Spec.Annotations.Labels[metricsPathLabel]
+		if !ok {
+			metricsPath = "/metrics"
+		}
+		// mode can be tasks or exporter
+		metricsMode, ok := service.Spec.Annotations.Labels[metricsModeLabel]
```
bquenin, line 112: I think it would more readable to check for unexpected values in the `switch` below, using a `default` statement.

monitoring/promctl/main.go
```diff
@@ -41,31 +48,122 @@ var prometheusArgs = []string{
...
+			fmt.Printf("Warning: wrong metrics mode (%s) for service %s, force it to tasks\n", metricsMode, name)
+			metricsMode = "tasks"
+		}
+		fmt.Printf("discovered service %s on port %d and path %s, mode %s\n", name, metricsPort, metricsPath, metricsMode)
+		s, ok := networkResource.Services[name]
+		if !ok {
+			return nil, fmt.Errorf("Warning: service %s not found in network %s\n", name, monitoringNetwork)
+		}
+		switch metricsMode {
+		case "tasks":
```
bquenin: Defining constants for expected values (`tasks` and `exporter`) would be nice since you refer to these several times.

monitoring/promctl/main.go
```diff
@@ -74,20 +172,53 @@ func update(pid int, client *docker.Docker, configurationTemplate string, config
...
 	}
 	if len(hostnames) == 0 {
-		return errors.New("host list is empty")
+		return nil, errors.New("host list is empty")
 	}
-	inventory := &Inventory{Hostnames: hostnames, DockerEngineMetricsPort: dockerEngineMetricsPort, SystemMetricsPort: systemMetricsPort}
+	return hostnames, nil
+}
+
+func update(pid int, client *docker.Docker, configurationTemplate string, configuration string) error {
```
bquenin: `pid` parameter is not used

bquenin (changes requested):
This code is doing a lot of things and I think a few unit tests wouldn't hurt.
Also, it doesn't seem to work:
image

Here is my docker environment:
```
$ cat /etc/docker/daemon.json
{
	"dns": ["8.8.8.8","10.252.252.252"],
	"metrics-addr" : "0.0.0.0:9323",
	"experimental" : true,
	"storage-driver": "overlay2"
}
$ docker version
Client:
 Version:      17.06.1-ce
 API version:  1.30
 Go version:   go1.8.3
 Git commit:   874a737
 Built:        Thu Aug 17 22:51:12 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.06.1-ce
 API version:  1.30 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   874a737
 Built:        Thu Aug 17 22:50:04 2017
 OS/Arch:      linux/amd64
 Experimental: true
```

ndegory:
@bquenin not sure why you don't have the dashboards. Can you also screenshot your http://alerts.local.appcelerator.io/targets and http://alerts.local.appcelerator.io/config?

bquenin:
@ndegory here it is:
image
image

ndegory:
@bquenin did it ever work for you on a Linux workstation? It seems the issue is that the hostname is not resolvable from the Prometheus container. This is not an issue on Docker for {Mac,Windows}, because it's the VM IP, and not an issue on real swarm cluster because the IPs are accessible.

bquenin:
@ndegory it was working in the past but I haven't checked recently. It seems that the issue is coming from the way the config is computed:
```
- job_name: docker-engine
  scrape_interval: 15s
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
  static_configs:
  - targets:
    - u16-bbb127f9eae4:9323
```
I don't know where this comes from though, but my machine name is only `u16`, not `u16-bbb127f9eae4`.

bquenin (approved):

PR #1652, opened 2017-08-23
Exclusive lock on etcd replicas

### LINKED ISSUE #1642: etcd cluster instability
etcd is deployed with 3 replicas on 3 different nodes, which allows to lose a node and still get a functional cluster (raft consensus). Each replica uses a local Docker volume to persist its data.
When a worker node hosting an etcd replica disappears, Swarm launches a new task on one of the remaining nodes to reach the desired replica count. The new container is unable to start, since the data from the volume is already used by a running container.

The list of tasks of the etcd service after a worker node disappears. The missing node is the one with a node ID instead of a hostname. The new task shares the same host as an existing task.
```
$ docker service ps amp_etcd
ID                  NAME                IMAGE                      NODE                        DESIRED STATE       CURRENT STATE          ERROR               PORTS
10hvp87h5maq        amp_etcd.1          appcelerator/etcd:3.1.10   ip-192-168-0-52             Running             Running 37 hours ago
4lfes0v8xokt        amp_etcd.2          appcelerator/etcd:3.1.10   ip-192-168-0-52             Running             Running 2 hours ago
x6okutel1df2         \_ amp_etcd.2      appcelerator/etcd:3.1.10   9mindui4785i7ws9reoe67c75   Shutdown            Running 37 hours ago
ka8pyg28aima        amp_etcd.3          appcelerator/etcd:3.1.10   ip-192-168-16-228           Running             Running 37 hours ago
```

Logs from the new container.
```
$ docker logs <container_id>
2017-08-16 07:37:07.388689 W | etcdserver: another etcd process is running with the same data dir and holding the file lock.
2017-08-16 07:37:07.388712 W | etcdserver: waiting for it to exit before starting...
Error:  context deadline exceeded
```
Interestingly, this tasks doesn't error, and is waiting indefinitely. This doesn't allow Swarm to schedule it to another node.

Solution to be considered: 
- global service instead of replica count? That would put the constraint on the definition of the worker group, which is dangerous
- configure the deploy restart policy for the etcd service to make sure the new task is run after enough time to give time to the infrastructure to launch a new node. That's at best a guess and would reduce the risk, but won't fix it
- modify the appcelerator/etcd image to first check if another etcd container is running on same host, and exit if it's the case

ndegory:
- Fix #1642 (etcd cluster instability)
   Etcd waits indefinitely when the /data volume is already in use by another container.
   Adding a flock fails the container early, which allow Swarm to reschedule the container on another host, until it finds a host without an existing etcd task running.
- Bump etcd version to 3.2.6
- tune the restart policy in cluster mode


JosephGJ (approved):
LGTM

PR #1658, opened 2017-08-25
Fix metrics-addr Docker option for AWS clusters

ndegory:
follow-up of #1645 

on AWS deployments, the docker engines were configured to listen to the private IP, which is not compatible anymore with the IP used by the docker engine metrics exporter service (using the docker bridge interface).
This PR fixes the Docker metrics in the cluster dashboard.

Also adds security group rules to open direct access to the dashboard (without the haproxy vhost)


bquenin (approved):
LGTM

PR #1671, opened 2017-09-05
make clean will also remove the images

ndegory:
`(amp)make clean` used to remove the binaries, but the Docker images were left intact.
The Docker image is a binary artifact too and so should be removed.
The other advantage is that Docker will remove all intermediate image used for the build, hence reducing any cache side effects at built time.

## Verification

```
ampmake clean build
```


bquenin (approved):
LGTM

PR #1674, opened 2017-09-06
Resources limits on core services

ndegory:
- Now that the Prometheus task runs on a manager node, we want to make sure that it doesn't swallow all the resources of the node, which would be detrimental to the Swarm.
  For the same reason, a few services such as the exporters and the agent shouldn't be allowed to use too much memory and cpu, it's safe to allocate a low value for them.
  Adding reservations is also a good way to make sure Swarm knows how to schedule them efficiently.
- When Swarm is unable to allocate the tasks to a worker node, ampagent will fail quickly instead of waiting for the pull timeout, which was the case before.
- Logs from the ampagent used to show a few duplicates, it is now fixed.
- `amp logs` won't display the cluster deployment logs anymore (unless `-i` option is used).
- better tests: more information when it fails, and removed an infinite loop

## Verification
local deployment:
```
amp cluster create
# verify the dashboard are available on http://dashboard.local.appcelerator.io/
amp cluster rm
```
aws deployment:
```
amp cluster create --provider=aws --aws-region us-west-2 --aws-stackname STACK_NAME --aws-parameter KeyName=KEYPAIR_NAME --aws-sync
# verify the dashboards from the MetricsURL output
amp cluster rm --provider aws --aws-region us-west-2 --aws-stackname STACK_NAME
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ script:
   - ampmake clean build-base build-local-plugin build-cli
   - ls -la bin/linux/amd64
   - amp cluster create --local-fast
-  - TESTINCLUDE=platform/testing testrunner platform/tests
+  - PRINT_PROGRESS=1 TESTINCLUDE=platform/testing testrunner platform/tests
   - docker build -t appcelerator/amp-integration .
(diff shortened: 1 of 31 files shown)


ndegory:
not a priority anymore, closing for now

PR #1675, opened 2017-09-07
Fix network removal in amp cluster rm

### LINKED ISSUE #1670: amp cluster rm sometimes fails to remove network or volumes
Because the service deletion is asynchronous, when ampagent tries to delete the network (and sometimes the volumes), an error may occur.

```
Error: Error response from daemon: rpc error: code = FailedPrecondition desc = network iru82dgb7v0voleym7xt0twbs is in use by service p2i5m8m5e7jeei12knm25fww7
```

These deletions should be retried a few times until it works.

ndegory:
Fix #1670

### DIFF
--- a/cluster/agent/cmd/install.go
+++ b/cluster/agent/cmd/install.go
@@ -426,8 +426,18 @@ func RemoveNetwork(name string) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


ndegory:
not a priority anymore, closing for now

Issue #1694, opened 2017-10-06
Use Docker multistage builds


bquenin:
The idea is to evaluate multistage builds vs builder pattern (what we have currently).

The only thing I'm concerned about with multistage build is the fact that it's probably going to make the build non-incremental (i.e. rebuild everything, every time, regardless of actual changes). Indeed, the source code needs to be copied into the Docker image in order to be built. Today, we mount the source code when executing `ampmake` or we use `make` which ensure we rebuild only what is needed.

I'll go ahead and implement MS builds on a branch so we can take an informed decision.

bquenin:
Closing for now as it doesn't seem to improve anything in our situation.

PR #1695, opened 2017-10-07
Reintegrate cluster binaries to top-level project

### LINKED ISSUE #1691: Reintegrate `cluster` binaries to top-level project

bquenin:
Fix #1691 

## How to test
```
ampmake clean build
amp cluster create
```


Makefile
```diff
@@ -75,8 +75,7 @@ clean: clean-protoc cleanall-cli clean-server clean-beat clean-agent clean-monit
 DOCKER_CMD ?= "docker"
 
 build-base: protoc build-server build-gateway build-beat build-agent build-monit
-build-local-plugin: build-amp-local build-ampagent
```
ndegory, line 78: this target was used by the Travis build, no need to build the aws plugin (and soon the azure plugin, etc) when we can only test locally.

bquenin: even if we're not actually using them, it's still good to make sure they actually build

ndegory: sure, but the build time is increasing significantly

ndegory (changes requested):
```
make VERSION=latest build-ampagent
make build-ampagent
docker run --rm appcelerator/ampagent:latest --help | grep version
2017/10/07 00:22:23 ampctl (version: latest, build: 18ddddef)
# this is OK
docker run --rm appcelerator/ampagent:0.16.0-dev --help | grep version
2017/10/07 00:22:45 ampctl (version: latest, build: 18ddddef)
# this is not OK

On master:
make build-amp-agent
docker run --rm appcelerator/ampagent:0.16.0-dev --help | grep version
2017/10/07 00:24:36 ampctl (version: 0.16.0-dev, build: 68d5f96f)
```
we have to make sure that the build of the binary is done with the right tag, and not from the cache.

ndegory:
sorry for the header style in the review, it was supposed to be a commented line...

ndegory (approved):

ndegory:
still the issue with the `make VERSION=latest build-ampagent`, but that should be fixed when the multi staged build will be implemented

bquenin:
For reference #1694

PR #1687, opened 2017-10-06
Use dep as vendoring tool

bquenin:
Fix #1689
Fix #1690 

## How to test
```
ampmake clean build
```

PR #1696, opened 2017-10-07
Optimize plugins image size

ndegory:
#1687 had a side effect on the plugin images: they grew larger, the binaries are bigger because of the merged vendors. That means that the binaries size can be optimized.

The addition of the _f_ and _w_ flags in the default LDFLAGS helps to reduce the size of most of the images by half (amplifier was 54M, is 25M now).

```
for i in ampagent amp-local amp-aws; do docker image ls --format '{{.Repository}} {{.Size}}' appcelerator/${i}:0.16.0-dev ; done
make build-ampagent build-amp-local build-amp-aws
for i in ampagent amp-local amp-aws; do docker image ls --format '{{.Repository}} {{.Size}}' appcelerator/${i}:0.16.0-dev ; done
```
Size of ampagent should reduce significantly.


ndegory:
strip should not be used, according to [the moby recommendations for go binaries](https://github.com/moby/moby/blob/2a95488f7843a773de2b541a47d9b971a635bfff/project/PACKAGERS.md#stripping-binaries).

bquenin:
It looks like it's safe to proceed with linker flags `-s -w`. It preserves the annotations required for stack traces.
As you mentioned, `upx` is not encouraged :)

ndegory:
@bquenin, actually `strip` is discouraged, but `upx` is encouraged. It's not removing content, it's just compressing the binary.

bquenin:
@ndegory I wouldn't say that upx is encouraged. There's still an ongoing issue golang/go#16706 showing that upx support varies from version to version. I'm not saying we shouldn't use upx, but I want to be sure it's not going to introduce some weird side effects.

ndegory:
np, I'll remove it

bquenin (approved):
LGTM (tested on Linux)

PR #1703, opened 2017-10-11
AMI update with Docker 17.09

ndegory:
cloudformation template updated with new AMI (patched and with Docker 17.09, was 17.06.1 before).


bquenin (approved):
LGTM

PR #1720, opened 2017-10-13
Stackfiles by feature

bquenin:
This PR groups services by feature into stack files:
- Core stack
- Logs stack
- Metrics stack
- Proxy stack

## How to test
```
$ ampmake clean build
$ amp cluster create
```
Check travis build


neha-viswanathan:
Can you add `Deploying stack <stackname>` in the output for clarity?

bquenin:
Done.

neha-viswanathan (approved):

cluster/ampagent/cmd/install.go
```diff
@@ -148,10 +137,11 @@ func serviceDeploymentMode(c docker.APIClient, labelKey string, labelValue strin
 }
 
 // returns sorted list of yaml file pathnames
-func getStackFiles(path string, clusterMode map[string]string) ([]string, error) {
+func getStackFiles(path string, deploymentMode string) ([]string, error) {
 	if path == "" {
 		path = "./stacks"
 	}
+	path += "/" + deploymentMode
```
ndegory, line 144: a consequence of the change is that all stack files are duplicated, even when there's no reason to duplicate it (the proxy stack is the same whatever the cluster mode). That means that if somebody forgot to report a change on both files, we'll miss the difference during the tests. Why not leveraging the _include_ syntax to load service definitions from a common file? That should also be done on the stacks that are different based on the cluster mode, but contains many services that are not different (in the core stack only etcd is dependent on the cluster mode).

bquenin: Sounds good! Please open an issue.

PR #1729, opened 2017-10-19
Remove references to stats or metrics

### LINKED ISSUE #1728: Remove stats from agent/ampbeat/kibana
Since we're using Prometheus to collect containers metrics, it makes no sense to collect them through our existing chain anymore. Therefore, the stats collection needs to be removed from:

- `agent`
- `ampbeat`
- `kibana`

Also, the `amp stats` command needs to be disabled for now but not removed. The `stats` API needs to be disabled as well.
We will restore both the command line and the API in a future PR when we plug it to the Prometheus API.

neha-viswanathan:
Closes #1728 

#### To test

Check Travis build.

Also check Kibana at http://kibana.local.appcelerator.io/ for only logs data.


bquenin (approved):
LGTM

PR #1730, opened 2017-10-19
Container metrics

### LINKED ISSUE #1727: Add Docker container metrics
Metrics about containers (memory, cpu, etc) were gathered when the AMP metrics were based on InfluxDB / Telegraf, and later were also included in Elasticsearch through the amp agent.
Having them in Prometheus with the other metrics would be helpful to build relevant dashboards.

ndegory:
Fix #1727 
- new global service gathering the metrics from the containers
- new dashboard with memory, cpu and IO usage, with filters on service, stack and node. The dashboard is based on https://grafana.com/dashboards/179.

image

PS: for now it's using a custom made cadvisor image, there's a bug in the official one that breaks the service on docker engines running with overlay2.


bquenin (approved):
LGTM (tested on linux)

PR #1733, opened 2017-10-25
Remove build public key from AMI

### LINKED ISSUE #1663: AWS the public key used to build the AMI should be deleted
The key used to build the AMI is still in the AMI.

ndegory:
- Fix #1663 (clean up authorized_keys)
- Remove obsolete parameter (DockerChannel)
- AMI build is now forcing the stable channel for docker installation

CloudFormation template is already pushed on s3.

## How to test

Deploy the cloudformation template from the AWS console, by uploading the local template.
Once the cluster is deployed, check the dashboards (URL in the stack output) to verify that you have the expected number of node, all running on Docker 17.09. SSH to one instance and check that the authorized keys of the ubuntu user only contains your public key.

### DIFF
--- a/examples/clusters/aws-swarm-asg.yml
+++ b/examples/clusters/aws-swarm-asg.yml
@@ -8,27 +8,27 @@ Mappings:
     # N Virginia
     us-east-1:
       Ubuntu: ami-e4139df2
-      Default: ami-7ea06904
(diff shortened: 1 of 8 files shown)


bquenin (approved):
LGTM

PR #1739, opened 2017-10-26
NFS endpoint

ndegory:
- new option in the cloudformation template for an EFS filesystem. It's then available to the full VPC and can be mounted directly by a container
- sample stack with a nginx container mounting the EFS FS and sharing the files on the replicas

## How to test
```
$ amp cluster create --provider aws --aws-region us-west-2 --aws-stackname nde-efs --aws-parameter KeyName=xxx --aws-parameter UserWorkerSize=2 --aws-parameter CoreWorkerSize=1 --aws-parameter ManagerSize=1 --aws-parameter NFSEndpoint=true --aws-sync
# take note of the output of the stack, in particular the DNSTarget and the NFSEndpoint
# update your /etc/hosts or update your dns so that you have an alias to DNSTarget (you should have something like youralias.yourdomain and youralias.nfs.examples.youralias.yourdomain resolving to DNSTarget)
# cat examples/stacks/nfs/README.md
# follow the instructions in the readme to try the sample
```

### DIFF
--- a/cluster/plugin/aws/main.go
+++ b/cluster/plugin/aws/main.go
@@ -54,6 +54,7 @@ func scanEvents(mode string) error {
 		"AWS::EC2::VPC":                      true,
 		"AWS::CloudFormation::WaitCondition": true,
 		"AWS::AutoScaling::AutoScalingGroup": true,
+		"AWS::EFS::FileSystem":               true,
 	}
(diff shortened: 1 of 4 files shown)


bquenin (approved):
LGTM

PR #1754, opened 2017-11-11
Network isolation

ndegory:
https://techweb.axway.com/jira/browse/AMP-49

- replace ampnet by 3 default overlay networks: core, monit and public
- update of stack samples to show how to use these default networks and stack dedicated overlay networks


bquenin (approved):
LGTM

PR #1762, opened 2017-11-17
Adding pagination to the logs API

bquenin:
This PR adds pagination support to the logs API.
The underlying implementation relies on ES SearchAfter feature described here: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-after.html

The `Get` method in the log API now supports a `from` parameter. This parameter can be set based on a previous reply from the API. Indeed, each log entry now returns a FromID field which can be used as a starting point for subsequent queries.

## How to test
Check Travis build.
Integration tests have been updated accordingly in `tests/integration/logs/logs_test.go`.


ndegory (approved):
LGTM

PR #1766, opened 2017-11-21
Implement cluster status and ls command in the API

ndegory:
Fix AMP-50

amp cluster status was implemented in the provider plugin (and not in all plugins), it is now part of the API. It also now contains more information about the context of the cluster: provider, region, output from the stack.

amp cluster ls and amp cluster node ls implementations were in the provider plugins and were not completed. amp cluster node ls is now part of the API, and provide extra information compared to what docker node ls provides: Version of the docker engine, available CPU and memory. It can also filter by labels (engine or node).

This PR also include vendor updates necessary for this PR and the following one (object store API).

## Verification
```
ampmake build
# deploy a cluster on AWS, signup and login
amp -k -s CLUSTER_URL cluster status
# tells if it's deployed on AWS or any other cloud provider, the name of the cluster, the region (when it applies), and stack output data when available
amp -k -s CLUSTER_URL cluster node ls
# should show the full list of nodes, with their version, cpu and memory.
amp -k -s CLUSTER_URL cluster node ls --label amp.type.user
# only show the list of nodes with this label
amp -k -s CLUSTER_URL cluster node inspect ANY_NODE_ID
# shows detailed information about this node, including its labels and plugins
```


cli/command/cluster/nodeList.go
```diff
@@ -51,14 +75,52 @@ func nodeList(c cli.Interface) error {
...
+		return nil
+	}
 	w := tabwriter.NewWriter(os.Stdout, 0, 8, 2, ' ', 0)
-	fmt.Fprintln(w, "ID\tHOSTNAME\tSTATUS\tAVAILABILITY\tMANAGER STATUS")
-	for _, node := range reply.Nodes {
-		leader := ""
-		if node.ManagerLeader {
-			leader = "Leader"
+	if len(reply.Nodes) == 1 {
+		fmt.Println("Found a single node, showing detailed information")
```
bquenin: This should not be printed on the output. Put it in the command help.

cli/command/cluster/nodeList.go
```diff
@@ -51,14 +75,52 @@ func nodeList(c cli.Interface) error {
...
+		fmt.Fprintf(w, "Engine Labels\n")
+		for k, v := range node.EngineLabels {
+			fmt.Fprintf(w, "\t- %s=%s\n", k, v)
+		}
+		fmt.Fprintf(w, "Engine Plugins\n")
+		for _, p := range node.EnginePlugins {
+			fmt.Fprintf(w, "\t- Type = %s, Name = %s\n", p.Type, p.Name)
+		}
+	} else {
+		fmt.Println("Found multiple nodes, for detailed information about a particular node, you can filter with the --id or --name option")
```
bquenin: This should not be printed on the output. Put it in the command help.

bquenin (changes requested):
It works but it feels that we have two commands combined into one.
I think it would be better to add another command such as `amp cluster node inspect` or `amp cluster node info` that would take a node ID as a parameter and displays details about the node. That would also be more consistent with the existing UX.

ndegory:
right, a node inspect command would make sense.

bquenin (approved):

PR #1768, opened 2017-11-21
local plugin: check for exited ampagent containers

ndegory:
Fix AMP-69 (can't run amp cluster rm (ampagent already in use))

the local plugin now checks if an ampagent container already exists. If it's stopped, it removes it and continue its business. If it's running, it prints a useful error for the user.

## Verification

```
make build-plugin-local
docker run --name ampagent alpine
amp cluster create
# it should remove the stopped ampagent container (it will be displayed), and continue with the cluster creation
```


cluster/plugin/local/plugin/plugin.go
```diff
@@ -118,7 +118,12 @@ func containerLogs(ctx context.Context, c *docker.Client, id string) {
 
 // RunAgent runs the ampagent image to init (action ="install") or destroy (action="uninstall")
 func RunAgent(ctx context.Context, c *docker.Client, action string, opts *RequestOptions) error {
-	containerName := ContainerName
+	// first remove any exited or dead container with same name
+	// this can happen when Docker crashes while the plugin and the agent are running
+	err := removeExitedAgent(c)
```
bquenin: `if err := removeExitedAgent(c); err != nil {`

bquenin (changes requested):
LGTM. Tested on Linux.

bquenin (approved):

PR #1769, opened 2017-11-21
AMI rebuild, fix race condition on networks with system prune

ndegory:
- Fix amp-78 (amp cluster deploy failure because of missing overlay network)
- Updated AMI


bquenin (approved):
LGTM

PR #1770, opened 2017-11-22
AWS deployment improvements

ndegory:
- Fix amp-73 (warning when credentials are not provided)
- Fix amp-76 (allow credentials to be passed as arguments)
- region can be read from the .aws/config file
- option --aws-sync now defaults to true
- new --aws-profile option, for credentials file with multiple profiles
- stacks in ROLLBACK_COMPLETE status were ignored by the rm command

The only way to provide credentials was to create a .aws/credentials file. It's still the recommended way to do it. We now have an alternative way to pass the credentials: with the --aws-access-key-id and --aws-secret-access-key.

### DIFF
--- a/cli/command/cluster/create.go
+++ b/cli/command/cluster/create.go
@@ -38,6 +38,8 @@ func NewCreateCommand(c cli.Interface) *cobra.Command {
 	flags.String("aws-stackname", "", "The name of the AWS stack that will be created")
 	flags.Bool("aws-sync", false, "If true, block until the command finishes (default: false)")
 	flags.String("aws-template", "", "UNSUPPORTED: cloud formation template url")
+	flags.String("aws-access-key-id", "", "aws credential: access key id")
+	flags.String("aws-secret-access-key", "", "aws credential: secret access key")
 
 	return cmd
 }
--- a/cli/command/cluster/create.go
+++ b/cli/command/cluster/create.go
@@ -40,6 +40,7 @@ func NewCreateCommand(c cli.Interface) *cobra.Command {
 	flags.String("aws-template", "", "UNSUPPORTED: cloud formation template url")
 	flags.String("aws-access-key-id", "", "aws credential: access key id")
 	flags.String("aws-secret-access-key", "", "aws credential: secret access key")
+	flags.String("aws-profile", "default", "aws credential: profile")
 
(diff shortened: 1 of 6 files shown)


cli/command/cluster/create.go
```diff
@@ -36,7 +36,7 @@ func NewCreateCommand(c cli.Interface) *cobra.Command {
 	flags.StringSlice("aws-parameter", []string{}, "Key-value pairs to pass through to the AWS CloudFormation template")
 	flags.String("aws-region", "", "The region to use when launching the instance")
 	flags.String("aws-stackname", "", "The name of the AWS stack that will be created")
-	flags.Bool("aws-sync", false, "If true, block until the command finishes (default: false)")
+	flags.Bool("aws-sync", true, "If true, block until the command finishes")
```
neha-viswanathan, line 39: Adding `(default: true)` to the help can be useful

cli/command/cluster/remove.go
```diff
@@ -28,7 +28,7 @@ func NewRemoveCommand(c cli.Interface) *cobra.Command {
 	// aws options
 	flags.String("aws-region", "", "Region to use to delete the instance")
 	flags.String("aws-stackname", "", "Name of the AWS stack to be deleted")
-	flags.Bool("aws-sync", false, "If true, block until the command finishes (default: false)")
+	flags.Bool("aws-sync", true, "If true, block until the command finishes")
```
neha-viswanathan, line 31: Adding `(default: true)` to the help can be useful

neha-viswanathan (changes requested):
Can you please update the PR description with test steps?

cli/command/cluster/create.go
ndegory: did you compile and run the help? It's displayed for me:
```      --aws-sync                       If true, block until the command finishes (default true)```

cli/command/cluster/remove.go
ndegory: ditto

cli/command/cluster/create.go
neha-viswanathan: yeah just checked 👍

neha-viswanathan (approved):
LGTM

PR #1771, opened 2017-11-28
Prometheus alerting rules updates

ndegory:
- Fix amp-86 (link to prometheus in alertmanager alerts)
- Fix amp-87 (migrated sample alert rules file for Prometheus 2.0)
- Fix amp-90 (--format option for service inspect command)
- Bump alertmanager version to 0.11.0

## Verification

```
make build-cli build-ampagent
docker config rm prometheus_alerts_rules
docker config create prometheus_alerts_rules examples/monitoring/prometheus_alerts.rules
amp cluster create
amp -k user signup
# check amp-87
amp -k service logs -i amp_prometheus
# there should be no error related to the alerting rules
# check amp-90 (there's also a unit test for this one)
amp -k service inspect amp_prometheus --format '{{ (index .Spec.TaskTemplate.ContainerSpec.Configs 0 ).ConfigName }}'
# check amp-86
# go to prometheus: alerts.local.appcelerator.io
# wait for an alert to fire (you will have to trigger it, that's an exercice for the reviewer)
```
image
```
# go to alertmanager: alertmanager.local.appcelerator.io
# the alert should be there, check the link to the Prometheus alert (Source), it should point to Prometheus.
```
image

### DIFF
--- a/monitoring/promctl/main.go
+++ b/monitoring/promctl/main.go
@@ -36,6 +36,8 @@ const (
 	metricsModeLabel     = "io.amp.metrics.mode"
 	metricsModeTasks     = "tasks"
 	metricsModeExporter  = "exporter"
+	externalURLEnv       = "PROMETHEUS_EXTERNAL_URL"
+	externalURLOption    = "--web.external-url"
 )
 
 var prometheusArgs = []string{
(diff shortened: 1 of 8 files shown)


monitoring/promctl/main.go
```diff
@@ -299,6 +310,9 @@ func main() {
 	RootCmd.PersistentFlags().StringVar(&host, "host", defaultHost, "host")
 	RootCmd.PersistentFlags().Int32VarP(&period, "period", "p", defaultPeriod, "reload period in minute")
 
+	if err := addExternalURL(&prometheusArgs); err != nil {
```
bquenin: Since your function `addExternalURL` doesn't return an error, it seems unnecessary to have a function here. Something like this should be enough:
```
// Set Prometheus external URL if provided
	if url := os.Getenv(externalURLEnv); url != "" {
		prometheusArgs = append(prometheusArgs, fmt.Sprintf("%s=%s", externalURLOption, url))
	}
```

bquenin (changes requested):
LGTM. Minor comments.

ndegory:
@bquenin, I've updated with your suggestion, thanks.

bquenin (approved):

PR #1775, opened 2017-11-30
Cleanup the email notification for user verification

ndegory:
cleanup of the user verification email html code.

## Verification

```
make build
amp cluster create --provider aws --aws-stackname STACK_NAME
# enable sendgrid on the remote cluster
amp -k -s REMOTE_URL user signup --name LOGIN --email VALID_EMAIL --password RANDOM_STRING
# check the email notification
```
image
(ignore the insecure part in the screenshot, it's not in the PR anymore)

### DIFF
--- a/pkg/mail/accountVerification.go
+++ b/pkg/mail/accountVerification.go
@@ -3,20 +3,38 @@ package mail
 var accountVerificationBody = `
 <!DOCTYPE html>
 <html>
-    <body style="background-color:white">
-        <div style="color:#404040;">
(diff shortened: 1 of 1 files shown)


bquenin:
I don't think it's valuable to change the message in the template as this is not the expected behavior. A production environment should have a valid certificate and therefore doesn't require this clarification. It leaves the development/trial use case and I think it would be better to explain this in the documentation.

ndegory:
ok, hint removed from the template.

ndegory:
closing for now, not planned in the current sprint

PR #1777, opened 2017-12-01
CLI defaults to 127.0.0.1

ndegory:
Fix AMP-116

On recent redhat systems, the CLI is unable to reach amplifier with localhost:50101.
Switching the default to 127.0.0.1 works.


bquenin (approved):
LGTM

PR #1780, opened 2017-12-05
Alpine 3.7 and images update

ndegory:
Fix amp-112

- elastic 6.0.0
- etcd 3.2.11
- cadvisor 0.28.3
- node-exporter 0.15.2
- use fixed tag instead of latest when possible
- all images based on Alpine rebased to Alpine 3.7


bquenin (approved):
LGTM

PR #1787, opened 2017-12-14
[WIP] Options to turn proxy off for AWS deployment

bquenin:
https://techweb.axway.com/jira/browse/AMP-64

This PR provides options to disable Proxy deployment for AWS cluster deployments.

## How To Test
```
$ amp cluster create --no-proxy --provider aws --aws-region $REGION --aws-stackname $STACKNAME --aws-parameter KeyName=$KEYNAME --aws-parameter UserWorkerSize=1 --aws-parameter CoreWorkerSize=1 --aws-parameter ManagerSize=1 --aws-sync
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ script:
   - sudo modprobe ip_vs ip_vs_rr
   - docker swarm init --advertise-addr 127.0.0.1
   - docker secret create amplifier_yml tests/test.amplifier.yml
-  - amp cluster create --local-no-metrics
+  - amp cluster create
 # Tests: CLI
   - sudo apt-get install pcregrep
   - sudo mkdir -p $HOME/.config/amp
--- a/cluster/ampagent/main.go
+++ b/cluster/ampagent/main.go
@@ -34,15 +34,19 @@ func main() {
 	rootCmd.AddCommand(cmd.NewInstallCommand())
 	rootCmd.AddCommand(cmd.NewUninstallCommand())
 
-	// These flags pertain to install, but need to be enabled here at root and persist for when it is invoked with no subcommand
-	rootCmd.PersistentFlags().BoolVar(&cmd.InstallOpts.NoLogs, "no-logs", false, "Don't deploy logs stack")
-	rootCmd.PersistentFlags().BoolVar(&cmd.InstallOpts.NoMetrics, "no-metrics", false, "Don't deploy metrics stack")
-	rootCmd.PersistentFlags().BoolVar(&cmd.InstallOpts.NoProxy, "no-proxy", false, "Don't deploy proxy stack")
-
-	// Environment variables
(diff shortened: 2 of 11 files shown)


.travis.yml
```diff
@@ -22,7 +22,7 @@ script:
   - sudo modprobe ip_vs ip_vs_rr
   - docker swarm init --advertise-addr 127.0.0.1
   - docker secret create amplifier_yml tests/test.amplifier.yml
-  - amp cluster create --local-no-metrics
+  - amp cluster create --no-metrics
```
ndegory: btw, it's cool that it's deploying faster for tests, but that means we miss any regression on the metrics layer.
I think we should include it.

examples/clusters/aws-swarm-asg.yml
```diff
@@ -229,6 +229,27 @@ Parameters:
     Default: 8080
     MinValue: 1025
     MaxValue: 32768
+  NoLogs:
```
ndegory, line 232: I would prefer a more generic approach, that won't require changing the template and rebuild AMI next time we add a sub stack (I have a branch with a new sub stack for node management). A parameter such as "ApplicationEnv" for instance, in which we'll have something like "NO_METRICS=true REGISTRATION=true NOTIFICATIONS=email". We'll have to parse it in the userdata, but at least it would be generic.

examples/clusters/userdata-aws-manager
```diff
@@ -20,6 +20,11 @@ AMPAGENT_VERSION=${APP_VERSION:-latest}
 SYSTEMD_DOCKER_OVERRIDE=/etc/systemd/system/docker.service.d/docker.conf
 SYSV_DOCKER_DEFAULT=/etc/default/docker
 
+export TAG=${APP_VERSION:-latest}
```
ndegory: not sure about setting the TAG variable to the tag of the ampagent image. It will break the way ampagent sets the environment for core stack files (when tag is not set, it uses the Version var, which is something like 0.18.0-dev on dev versions, 0.18.0 on released versions, and latest on dev builds for cloud deployments (make VERSION=latest clean-ampagent build-ampagent).

ndegory (changes requested):

examples/clusters/aws-swarm-asg.yml
bquenin: While I don't disagree fundamentally, I think this can be part of a separate PR

cluster/ampagent/main.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/appcelerator/amp/cluster/ampagent/cmd"
 	"github.com/spf13/cobra"
+	"fmt"
```
ndegory: cluster/ampagent/main.go:9:2: imported and not used: "fmt"

PR #1789, opened 2017-12-14
HA consul example for APIRS

ndegory:
Fix AMP-135

Consul cluster with 3 members, with local data persistency.
Consul by default runs in dev mode without persistency. Changing the command to set it to production mode (allowing persistency in the mounted volume). HA is implemented with 3 services (member-a, member-b and member-c) with a single replica each but sharing the same service alias (consul). Each member has a dedicated volume on the NFS server.

Updated sample dashboards.

image


ndegory:
superseded by #1802

PR #1791, opened 2017-12-19
[tls.1] Fix Docker client pkg for remote connections

ndegory:
This PR is required for _Protected socket on manager nodes_

- fixes in pkg/docker package to allow connection with TLS options. This will be used in a subsequent PR by amplifier and prometheus.

## Verification

automated build

### DIFF
--- a/cluster/ampagent/Dockerfile
+++ b/cluster/ampagent/Dockerfile
@@ -1,4 +1,4 @@
-FROM scratch
+FROM appcelerator/alpine:3.7.0
 
 COPY defaults /defaults
 COPY stacks /stacks
--- a/cluster/ampagent/cmd/checks.go
+++ b/cluster/ampagent/cmd/checks.go
@@ -1,12 +1,9 @@
 package cmd
 
 import (
+	"fmt"
 	"log"
 
-	"fmt"
-
-	"docker.io/go-docker"
-	"docker.io/go-docker/api"
 	"docker.io/go-docker/api/types"
 	"docker.io/go-docker/api/types/versions"
 	ampdocker "github.com/appcelerator/amp/pkg/docker"
@@ -42,6 +39,10 @@ func Checks(cmd *cobra.Command, args []string) error {
 	if !checksOpts.version && !checksOpts.labels && !checksOpts.scheduling {
 		checksOpts.all = true
 	}
+	Docker = ampdocker.NewEnvClient()
+	if err := Docker.Connect(); err != nil {
+		return err
+	}
 	if checksOpts.version || checksOpts.all {
 		if err := VerifyDockerVersion(); err != nil {
 			log.Println("Version test: FAIL")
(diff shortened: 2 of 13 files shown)


cluster/ampagent/cmd/install.go
```diff
@@ -36,7 +36,7 @@ type InstallOptions struct {
 }
 
 var InstallOpts = &InstallOptions{}
-var Docker = ampdocker.NewClient(ampdocker.DefaultURL, ampdocker.DefaultVersion)
+var Docker *ampdocker.Docker
```
bquenin: Removing the initialization of the `Docker` global variable and then initializing it each time it's referenced is not great. It would be better to keep the initialization in a single place and avoid copy/pasting.

cluster/ampagent/cmd/checks.go
```diff
@@ -42,6 +39,10 @@ func Checks(cmd *cobra.Command, args []string) error {
 	if !checksOpts.version && !checksOpts.labels && !checksOpts.scheduling {
 		checksOpts.all = true
 	}
+	Docker = ampdocker.NewEnvClient()
```
bquenin: here

cluster/ampagent/cmd/install.go
```diff
@@ -49,6 +49,7 @@ func NewInstallCommand() *cobra.Command {
 }
 
 func Install(cmd *cobra.Command, args []string) error {
+	Docker = ampdocker.NewEnvClient()
```
bquenin: here

cluster/ampagent/cmd/uninstall.go
```diff
@@ -17,6 +17,7 @@ func NewUninstallCommand() *cobra.Command {
 }
 
 func Uninstall(cmd *cobra.Command, args []string) error {
+	Docker = docker.NewEnvClient()
```
bquenin: here

cluster/ampagent/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM scratch
+FROM appcelerator/alpine:3.7.0
```
bquenin, line 1: Why?

bquenin (changes requested):
How to test?

cluster/ampagent/Dockerfile
ndegory: TLS won't work without the ca-certificates package, which is installed in appcelerator/alpine:3.7.0.

cluster/ampagent/cmd/install.go
ndegory: right, I'll move the initialization back to the global variable assignment.

ndegory:
For this PR, test is done with the automated test in Travis, since we're not using the TLS connection yet (part of another PR since it has been cut in smaller pieces).

cluster/ampagent/Dockerfile
bquenin: ok, got it

ndegory:
@bquenin does the new commit answer the changes you requested?

bquenin (approved):

ndegory:
superseded by #1802

PR #1798, opened 2018-01-05
Enhance cloud provider detection

ndegory:
Detection of cloud provider based on /sys/hypervision/uuid may fail (it's maybe related to ENA). In this case, falling back to /sys/devices/virtual/dmi/id/product_uuid provides the information.

## Verification

Deploy a cluster on AWS and deploy locally, and run the amp cluster status command to make sure the Provider is correctly identified.


ndegory:
superseded by #1802

PR #1801, opened 2018-01-11
Network Load Balancers

ndegory:
Replacement of classic ELB with network LB for the manager nodes (internal and external).
NLB bring a few advantages:
- less expensive
- low latency
- high throughput
- static IPs
- zonal isolation

## Verification
- Deploy a cluster on AWS (https://github.com/appcelerator/amp/wiki/Deploy-a-dev-cluster-on-AWS)
- have a look at the cluster dashboard
- make sure amplifier can be addressed by the CLI

PS: the "Docker socket protected with TLS" commit is part of another open PR, it is here only for dependency reason.


ndegory:
superseded by #1802

